PR #40, opened 2015-03-03
Add title to back link

felixinx:
Better way to see. Note that when translating lock, the `$title` variable is empty by default.


glena:
Fixed error and merged manually for next release

PR #41, opened 2015-03-10
Fix sql session

glena:
Fixed session management and issue with schema being created every time the plugin is loaded.


glena:
merged into dev manually

PR #32, opened 2015-01-07
Fixes shortcodes (#33) and Wordpress login (#34)

sandrinodimattia:
Fixes #33 and #34

Issue #54, opened 2015-04-20
Check text on "allow signup" option in plugin settings


glena:
https://ask.auth0.com/t/wordpress-anyone-can-register-and-allow-signup/793

Issue #52, opened 2015-04-14
Auth0 users with different accounts but same username will not be able to log into the site


RedEdgeKatelyn:
With the [recent commit] (https://github.com/auth0/wp-auth0/commit/40377d7a14417d3ceea450bc5822007bdee3f9d5), the error triggered by username collisions is no longer suppressed, but this means that users different accounts with the same username will receive an error message telling them that their username already exists. 

The issue arises because Auth0 usernames are simply the first part of the email address, up to the @ symbol. But, several users can have the same username, since Auth0 only requires the full email to be unique.

Would it be possible to switch from using Auth0 usernames (which are not unique) to Auth0 email addresses (which are) for the WP username?

glena:
It is not as easy as change from username to email. The problem is that in cases where the connection (ie twitter or instagram) doesn't provide an email and only the username. As soon we figure out the best way to handle it we will release a new version.

The idea of the last update is to prevent the security issue that you notify us.

Thank you very much for you collaborations.

glena:
Done => https://github.com/auth0/wp-auth0/pull/55/files#diff-6e06369d1eb2c0b065f6192a23eed41cR39

Issue #48, opened 2015-04-10
Error: Could not create user. The registration process is not available.


cancan101:
I am getting the following error after I login:
```
Error: Could not create user. The registration process is not available.
```

This seems to be caused by this: https://github.com/auth0/wp-auth0/blob/fed6a4916de0e74d365653f3a6e4a6cb0e78c6d6/lib/WP_Auth0_Options.php#L9 which means that you need to turn on WP user creation to have newly visiting Auth0 users work. This seems wrong.

glena:
@cancan101 if you turn the WP "Anyone can register" off, the signup is disabled on lock, why does it seem wrong to you?

![image](https://cloud.githubusercontent.com/assets/5647310/7454873/3af0c704-f24e-11e4-9719-337a8a7a6cfb.png)

cancan101:
Presumably the user has already registered, just not with the WP site. Unchecking this means that the WP will not allow an already registered user to "login for the first time" to the WP site.

glena:
oh, ok. thanks

PR #55, opened 2015-04-27
Auth WP V1.2

glena:
Added new authentication with JWT for api calls.
Merged PR #41, #40, #32


glena:
Issues: #54, #4, #48, #44.
Added integration with auth0/wp-jwt-auth

glena:
Rollback issue #4 "can not be done because we cant restrict the registration with social networks"

glena:
Auth0 users with different accounts but same username will not be able to log into the site #52

glena:
Added support for implicit flow -> Enhancement: Allow WordPress plugin to work in enterprise environment without internet access #42

glena:
SDK Client headers spec compliant #61

PR #65, opened 2015-05-19
Fix issue with email verified requirement

glena:


cancan101:
Shouldn't this be handled by an Auth0 rule?

Issue #57, opened 2015-05-04
Enable new users creation on JWT authentication


glena:

glena:
New jwt auth integration + auto user creation with JWT #70

PR #70, opened 2015-05-27
New jwt auth integration + auto user creation with JWT

glena:


glena:
Enable new users creation on JWT authentication #57

PR #138, opened 2016-02-07
Fix use "self" outside class scope.

elct9620:
The WP_Auth0 class implement `insertAuth0Error` method.
The function `get_currentauth0userinfo` is defined outside class, but it calls `insertAuth0Error` using `self` not `WP_Auth0`.


glena:
Good catch!

I will be releasing a new version soon (check the dev branch).

It would be awesome if you can target this PR to this branch, if not, I will add it as soon I have some free time before the release.

Thanks for your contribution.

glena:
fixed on V.Next

PR #181, opened 2016-03-31
Added Scope Resolution to unserialize function call

caseyjbenko:
The un-scoped call to unserialize has been observed to call functions from other plugins. Adding Scope Resolution ensures the correct function is called.


glena:
thx :)

PR #213, opened 2016-06-10
Fix infinite loop when single logout enabled

emiliorcueto:
When single logout is enabled, this handler references an undefined variable forcing the user to be logged out.  The variable is defined in `/lib/WP_Auth0_LoginManager.php` as `$user_profile`  not `$profile` as currently written in handler logic.

### DIFF
--- a/templates/auth0-singlelogout-handler.php
+++ b/templates/auth0-singlelogout-handler.php
@@ -2,7 +2,7 @@
 <script type="text/javascript">
 (function(){
 
-  var uuids = '<?php echo $profile->user_id; ?>';
+  var uuids = '<?php echo $user_profile->user_id; ?>';
   document.addEventListener("DOMContentLoaded", function() {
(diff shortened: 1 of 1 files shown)


schamp:
I ran into this myself, but wasn't able to track it down.
+1

glena:
sorry about that, we went though a big refactor to change the way it stores the auth0 data (now it uses user meta).

I will try to tackle this as soon as posible.

Thanks for reporting and helping with the troubleshoot

glena:
I just checked this, this was already fixed in the dev branch last week (https://github.com/auth0/wp-auth0/blob/dev/templates/auth0-singlelogout-handler.php)

I am releasing the new version with the fix today

glena:
@emiliorcueto @schamp can you guys upgrade to `3.1.2`?

schamp:
That seemed to work for me, thanks!

PR #222, opened 2016-07-27
Force logo to https

lesaff:
Otherwise if accessed from an https page, it will complain re: mixed content


lorengordon:
Just ran into this as well, and came here trying to figure out how to fix it. :D

PR #233, opened 2016-08-30
Decode the json string 'language_dictionary'

kuljaninemir:
The json string retrieved from settings ($settings['language_dictionary']) is never decoded. This decodes the json string so it actually becomes an array in $settings.


glena:
Thx for the contribution. I will apply this changes manually on the `dev` branch (https://github.com/auth0/wp-auth0#contributions) and it will be ready for the next release.

PR #235, opened 2016-09-05
Tolerate Multisite deployments

zephod:
WordPress Multisite allows users to move between blogs in the network, but the `get_users` query will only return users who registered on _the currently active blog_. 

Auth0 needs to take care of this edge case, because when merging with an existing user, it crashes with `Email already exists` where the user is already signed up to another blog in the network.

It appears that passing `blog_id = 0` switches off the filter and allows you to query all existing users.


glena:
I am applying this change manually on the dev branch.

PR #251, opened 2016-10-13
Add filter to allow auto-provisioning of users, idenpendent of WP setting

marco-s:
We needed to keep WP registration closed, but enabled auto-provisioning of users via SSO (trusted sources). I've simply added a filter to enabled us to do that.

I did see that you had an option in the settings to enabled auto-provisioning of users, independently of WP registration setting. So, there is probably a reason why you removed the functionality.

Would it be possible to merge in this change to allow us to use & update the plugin w/o the need to modify it? Please let us know. Thanks!

### DIFF
--- a/lib/WP_Auth0_Options.php
+++ b/lib/WP_Auth0_Options.php
@@ -16,7 +16,7 @@ class WP_Auth0_Options extends WP_Auth0_Options_Generic {
 		if ( is_multisite() ) {
 			return users_can_register_signup_filter();
 		}
-		return get_site_option( 'users_can_register', 0 ) == 1;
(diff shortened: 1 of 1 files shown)


glena:
Can you explain what is the idea behind the filter or share an example on how are you going to use the filer?

Seems that it will not be contextual on the user profile (to validate the source for example), because you are not passing any info related to the user. If this is the case, I guess the switch in the settings should do the trick since it is doing exactly the same thing https://github.com/auth0/wp-auth0/blob/95d7ead03bd2e4d0114339b6a81165fd21fb159b/lib/WP_Auth0_UsersRepo.php#L96-L97

marco-s:
Yep, thanks, @glena! I did overlook the setting in "Advanced" and only saw the one in "Basic". I'll close this PR.

PR #258, opened 2016-12-10
Fix issue with accessing lock via http:// on https:// connections

jlyon:


glena:
Thank you for the contribution

PR #261, opened 2016-12-27
Fix Yslow error: The following resources have identical contents, but…

jlyon:
… are served from different URLs: https://cdn.auth0.com/js/lock/10.3/lock.min.js

When running my site through YSlow I get the following error: 

The following resources have identical contents, but are served from different URLs. Serve these resources from a consistent URL to save 1 request(s) and 681.6KiB.
https://cdn.auth0.com/js/lock/10.3/lock.min.js
https://cdn.auth0.com/js/lock/10.3/lock.min.js?ver=4.7

SSO is enabled.  The SSO template seems to include Lock manually before its inline JS code. I just removed this <include>. This fixed the issue for me, but further testing may be required.

### DIFF
--- a/templates/auth0-sso-handler-lock10.php
+++ b/templates/auth0-sso-handler-lock10.php
@@ -1,4 +1,3 @@
-<script id="auth0" src="<?php echo $cdn ?>"></script>
 <script type="text/javascript">
 document.addEventListener("DOMContentLoaded", function() {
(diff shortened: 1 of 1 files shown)


cocojoe:
Is this still an issue?

jlyon:
I'm not sure. I stopped using SSO because of this issue.

PR #270, opened 2017-01-17
Fixed four issues: user linking for redirect, implicit flow p2, implicit flow redirect loop, support non-encoded secrets

mostekcm:


mostekcm:
Need to fix the secret encoding first!

PR #290, opened 2017-02-14
Lock 10 Fix for Single Log Out

codyayers:
Following the same pattern [here](https://github.com/auth0/wp-auth0/blob/master/lib/WP_Auth0_LoginManager.php#L53) to include a compatible Lock 10 version of `auth0-singlelogout-handler-lock10.php` when Lock 10 is enabled in the Auth0 Wordpress plugin settings. Please see support ticket #19829


glena:
It needs to be handled a little bit different.
Since new versions of Lock does not exposes auth0.js, we need to inject that js instead of using the lock one.

I will try to do the change this week and let you know

glena:
@codyayers if you can give a try to this PR https://github.com/auth0/wp-auth0/pull/294 I can release it tomorrow

codyayers:
Closing this PR because it was taken care of by @glena with PR [#294](https://github.com/auth0/wp-auth0/pull/294)

PR #294, opened 2017-02-20
fix SSO and SLO with lock 10

glena:
RE #290


codyayers:
I've tested this and it is working, thanks!

glena:
Awesome! I will release it tomorrow morning and let you know for you to upgrade

glena:
`3.2.18` is out with this change :)

Issue #308, opened 2017-04-13
Avoiding "registration process is not available" with SSO for Auth0 users who do not and should not have a WordPress account


idpaterson:
Consider an App with Auth0 users and a WordPress/WooCommerce site with registration disabled where accounts are only created when a subscription to the App is purchased. While SSO is convenient for Auth0 users who also have a WordPress account (these people purchased and manage the subscription and also have access to the App, let's call them the *Billing Managers*), other valid Auth0 users who do not and should not have a WordPress account (these people can access the App only, let's call them *Add-on Users*) are not able to view any page of the site without getting redirected to a `wp_die` page. 

Billing Managers purchase a subscription through WooCommerce which creates a corresponding account (Customer role), then separately assign Add-on Users who can also access the App but have no need for a WordPress account. When a Billing Manager signs into the App then returns to any page of the WordPress site Auth0 reloads the page with the user signed in to the corresponding WordPress Customer account. However, when an Add-on User visits the WordPress site Auth0 redirects to the following error page:

<img width="838" alt="wordpress_ _error" src="https://cloud.githubusercontent.com/assets/507058/25013478/b57b5caa-2041-11e7-8c2c-68a5c1c8fa54.png">

Is there any way to ignore this "registration process is not available" case? The Add-on User is simply visiting a page on the site, they are not trying to sign in. Ideally there would be no redirect and no overt indication to the Add-on User that they are not logged in or not able to create an account. Currently they are unable to visit any page of the WordPress site which seems to only be avoided by disabling SSO.

Potentially relevant settings:

- [x] Single Sign On (SSO)
- [x] Single Logout
- [ ] Multifactor Authentication (MFA)
- [ ] Auto provisioning
- [ ] Use passwordless login
- [x] Force HTTPS callback
- [x] Use Lock 10
- [ ] Remember users session	
- [ ] Link users with same email
- [x] Users Migration
- [x] Auth0 Implicit flow (also tried disabled)
- [x] Requires verified email	
- [x] Auto Login (no widget) (also tried disabled)
- Auto Login Method: Username-Password-Authentication

PR #311, opened 2017-04-17
Use literal 'wp-auth0' rather than WPA0_LANG constant

idpaterson:
Translation tools are generally not able to map a runtime constant like `WPA0_LANG` to a value for the text domain and therefore most of the text within this plugin was not automatically available for translating. A literal string value for the text domain ensures better compatibility.

This pull request updates all occurrences of `WPA0_LANG` to the literal `'wp-auth0'` and adds the text domain in a few places where it was missing. The literal value uses either single or double quotes for consistency with the adjacent translated text. The `WPA0_LANG` constant is still defined to avoid backwards compatibility issues for anything that may still use it.

This problem was discovered when working on #308. I attempted to use the popular Loco translation plugin to change the error message text but only a few strings (those with the default text domain) were available for translation. I was able to use `*` as a wildcard in Loco but then the particular error message that I needed to change was not translatable (see #312).

### DIFF
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -9,7 +9,7 @@
 define( 'WPA0_PLUGIN_FILE', __FILE__ );
 define( 'WPA0_PLUGIN_DIR', trailingslashit( plugin_dir_path( __FILE__ ) ) );
 define( 'WPA0_PLUGIN_URL', trailingslashit( plugin_dir_url( __FILE__ ) ) );
-define( 'WPA0_LANG', 'wp-auth0' );
+define( 'WPA0_LANG', 'wp-auth0' ); // deprecated; do not use for translations
 define( 'AUTH0_DB_VERSION', 14 );
 define( 'WPA0_VERSION', '3.2.23' );
 
--- a/lib/WP_Auth0_Api_Operations.php
+++ b/lib/WP_Auth0_Api_Operations.php
@@ -179,7 +179,7 @@ class WP_Auth0_Api_Operations {
 			$connections = WP_Auth0_Api_Client::search_connection( $domain, $app_token, $strategy );
 
 			// if ( ! $connections ) {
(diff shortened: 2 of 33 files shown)


cocojoe:
Can you fix conflicts please, also why doesn't a constant work?

idpaterson:
WordPress [strongly suggests against it](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#text-domains) because `gettext` which parses these files does not interpret PHP. It knows how to pull out a literal text domain but cannot map a variable or constant to the corresponding value, so you end up with an incorrect or missing text domain that inhibits translation. 

Thanks for taking a look, I will fix up the conflicts.

cocojoe:
Ah good thanks for the reference.

idpaterson:
It seems this once again has conflicts; I'm not inclined to continue fixing conflicts for every minor typo fix that is committed. Please let me know whether this will actually be merged.

cocojoe:
Sorry about this @glena any feelings against this? seems to be the preferred method.

cocojoe:
@idpaterson Can you rebase to master and fix conflicts thanks.

idpaterson:
I am happy to rebase and fix the conflicts if this is en route to acceptance but will wait until you've all agreed that this is an acceptable change.

cocojoe:
This should be fine, please rebase 😄

idpaterson:
This should be good to go, note that there are both text domains switched from the `WPA0_LANG` constant and newly added text domains that were previously blank. There are also a few alternate domains (`wp_auth0_widget_domain`, `wpb_widget_domain`) that were not modified.

cocojoe:
Oh, sorry @idpaterson can you rebase to dev please, same goes for your other PR Thx. So close 😄

idpaterson:
Rebased this and #312 back to dev

glena:
LGTM

PR #312, opened 2017-04-17
Added translation support for a few user-facing exception messages

idpaterson:
A few of the user-facing exception messages regarding existing accounts and inability to sign up were not translatable. This does not affect developer-facing exceptions.

This issue was first encountered in #308 to which a fix has not yet been determined. I was unable to change the default "Could not create user. The registration process is not available. Please contact your site’s administrator." message to say something more specific like "Billing login is available only for subscription managers."

See #311 for the rationale of using 'wp-auth0' rather than `WPA0_LANG` in this commit.


idpaterson:
As a companion to #311 this is also now rebased on master.

PR #313, opened 2017-04-24
Disable base64_encoded by default

thameera:
Since all new clients created are not base64 encoded, it does not make
sense to keep this turned on by default.


cocojoe:
Setup wizard fine.

PR #317, opened 2017-05-22
Improve redirect_login error logging, JWT leeway

cocojoe:
JWTDecode exceptions are now caught and logged in Auth0 Error Log, which should aid CS in debugging in general.  A leeway (could be increased) for clock skew has been added to iat, nbf and exp claims. 
Related to a current ZD Ticket


cocojoe:
auth0_log_token

PR #328, opened 2017-07-24
Prevent script errors throughout the admin panel when anonymous data metrics are disabled

idpaterson:
The `metricsTrack` function, used throughout the admin panel, is undefined if the setting to allow anonymous analytics is disabled. This change defines a no-op function of the same name when analytics are disabled to avoid the script errors.

More than just a few errors in the console, this problem was causing the switches on the Connections page to not work. Analytics click handlers on the switches were tossing an exception which blocked the code that handled the switch functionality.

### DIFF
--- a/lib/WP_Auth0_Metrics.php
+++ b/lib/WP_Auth0_Metrics.php
@@ -66,6 +66,15 @@ class WP_Auth0_Metrics {
       </script>
     <?php
 		}
+		else {
+?>
+      <script>
+        function metricsTrack() {
+          // Metrics are disabled
+        }
+      </script>
+    <?php
+		}
 	}
 
 }


cocojoe:
Can you show me a screenshot of errors, thx
To recreate I just disable 'anonymous analytics'?

idpaterson:
I am not sure whether the particular cocktail of settings we've cooked up is related but in the environment to which I have access, the errors are directly related to toggling the analytics option.

Here is the behavior of the Connections tab with analytics disabled:
> analytics disabled

Here you can see the difference when analytics is enabled. Additionally, the inaccurate warning about unsaved changes that appears in this gif when I press the Save Changes button is caused by the `metricsTrack` function not existing.
> analytics enabled

I see the following errors when toggling a connection:
toggle connection

and when pressing Save Changes:

save changes

cocojoe:
Some feedback from @rdowinton "it looks like the toggle connection button doesn't work even if the anonymous data is enabled (shows `Uncaught TypeError: Cannot read property '1' of null` in console), but if you disable anonymous data it also shows the `metricTrack` error
and his PR does fix the `metricsTrack` error but not the other"

idpaterson:
@cocojoe @rdowinton the pull request fixes errors related to disabling metrics; I have no intention to repair other unrelated bugs in this PR.

PR #339, opened 2017-08-21
Fix login URL

nmec:
Replaces usages of `home_url( '/wp-login.php' )` with `wp_login_url()`.

This generates the correct redirect URL if WordPress is not installed in the root directory: https://codex.wordpress.org/Giving_WordPress_Its_Own_Directory


cocojoe:
Should be covered by #360

PR #356, opened 2018-01-08
Switching wizard admin user creation to use /dbconnections/signup 

joshcanhelp:
In testing dev, @cocojoe and I found that the /api/v2/users endpoint used in `WP_Auth0_Api_Client::create_user` would not authenticate because the API Client being used was not being activated by default for the Database Connection that was being made. We created a new method, `WP_Auth0_Api_Client::signup_user` to use the Authentication signup endpoint.

### DIFF
--- a/lib/initial-setup/WP_Auth0_InitialSetup_AdminUser.php
+++ b/lib/initial-setup/WP_Auth0_InitialSetup_AdminUser.php
@@ -20,23 +20,18 @@ class WP_Auth0_InitialSetup_AdminUser {
 
 		$current_user = wp_get_current_user();
 
-		$db_connection_name = $this->a0_options->get( "db_connection_name" );
-		$domain = $this->a0_options->get( 'domain' );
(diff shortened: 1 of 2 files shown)


lib/initial-setup/WP_Auth0_InitialSetup_AdminUser.php
```diff
@@ -20,23 +20,18 @@ public function callback() {
...
-		$db_connection_name = $this->a0_options->get( "db_connection_name" );
-		$domain = $this->a0_options->get( 'domain' );
-		$jwt = $this->a0_options->get( 'auth0_app_token' );
-
 		$data = array(
+			'client_id' => $this->a0_options->get( 'client_id' ),
 			'email' => $current_user->user_email,
 			'password' => $_POST['admin-password'],
-			'connection' => $db_connection_name,
-			'email_verified' => true
```
cocojoe: So it seems there is a change in behaviour here as the user will need to manually verify their email. I am okay with this.

PR #361, opened 2018-01-12
Fixing "Algorithm not allowed" error during user migration

joshcanhelp:
JWT::decode() expects a third parameter for allowed algorithms that was not being passed in the migration process.


cocojoe:
Looks like I missed that one

Floppy:
Just found this after a day of chasing this problem. Thanks for fixing it already! Any chance of a speedy patch release? :)

cocojoe:
Yes, we are working on a few fixes and are wrapping these up into a patch release for early next week.

PR #364, opened 2018-01-16
Fixed login flow for new tenants, refactored verification email resend

joshcanhelp:
New tenants were unable to create accounts in WordPress because the login was using the deprecated ID token process (legacy accounts had a setting activated that would allow this). This PR switches the login process to use access_tokens, which works for legacy and new. 

This also refactors the resend email process, which was exposing management tokens. 

Caveat here: the WP-created client needs to be authorized on the management API before the resend email process will work.


lib/WP_Auth0_Email_Verification.php
```diff
@@ -0,0 +1,87 @@
...
+	 * AJAX handler to request that the verification email be resent
+	 * Triggered in $this->render_die
+	 */
+	public static function ajax_resend_email() {
+
+		check_ajax_referer( self::$resend_nonce_action, 'nonce' );
+
+		$connect_info = WP_Auth0_Api_Client::get_connect_info();
+
+		$token = WP_Auth0_Api_Client::get_token(
```
lbalmaceda: is it needed to request a new token on each resend trigger?

lib/WP_Auth0_Lock10_Options.php
```diff
@@ -196,7 +196,7 @@ public function has_custom_signup_fields() {
   }
 
   public function get_sso_options() {
-    $options["scope"] = "openid ";
+    $options["scope"] = "openid email identities ";
```
lbalmaceda: I don't remember if `identities` is a valid scope. Is it?

lib/WP_Auth0_LoginManager.php
```diff
@@ -170,15 +170,15 @@ public function init_auth0() {
       }
     } catch (WP_Auth0_LoginFlowValidationException $e) {
 
-      $msg = __( 'There was a problem with your log in', 'wp-auth0' );
+      $msg = __( 'There was a problem with your log in. ', 'wp-auth0' );
```
lbalmaceda: i18n?

lib/WP_Auth0_LoginManager.php
```diff
@@ -170,15 +170,15 @@ public function init_auth0() {
...
       $msg .= ' '. $e->getMessage();
       $msg .= '<br/><br/>';
       $msg .= '<a href="' . wp_login_url() . '">' . __( '← Login', 'wp-auth0' ) . '</a>';
       wp_die( $msg );
 
     } catch (WP_Auth0_BeforeLoginException $e) {
 
-      $msg = __( 'You have logged in successfully, but there is a problem accessing this site', 'wp-auth0' );
+      $msg = __( 'You have logged in successfully, but there is a problem accessing this site. ', 'wp-auth0' );
```
lbalmaceda: i18n?

lib/WP_Auth0_LoginManager.php
```diff
@@ -241,27 +241,11 @@ public function redirect_login() {
...
-
-        // validate that this JWT was made for us
-        if ( $this->a0_options->get( 'client_id' ) !== $decodedToken->aud ) {
-          throw new Exception( 'This token is not intended for us.' );
-        }
-
-        $response = WP_Auth0_Api_Client::get_user( $domain, $data->id_token, $decodedToken->sub );
-      }
+  
+      $data->id_token = null;
```
lbalmaceda: why you remove the id_token but not the access_token?

lib/WP_Auth0_LoginManager.php
```diff
@@ -551,5 +529,15 @@ protected function query_vars( $key ) {
...
-
+  
+  /**
+   * DEPRECATED 3.4.1
+   *
+   * @param $userinfo
+   * @param $id_token
+   */
+  private function dieWithVerifyEmail( $userinfo, $id_token = '' ) {
+    trigger_error( __( 'Method dieWithVerifyEmail is deprecated.', 'wp-auth0' ), E_USER_DEPRECATED);
```
lbalmaceda: I don't fully understand the `trigger_error` method. Is it just logging an error on the console or exploding over the air? If the latter then this is no deprecation. This is breaking the api 🤕 . Old methods must remain the same, working without changing their signature at all. Normally we achieve this by re-routing the call to the new method that will *eventually* replace it. On a side note, we prefer to handle deprecations on a separate PR so the changelog is clearer. Would you mind rolling back this change?

lib/exceptions/WP_Auth0_EmailNotVerifiedException.php
```diff
@@ -1,13 +1,21 @@
...
 
 	public $userinfo;
 	public $id_token;
 
-	public function __construct( $userinfo, $id_token ) {
+	/**
+	 * WP_Auth0_EmailNotVerifiedException constructor.
+	 *
+	 * @param stdClass $userinfo - userinfo object returned from Auth0
+	 * @param string $id_token - DEPRECATED 3.4.1, should not be output in any template
```
lbalmaceda: Move to a separate PR please

lib/WP_Auth0_UsersRepo.php
```diff
@@ -161,41 +161,51 @@ public function create( $userinfo, $token, $access_token = null, $role = null, $
...
 
-	public function find_auth0_user( $id ) {
+	/**
+	 * Look for and return a user with an Auth0 ID
+	 *
+	 * @param string $id - An Auth0 user ID, like "provider|id"
+	 *
+	 * @return null|WP_User
+	 */
+	public function find_auth0_user ( $id ) {
```
lbalmaceda: Just asking. Why is this method needed?

assets/js/die-with-verify-email.js
```diff
@@ -0,0 +1,30 @@
+/* globals jQuery, console, WPAuth0EmailVerification */
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - The process dies using wp_die() and only happens if the incoming user needs to verify the email. Baiscally: process stops, nothing proceeds.

assets/js/die-with-verify-email.js
```diff
@@ -0,0 +1,30 @@
...
+        var postData = {
+            action: 'resend_verification_email',
+            nonce: WPAuth0EmailVerification.nonce,
+            sub: WPAuth0EmailVerification.sub
+        };
+
+        $.post( WPAuth0EmailVerification.ajaxUrl, postData )
+            .done( function( data ) {
+
+                if ( '1' === data ) {
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - Unfortunately WP has a default magic number for AJAX calls ("-1" if you don't pass the nonce check) so a bit of a bad pattern WP folks follow. Will adjust to something more clear.

lib/WP_Auth0_Api_Client.php
```diff
@@ -2,6 +2,58 @@
...
+		if ( empty( $domain ) ) {
+			$a0_options = WP_Auth0_Options::Instance();
+			$domain = $a0_options->get( 'domain' );
+		}
+
+		if ( ! empty( $path[0] ) && '/' === $path[0] ) {
+			$path = substr( $path, 1 );
+		}
+
+		return "https://{$domain}/{$path}";
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - Treating `$domain` as a sub-domain is pretty much par for the course in this file. Agree that there could be better sanitization at some point but all API calls in the plugin will fail if that's not formatted right.

lib/WP_Auth0_Api_Client.php
```diff
@@ -2,6 +2,58 @@
...
+	 * Generate the API endpoint with a provided domain
+	 *
+	 * @since 3.4.1
+	 *
+	 * @param string $path - API path appended to the domain
+	 * @param string $domain - domain to use, blank uses default
+	 *
+	 * @return string
+	 */
+	public static function get_endpoint( $path = '', $domain = '' ) {
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - Everything is `public static` in that file, for better or worse. This is a case of consistency vs accuracy :)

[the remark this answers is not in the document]
lbalmaceda: But if this file is going to end up being the "API Client" then I would prefer not to leave this method open for users to use, as it's only used internally, and also because once we publish this we can't remove it without breaking the public api.

lib/WP_Auth0_Api_Client.php
```diff
@@ -2,6 +2,58 @@
...
+	 * @param string $opt
+	 *
+	 * @return string|array
+	 */
+	public static function get_connect_info( $opt = '' ) {
+
+		$a0_options = WP_Auth0_Options::Instance();
+
+		$connect_info = array(
+			'domain' => $a0_options->get( 'domain' ),
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - How do you mean?

lib/WP_Auth0_Api_Client.php
```diff
@@ -2,6 +2,58 @@
...
+			'client_id' => $a0_options->get( 'client_id' ),
+			'client_secret' => $a0_options->get( 'client_secret' ),
+			'connection' => $a0_options->get( 'db_connection_name' ),
+			'audience' => self::get_endpoint( 'api/v2/' ),
+		);
+
+		if ( empty( $opt ) ) {
+			return $connect_info;
+		} else {
+			return ! empty( $connect_info[ $opt ] ) ? $connect_info[ $opt ] : '';
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - Agreed, a little sloppy here.

lib/WP_Auth0_Api_Client.php
```diff
@@ -149,6 +226,44 @@ public static function search_users( $domain, $jwt, $q = "", $page = 0, $per_pag
...
+	 * @return bool
+	 */
+	public static function resend_verification_email( $access_token, $user_id ) {
+
+		$response = wp_remote_post(
+			self::get_endpoint( 'api/v2/jobs/verification-email' ),
+			array(
+				'headers' => self::get_headers( $access_token ),
+				'body' => json_encode( array(
+					'user_id' => $user_id,
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - Validated upstream

lib/WP_Auth0_EditProfile.php
```diff
@@ -185,21 +185,23 @@ function DeleteMFA(event) {
 	}
 
 	public function show_change_password() {
-    $current_user = get_currentauth0user();
-    $user_profile = $current_user->auth0_obj;
+		$current_user = get_currentauth0user();
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - I don't like that format either but renaming all that is a bit out of scope on this. Also this:

https://github.com/auth0/wp-auth0/blob/master/WP_Auth0.php#L419

lib/WP_Auth0_EditProfile.php
```diff
@@ -185,21 +185,23 @@ function DeleteMFA(event) {
...
+		$user_profile = $current_user->auth0_obj;
 
 		if ( empty( $user_profile ) ) return;
 
 		$connection = null;
 
-		foreach ( $user_profile->identities as $identity ) {
-			if ( $identity->provider === 'auth0' ) {
-				$connection = $identity->connection;
+		if ( ! empty( $user_profile->identities ) && is_array( $user_profile->identities ) ) {
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - probably (miniscule) but out of scope here. Just squashing a warning that was appearing.

lib/WP_Auth0_EditProfile.php
```diff
@@ -185,21 +185,23 @@ function DeleteMFA(event) {
...
+			foreach ( $user_profile->identities as $identity ) {
+				if ( $identity->provider === 'auth0' ) {
+					$connection = $identity->connection;
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - Again, out of scope ... any added lines here are formatting

lib/WP_Auth0_EditProfile.php
```diff
@@ -226,20 +228,21 @@ public function show_change_password() {
...
+		$current_user = get_currentauth0user();
+		$user_profile = $current_user->auth0_obj;
 
 		if ( $user_profile && !empty( $user_profile ) ) {
 			$connection = null;
-
-			foreach ( $user_profile->identities as $identity ) {
-				if ( $identity->provider === 'auth0' ) {
-					$connection = $identity->connection;
+			if ( ! empty( $user_profile->identities ) && is_array( $user_profile->identities ) ) {
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - Lots of improvements to be had here at some point :)

lib/WP_Auth0_Email_Verification.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+/**
+ * Class WP_Auth0_Email_Verification
+ */
+class WP_Auth0_Email_Verification {
+
+	private static $resend_nonce_action = 'auth0_resend_verification_email';
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - Probably, yes, but final classes aren't used anywhere in this project currently.

lib/WP_Auth0_Email_Verification.php
```diff
@@ -0,0 +1,87 @@
...
+	public static function init() {
+		add_action( 'wp_ajax_nopriv_resend_verification_email', 'wp_auth0_ajax_resend_verification_email' );
+	}
+
+	/**
+	 * Stop the login process and show email verification prompt
+	 *
+	 * @param object $userinfo
+	 */
+	public static function render_die( $userinfo ) {
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - Not only that but rendering as well! 🔫

lib/WP_Auth0_Api_Client.php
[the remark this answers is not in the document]
lbalmaceda: I don't know how common it is on wp, but if a value is being repeated across the class you would define a constant and use that instead. i.e. 

```php
const KEY_DOMAIN = "domain";

KEY_DOMAIN => $a0_options->get( KEY_DOMAIN )
```

lib/WP_Auth0_Email_Verification.php
```diff
@@ -0,0 +1,87 @@
...
+
+		$html = sprintf(
+			'<p>%s<br><a id="js-a0-resend-verification" href="#">%s</a></p>
+			<p><a href="%s?%d">%s</a></p>
+			<script>var WPAuth0EmailVerification={ajaxUrl:"%s",sub:"%s",nonce:"%s",e_msg:"%s",s_msg:"%s"}</script>
+			<script src="%s"></script>
+			<script src="%s"></script>',
+
+			__( 'Please verify your email and log in again.', 'wp-auth0' ),
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - `__( 'Translate me', 'text-domain' )` is the WordPress translation function. We don't actually have a PO file for this yet, should generate one to assist with translation.

lib/WP_Auth0_Email_Verification.php
joshcanhelp: @lbalmaceda - No, not required ... should probably be cached in a number of places.

lib/WP_Auth0_Email_Verification.php
```diff
@@ -0,0 +1,87 @@
...
+	 * Triggered in $this->render_die
+	 */
+	public static function ajax_resend_email() {
+
+		check_ajax_referer( self::$resend_nonce_action, 'nonce' );
+
+		$connect_info = WP_Auth0_Api_Client::get_connect_info();
+
+		$token = WP_Auth0_Api_Client::get_token(
+			$connect_info['domain'],
```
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - Constants are compile-time, not run-time and this comes from the DB.

lib/WP_Auth0_EditProfile.php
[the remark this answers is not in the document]
lbalmaceda: What do you mean by "out of scope", is this a public api breaking change?

lib/WP_Auth0_Lock10_Options.php
joshcanhelp: @lbalmaceda - Used the code below as an example:

https://github.com/auth0/wp-auth0/blob/master/lib/WP_Auth0_Lock_Options.php#L227

During debugging, I dumped the $userinfo object and has identities so I think 👍

lib/WP_Auth0_LoginManager.php
joshcanhelp: @lbalmaceda - yes?

lib/WP_Auth0_LoginManager.php
joshcanhelp: @lbalmaceda - yes?

lib/WP_Auth0_Email_Verification.php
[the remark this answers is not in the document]
lbalmaceda: I mean, why not change it to a `const`? (constant)

lib/WP_Auth0_LoginManager.php
joshcanhelp: @lbalmaceda - Not sure I'm totally clear on your question but this was basically 'if you have an id token, use it, otherwise use the access token' and we want to get rid of the first condition. I'm forcing it to use the access_token and left the `$data->id_token = null;` because I don't know what the downstream effects are of keeping that id_token around. Maybe there is another check like this down the line? Seemed safer to leave this than remove it.

lib/WP_Auth0_Email_Verification.php
lbalmaceda: can you fix this on this same PR?

lib/WP_Auth0_LoginManager.php
joshcanhelp: @lbalmaceda - `trigger_error` is just that, triggers an error of a certain type. Depending on the error type and the environment, it will be handled different ways. In the case of a deprecated error like this one, it's either displayed or not, won't stop the process from continuing. 

Happy to do whatever is prudent here but the idea here is that the function still exists and can be called as based but does the right thing with a soft error. Signature is essentially the same, just defaulting to a blank `$id_token` and the replacement function is called with the `$userinfo` it needs. The chances that this function is being used in the wild is very small but, if so, this should keep folks running. 

Let me know if you still want me to back it out and how to handle going forward!

lib/WP_Auth0_UsersRepo.php
joshcanhelp: @lbalmaceda - Finding the WP user with Auth0 info

lib/WP_Auth0_Api_Client.php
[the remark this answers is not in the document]
joshcanhelp: @lbalmaceda - I see what you're getting at here. Was a little confused with who you were referring to as "users" but I see the 💡

lib/exceptions/WP_Auth0_EmailNotVerifiedException.php
joshcanhelp: @lbalmaceda - Will wait for direction on the other one so I can do them both. Should they each be in their own PR?

lbalmaceda: Not necessary as long as the PR only deprecates stuff 👌

lib/WP_Auth0_LoginManager.php
cocojoe: Should probably change these back as this will break existing i18n for this message as it's changed.

lib/WP_Auth0_LoginManager.php
cocojoe: Should probably change these back as this will break existing i18n for this message as it's changed.

lbalmaceda: It's not that the main message changed but fixed grammatically, I see no need to rollback this change.

lib/WP_Auth0_Lock10_Options.php
cocojoe: This will only work in non OIDC conformance, now the plugin doesn't work in OIDC Conformant mode anyway. However, just want you to be aware. This can be refactored in 4.0.0 to use the MGMT API get user.

lbalmaceda: True. Then maybe it's worth to document it somewhere in the readme letting people know that it won't work on legacy clients.

lib/WP_Auth0_LoginManager.php
cocojoe: Well fixing it breaks it for existing use, no huge feeling either way.

cocojoe:
@joshcanhelp Have you tested on new and old (make sure still works) tenants?

joshcanhelp:
@cocojoe - Yes, walked through start to finish on old and new tenants

lib/WP_Auth0_LoginManager.php
joshcanhelp: We don't have a PO file in here that we keep up to date so if someone is generating that and translating for themselves, they'll be used to doing that on each release (or used to things breaking/not translating).

PR #365, opened 2018-01-19
updating CDN URLs for Lock and Auth.js

joshcanhelp:


cocojoe:
What about user who already using using 3.4.0? Will they get updated?

joshcanhelp:
@cocojoe - It seems to me like if we give folks control over the version then we don't want to just change it on an update, even if it seems net positive from our perspective. I do see that this was done for 3.4.0 but seems a little heavy-handed. What if someone is using an old version on purpose and we just wipe that and set it to something else?

joshcanhelp:
@cocojoe - That makes sense, assuming there is nothing breaking. Looking through the commits on the 2 releases since 11.0, seems fairly minor. I'll get another PR through

PR #366, opened 2018-01-19
Creating client_grant for management API

joshcanhelp:
Clients used with this plugin need to have a specific grant to the management API in order to handle certain actions like verification emails. The client also needs to have an app_type set, in this case "Regular Web Application."

This pull request:

- Adds the app_type and client_grant to the client creation process when the plugin is first installed
- Updates the DB version to 16 and attempts to fix the app_type and client_grant
- If this process fails, a flag is set as an option and a banner with instructions will appear until the process is complete
- Added a process to extra the audience from a manually-added app token during validation
- Added a check for the client grant when an app token is saved to re-run the client_grant process
- A few spelling and formatting errors throughout

### DIFF
--- a/lib/WP_Auth0_DBManager.php
+++ b/lib/WP_Auth0_DBManager.php
@@ -14,7 +14,9 @@ class WP_Auth0_DBManager {
 		if ($this->current_db_version === 0) {
 			$this->current_db_version = (int)get_site_option( 'auth0_db_version', 0 );
 		}
+
 		add_action( 'plugins_loaded', array( $this, 'check_update' ) );
+		add_action( 'admin_notices', array( $this, 'notice_failed_client_grant' ) );
 	}
 
 	public function check_update() {
@@ -23,9 +25,22 @@ class WP_Auth0_DBManager {
 		}
 	}
 
-	public function install_db() {
+	public function install_db( $version_to_install = null, $app_token = '' ) {
+
+		wp_cache_set( 'doing_db_update', TRUE, WPA0_CACHE_GROUP );
 
 		$options = WP_Auth0_Options::Instance();
+
+		if ( empty( $app_token ) ) {
+			$app_token = $options->get( 'auth0_app_token' );
+		}
+
+		$connection_id = $options->get( 'db_connection_id' );
+		$migration_token = $options->get( 'migration_token' );
+		$client_id = $options->get( 'client_id' );
+		$client_secret = $options->get( 'client_secret' );
+		$domain = $options->get( 'domain' );
+		$sso = $options->get( 'sso' );
 
 		if ($this->current_db_version === 0) {
 			$options->set('auth0_table', false);
@@ -38,16 +53,13 @@ class WP_Auth0_DBManager {
 		if ( $this->current_db_version <= 7 ) {
 			if ( $options->get( 'db_connection_enabled' ) ) {
 
-				$app_token = $options->get( 'auth0_app_token' );
-				$connection_id = $options->get( 'db_connection_id' );
-				$migration_token = $options->get( "migration_token" );
-
 				$operations = new WP_Auth0_Api_Operations( $options );
+
 				if ( !empty( $app_token ) &&
 					!empty( $connection_id ) &&
 					!empty( $migration_token ) ) {
 
-					$response = $operations->update_wordpress_connection(
+					$operations->update_wordpress_connection(
 						$app_token,
 						$connection_id,
 						$options->get( 'password_policy' ),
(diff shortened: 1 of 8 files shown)


cocojoe:
@joshcanhelp for these bigger commits, the formatting type corrections just add unnecessary noise to the review. I'm all for improving readability and consistent. However, in the future these would be better in a separate PR.

cocojoe:
Please rebase and a good time to try the mergetool 😄

joshcanhelp:
@cocojoe - GTG

lib/WP_Auth0_DBManager.php
```diff
@@ -38,16 +53,13 @@ public function install_db() {
...
-				$migration_token = $options->get( "migration_token" );
-
 				$operations = new WP_Auth0_Api_Operations( $options );
+
 				if ( !empty( $app_token ) &&
 					!empty( $connection_id ) &&
 					!empty( $migration_token ) ) {
 
-					$response = $operations->update_wordpress_connection(
+					$operations->update_wordpress_connection(
```
cocojoe: $response never used?

lib/WP_Auth0_DBManager.php
```diff
@@ -92,56 +104,159 @@ public function install_db() {
 
 		if ( $this->current_db_version < 12 ) {
 
-      if ( strpos( $cdn_url, '10.' ) === false ) {
-        $options->set('use_lock_10', false);
-      } else {
-        $options->set('use_lock_10', true);
-      }
+			if ( strpos( $cdn_url, '10.' ) === false ) {
```
cocojoe: I mentioned this but leaving it for other reviewers, in the future let's cover these changes in a separate internal PR, to keep the review process cleaner.

lib/WP_Auth0_LoginManager.php
```diff
@@ -529,15 +529,15 @@ protected function query_vars( $key ) {
...
-   * DEPRECATED 3.4.1
-   *
-   * @param $userinfo
-   * @param $id_token
-   */
-  private function dieWithVerifyEmail( $userinfo, $id_token = '' ) {
-    trigger_error( __( 'Method dieWithVerifyEmail is deprecated.', 'wp-auth0' ), E_USER_DEPRECATED);
-    WP_Auth0_Email_Verification::render_die( $userinfo );
-  }
+
```
cocojoe: Separate PR  in future.

lib/admin/WP_Auth0_Admin_Basic.php
```diff
@@ -220,47 +237,86 @@ public function auth0_delete_cache_transient() {
...
+			}
+		}
 
-		$error = '';
-		$completeBasicData = true;
 		if ( empty( $input['domain'] ) ) {
-			$error = __( 'You need to specify domain', 'wp-auth0' );
-			$this->add_validation_error( $error );
-			$completeBasicData = false;
+			$this->add_validation_error( __( 'You need to specify domain', 'wp-auth0' ) );
```
cocojoe: This is nice

lib/WP_Auth0_DBManager.php
joshcanhelp: @cocojoe - Understood. Can all formatting and minor stuff go together in a final cleanup PR or something? 

Most of the cases where there was formatting changes were part of a minor refactor. In this case, I moved all the vars we were using to the top of that method and reformatted a couple of the blocks.

lib/WP_Auth0_DBManager.php
joshcanhelp: Nope!

lib/WP_Auth0_DBManager.php
cocojoe: Yeah, it's not a priority. However, if you have some time and want to do a lint pass and also adopt some linting standards then it would be good to do. Although more likely this will be for the vNext when this can be standardised from the start.

joshcanhelp:
@cocojoe - If the upgrade works without manual intervention, you won't see anything at all. But, if not, you'll see this message at the top of all screens if you're an admin (`update_plugins` capability):

screenshot 2018-01-22 07 26 21

You can simulate this on your local install by turning off the Non-Interactive Client created by WP in the Auth0 Management API:

screenshot 2018-01-22 07 27 56

... and adding an option:

`update_option( 'wp_auth0_client_grant_failed', 1 );`

glena:
I think it looks good, I would add a patch to the client to set the `grant_types`. For wegular webapp you will only need `code` and `client_credentials`. If implicit is enabled, you should add `implicit` and remove `code`

joshcanhelp:
@glena - Looks like the clients we're creating have the three `grant_types` you mentioned, as well as "Refresh Token" by default. That's the case when you create a new one in the dashboard as well.

PR #367, opened 2018-01-22
Fixing DB version upgrade

joshcanhelp:
- Upgrading Lock 11.0.0 => 11.1
- Upgrading Auth.js 9.0.0 => 9.1
- Checking for an existing client grant for the management API before failing
- Admin notice if client grant is successful to info and catch edge cases

Waiting for #366 to be completed before submitting

### DIFF
--- a/lib/WP_Auth0_Api_Client.php
+++ b/lib/WP_Auth0_Api_Client.php
@@ -7,7 +7,7 @@ class WP_Auth0_Api_Client {
 	/**
 	 * Generate the API endpoint with a provided domain
 	 *
-	 * @since 3.4.1
+	 * @since 3.5.0
 	 *
 	 * @param string $path - API path appended to the domain
 	 * @param string $domain - domain to use, blank uses default
@@ -31,7 +31,7 @@ class WP_Auth0_Api_Client {
 	/**
 	 * Return basic connection information, or a specific value
 	 *
-	 * @since 3.4.1
+	 * @since 3.5.0
 	 *
 	 * @param string $opt - specific option needed, returns all if blank
 	 *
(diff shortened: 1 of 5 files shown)


lib/WP_Auth0_DBManager.php
```diff
@@ -259,6 +274,24 @@ public function notice_failed_client_grant() {
...
 
+	public function notice_successful_client_grant() {
+
+		if ( ! get_option( 'wp_auth0_client_grant_success' ) ) {
+			return;
+		}
+		?>
+		<div class="notice notice-success">
+			<p><?php
+				_e( 'As a part of the 3.4.1 upgrade, a client grant was created for the Auth0 Management API.', 'wp-auth0' );
```
cocojoe: I think this should be 3.5.0 now, there are a lot of fixes in this release so it's significant to warrant, same for all 3.4.1 references.

lib/WP_Auth0_Api_Client.php
```diff
@@ -550,7 +550,7 @@ public static function create_client_grant( $app_token, $client_id ) {
 		$data = array(
 			'client_id' => $client_id,
 			'audience' => self::get_connect_info( 'audience' ),
-			'scope' => self::ConsentRequiredScopes()
+			'scope' => self::get_required_scopes()
```
cocojoe: this add in the last PR?

joshcanhelp: @cocojoe - I'm not sure what you're asking here.

cocojoe: Point me to this method please

joshcanhelp: @cocojoe - Couldn't find it? 😆

https://github.com/auth0/wp-auth0/blob/master/lib/WP_Auth0_Api_Client.php#L221

Issue #348, opened 2017-09-21
API token expires causing update_user to fail


rdowinton:
The management API token seems to expire some time after installing the plugin. 

The result is that if a user updates their email address in their profile, it changes on the WordPress side but not the Auth0 side. They can still login via Auth0 with their old email address, and this then reverts the email address in WordPress.

Side note: this is also incorrectly logged as `update_users` instead of `update_user`.

joshcanhelp:
@rdowinton - Release 3.5.0 (#373) should take care of this for non-implicit users. Please test and close this if we're good to go here. Thanks!

PR #368, opened 2018-01-23
Added use Management API for user data

joshcanhelp:
This PR changes where user information comes from on login. Switching from the Auth API /userinfo to the Management API /users/user_id to get identities. 

* Added WP_Auth0_Api_Client::get_client_token() to quickly get a management token
* Updated WP_Auth0_Api_Client::resend_verification_email() to use this token function
* Replaced WP_Auth0_Api_Client::get_user_info() with WP_Auth0_Api_Client::get_user() to return user information during login (management API access is acquired as part of this upgrade), which fixes identities being returned
* Added read:users as a required scope
* Updated DB upgrade to use new WP_Auth0_Api_Client::get_client_token() function to confirm that a client_credentials token can be returned

### DIFF
--- a/lib/WP_Auth0_Api_Client.php
+++ b/lib/WP_Auth0_Api_Client.php
@@ -204,6 +204,42 @@ class WP_Auth0_Api_Client {
 		return $response;
 	}
 
+	/**
+	 * Get a client_credentials token using default stored connection info
+	 *
+	 * @since 3.4.1
+	 *
+	 * @return bool|string
+	 */
+	public static function get_client_token() {
+
+		$response = wp_remote_post( self::get_endpoint( 'oauth/token' ), array(
+				'headers' => self::get_headers( '', 'application/x-www-form-urlencoded' ),
+				'body' => array(
+					'client_id' => self::get_connect_info( 'client_id' ),
+					'client_secret' => self::get_connect_info( 'client_secret' ),
+					'audience' => self::get_connect_info( 'audience' ),
+					'grant_type' => 'client_credentials',
+				),
+			) );
+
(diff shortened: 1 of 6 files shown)


joshcanhelp:
@cocojoe @glena - After our conversation about the Implicit flow use case, I'm wondering if this (and probably a few other things) needs to get re-worked. Implicit users, if they really are blocked from calling the Auth0 server, won't benefit from a client grant, won't be able to use the resend email verification anyways, and won't have the user information that makes it possible to change password, email, etc. Maybe all of that is expected (maybe not?) in which case we probably need to check for that.

lib/WP_Auth0_Api_Client.php
```diff
@@ -204,6 +204,42 @@ public static function get_token( $domain, $client_id, $client_secret, $grantTyp
...
+	 * Get a client_credentials token using default stored connection info
+	 *
+	 * @since 3.4.1
+	 *
+	 * @return bool|string
+	 */
+	public static function get_client_token() {
+
+		$response = wp_remote_post( self::get_endpoint( 'oauth/token' ), array(
+				'headers' => self::get_headers( '', 'application/x-www-form-urlencoded' ),
```
cocojoe: where is `content-type` specified?

joshcanhelp: @cocojoe - `self::get_headers()`

joshcanhelp:
@cocojoe - RE: implicit scope ... it does, look a couple of lines down, you'll see implicit adds additional scopes.

joshcanhelp:
Fixes #348

Issue #355, opened 2017-12-11
"Override WordPress avatars" option doesn't appear to work with comments


thameera:
Even when the "Override WordPress avatars" in the Features tab is turned on, the picture for the comments section, etc still seem to use Gravatar instead of user's picture. However the user's profile page (in admin panel) shows the correct picture.

howarddc:
Here's the fix:

/plugins/auth0/WP_Auth0.php Line 160, add:

`if (is_object($id_or_email) && isset($id_or_email->user_id)) $id_or_email=$id_or_email->user_id;
`

The issue is that the comments version of this filter sends the full WP_Comment object as the second param to get_avatar, and the code was assuming it would always be an integer user ID. Instead, the user ID is a property of the object.

joshcanhelp:
From /wp-includes/pluggable.php L:2550:

"The Gravatar to retrieve. Accepts a user_id, gravatar md5 hash, user email, WP_User object, WP_Post object, or WP_Comment object."

Assigning to myself to address in the next next release.

PR #376, opened 2018-01-25
Fixed get_avatar hooked function to account for other user identifiers

joshcanhelp:
See #355 for details


cocojoe:
Got a screenshot of this in action?

PR #382, opened 2018-01-29
Fixed migration IPs being saved

joshcanhelp:


lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -441,8 +445,8 @@ public function basic_validation( $old_options, $input ) {
       if ( isset( $input['connections']['social_facebook_secret'] ) ) $input['connections']['social_facebook_secret'] = sanitize_text_field( $input['connections']['social_facebook_secret'] );
     }
 
-    $input['migration_ips_filter'] =  ( isset( $input['migration_ips_filter'] ) ? $input['migration_ips_filter'] : 0 );
-    $input['migration_ips'] = sanitize_text_field( $old_options['migration_ips'] );
+    $input['migration_ips_filter'] =  ( ! empty( $input['migration_ips_filter'] ) ? 1 : 0 );
```
cocojoe: The value of `migration_ips_filter` just used to be a boolean I take it?

joshcanhelp: @cocojoe - Boolean UX, yes. Technically integer, `value="1"` on the control

Issue #285, opened 2017-02-06
3.2.16 throws errors if Error Log is empty


mostekcm:
Two things need to be fixed for this.

1) wp-content/plugins/auth0/lib/WP_Auth0_ErrorManager.php
Add: 
$log = get_option('auth0_error_log', array());
         if (!is_array($log)) {
             $log = array();
          }

For some reason the $log is returning false instead of array().

2) auth0/templates/auth0-error-log.php
Add else condition after check for empty around the foreach.  Same reason.  Returning boolean instead of array when empty.

cocojoe:
@mostekcm this still relevant?

PR #384, opened 2018-01-29
Changed error handling

joshcanhelp:
- Using `__METHOD__` magic constant instead of hard-coded method name
- Improved WP_Auth0_ErrorManager:: insert_auth0_error() based on #285

### DIFF
--- a/lib/WP_Auth0_DBManager.php
+++ b/lib/WP_Auth0_DBManager.php
@@ -392,7 +392,7 @@ class WP_Auth0_DBManager {
 		if ( is_null( $userRows ) ) {
 			return;
 		} elseif ( $userRows instanceof WP_Error ) {
-			WP_Auth0_ErrorManager::insert_auth0_error( 'migrate_users_data', $userRows );
+			WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__, $userRows );
 			return;
 		}
 
@@ -425,7 +425,7 @@ class WP_Auth0_DBManager {
 		$results = get_users( $query );
 
 		if ( $results instanceof WP_Error ) {
-			WP_Auth0_ErrorManager::insert_auth0_error( 'findAuth0User', $userRow );
+			WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__, $results->get_error_message() );
 			return array();
 		}
 
--- a/lib/WP_Auth0_Routes.php
+++ b/lib/WP_Auth0_Routes.php
@@ -204,7 +204,7 @@ EOT;
 			}
 
 			if ($user instanceof WP_Error) {
-				WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__, $user );
+				WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__, $user->get_error_message() );
 				$user = array('error' => 'invalid credentials');
 			} else {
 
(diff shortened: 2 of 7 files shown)


lib/admin/WP_Auth0_Admin_Basic.php
```diff
@@ -229,7 +229,7 @@ public function render_basic_description() {
 	public function auth0_delete_cache_transient() {
 		if ( ! is_admin() ) return;
 
-		WP_Auth0_ErrorManager::insert_auth0_error( 'WP_Auth0_Admin_Basic::delete_cache_transient', 'deleting cache transient' );
+		WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__, 'deleting cache transient' );
```
cocojoe: I must be missing something here, it seems like it will just log this error every time and then it calls the method to delete cache?

lib/WP_Auth0_DBManager.php
```diff
@@ -425,7 +425,7 @@ public function get_auth0_users( $user_ids = null ) {
 		$results = get_users( $query );
 
 		if ( $results instanceof WP_Error ) {
-			WP_Auth0_ErrorManager::insert_auth0_error( 'findAuth0User', $userRow );
+			WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__, $results->get_error_message() );
```
cocojoe: Presume ` $results->get_error_message() ` is more useful than `$userRow` ?

lib/WP_Auth0_Routes.php
```diff
@@ -204,7 +204,7 @@ protected function migration_ws_get_user() {
 			}
 
 			if ($user instanceof WP_Error) {
-				WP_Auth0_ErrorManager::insert_auth0_error( 'migration_ws_get_user', $user );
+				WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__, $user );
```
cocojoe: In other places https://github.com/auth0/wp-auth0/pull/384/files#diff-277e60f8ce82a93a9c30ed5aa4c01dd9R189

You will show the name of the method called but not in this one e.g. `get_user_by` , does it have a `get_error_message` also available?

lib/WP_Auth0_DBManager.php
joshcanhelp: `$userRow` was never defined here

lib/admin/WP_Auth0_Admin_Basic.php
joshcanhelp: It looks to me like it's there as a message rather than an error. I've never had a use for it (or seen it in the logs) but I'm guessing it's there for a reason.

cocojoe: Well should be a general log vs error. However that can be a story for another day.

joshcanhelp:
@cocojoe - Totally agree on the log verbiage. I have a few ideas to improve in a way that can be re-used on the re-write. 

TY!

lib/WP_Auth0_Routes.php
joshcanhelp: @cocojoe - Actually ... that function looks for a `WP_Error` object and handles nicely:

https://github.com/auth0/wp-auth0/blob/master/lib/WP_Auth0_ErrorManager.php#L7

PR #385, opened 2018-01-29
Fixed wrong title and icon for login widget

joshcanhelp:
Swapped in `array_replace_recursive` instead of `array_merge_recursive` to avoid an array being created that the Lock widget couldn't understand.

### DIFF
--- a/lib/WP_Auth0_Lock10_Options.php
+++ b/lib/WP_Auth0_Lock10_Options.php
@@ -254,8 +254,7 @@ class WP_Auth0_Lock10_Options {
       $extraOptions["auth"]["redirectUrl"] = $this->get_code_callback_url();
     }
 
-    $options_obj = array_merge_recursive( $extraOptions, $options_obj  );
-    $options_obj = array_merge_recursive( $options_obj , $extended_settings );
+    $options_obj = array_replace_recursive( $extraOptions, $options_obj, $extended_settings );
 
     if ( ! $this->show_as_modal() ) {
       $options_obj['container'] = 'auth0-login-form';
--- a/lib/WP_Auth0_Lock_Options.php
+++ b/lib/WP_Auth0_Lock_Options.php
@@ -230,8 +230,7 @@ class WP_Auth0_Lock_Options {
 			$extraOptions["callbackURL"] = $this->get_code_callback_url();
 		}
 
-		$options_obj = array_merge( $extraOptions, $options_obj  );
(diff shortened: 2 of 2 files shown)


lib/WP_Auth0_Lock10_Options.php
```diff
@@ -254,8 +254,7 @@ public function get_lock_options() {
       $extraOptions["auth"]["redirectUrl"] = $this->get_code_callback_url();
     }
 
-    $options_obj = array_merge_recursive( $extraOptions, $options_obj  );
-    $options_obj = array_merge_recursive( $options_obj , $extended_settings );
+    $options_obj = array_replace_recursive( $extraOptions, $options_obj, $extended_settings );
```
cocojoe: So what is the logic here for duplicate entries. `extraOptions` are over ridden by `options_obj` but this can then be over ridden by `extended_settings`?

lib/WP_Auth0_Lock_Options.php
```diff
@@ -230,8 +230,7 @@ public function get_lock_options() {
 			$extraOptions["callbackURL"] = $this->get_code_callback_url();
 		}
 
-		$options_obj = array_merge( $extraOptions, $options_obj  );
-		$options_obj = array_merge( $options_obj , $extended_settings );
+		$options_obj = array_merge( $extraOptions, $options_obj, $extended_settings );
```
cocojoe: In the previous snippet you change `merge` to `replace` why does this one stay as merge?

lib/WP_Auth0_Lock10_Options.php
joshcanhelp: @cocojoe - Correct.

lib/WP_Auth0_Lock_Options.php
joshcanhelp: @cocojoe - Array dimensions, recursive vs. not recursive. `array_merge_recursive` can create arrays out of strings, whereas `array_replace_recursive` overrides. `array_merge` also overrides but it only goes 1 level deep. Looks like the options arrays are built differently.

joshcanhelp:
@cocojoe - Ready for re-review

lib/WP_Auth0_Lock10_Options.php
cocojoe: the naming seems odd. However, it is the same order as the original so fair enough

PR #303, opened 2017-03-27
Make state work after SSO login #302

jlyon:
See https://github.com/auth0/wp-auth0/issues/302

PR #388, opened 2018-01-31
Make state work after SSO login

jlyon:
This pull request rebases and replaces https://github.com/auth0/wp-auth0/pull/303.

I'm using this to redirect to a specific page after logging in via an Auth0 SSO link:
```
https://uuid.auth0.com/authorize/?connection_id=google-oauth2&client_id=xxx&redirect_uri=https://mywordpresssite.com/index.php?auth0=1&response_type=code&state=https%3A%2F%2Fmywordpresssite.com%2Fwp-admin%2Fedit.php%3Fpost_type%3Dpost
```

### DIFF
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -202,6 +202,7 @@ class WP_Auth0 {
 		$qvars[] = 'a0_action';
 		$qvars[] = 'auth0';
 		$qvars[] = 'code';
+		$qvars[] = 'state';
 		return $qvars;
 	}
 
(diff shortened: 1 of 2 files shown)


joshcanhelp:
@jlyon - Thanks for updating this. Can you walk me quickly through the steps to reproduce this? Also, if you can export your settings and paste them here, that would be helpful as well.

joshcanhelp:
@jlyon - Checking back in here ... the reason I asked about your setup is that `state` is not guaranteed to be a URL, in fact it's not set as a URL anywhere in the plugin (we're using an object). How we're handling that needs to improve (see WIP #398) but I don't think this is the way. 

More about `state` here:

https://auth0.com/docs/protocols/oauth2/oauth-state

I'm going to close this one out but check back in if you can help us understand how or why this works for you.

PR #389, opened 2018-01-31
Improved setup wizard client create process

joshcanhelp:
* Refactored `WP_Auth0_Api_Client::create_client()` to improve the URLs being saved by default and combine create with update (`web_origins` can now be pushed during create)
* Fix for blank site title causing client create to fail [reported here](https://wordpress.org/support/topic/auth0-setup-wizard-fails/)
* Helper functions on `WP_Auth0_DBManager` to grab correct links in the correct format

### DIFF
--- a/lib/WP_Auth0_Lock10_Options.php
+++ b/lib/WP_Auth0_Lock10_Options.php
@@ -33,7 +33,7 @@ class WP_Auth0_Lock10_Options {
   }
 
   public function get_code_callback_url() {
-    $protocol = $this->_get_boolean( $this->wp_options->get( 'force_https_callback' ) ) ? 'https' : '';
+    $protocol = $this->_get_boolean( $this->wp_options->get( 'force_https_callback' ) ) ? 'https' : null;
     return $this->wp_options->get_wp_auth0_url( $protocol );
   }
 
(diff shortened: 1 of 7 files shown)


joshcanhelp:
@glena - Main thing here is just making sure I'm using the right default URLs in `create_client`

lib/WP_Auth0_Lock10_Options.php
```diff
@@ -33,9 +33,8 @@ public function get_lock_show_method() {
   }
 
   public function get_code_callback_url() {
-    $protocol = $this->_get_boolean( $this->wp_options->get( 'force_https_callback' ) ) ? 'https' : null;
-
-    return site_url( 'index.php?auth0=1', $protocol );
+    $protocol = $this->_get_boolean( $this->wp_options->get( 'force_https_callback' ) ) ? 'https' : '';
```
cocojoe: Why '' vs null?

cocojoe:
Please test 😄  Might be worth adding some basic tests at least to ensure all the various URL generators/modifiers are valid.

lib/WP_Auth0_Lock10_Options.php
glena: it is used later `$site_url = empty( $protocol ) ? site_url( 'index.php' ) : site_url( 'index.php', $protocol );`.

Honestly it is the same but would prefer null too

joshcanhelp: @cocojoe @glena - I agree, change made!

joshcanhelp:
@glena - Let me know if anything else stands out to you

Issue #386, opened 2018-01-29
[ASK for help] Failed cross origin authentication	


S0c5:
#  Failed cross origin authentication	

## description

Hello guy!, when I'm trying to login in with auth0 using the defaults settings created by the wizard application and after did the sign up  WordPress sends me  this error, please

* could you clarify this one? 
* what did I do bad? 
## How to replicate? 

Please use a default and blank account for auth0 and clean WordPress installation with auth0 plugin, setup the wizard plugin and after that will appear the error. 

##  Error 

```JSON
{"auth0":"1","error":"invalid_request","error_description":"No verifier returned from client.","state":"HIDDEN_STATE"} 

```

## What did I do for figure out the error? 

I follow the back trace of the error until the rediect() function

### Test server

you could quickly test the bug. 

https://wps.pokemapa.xyz/wp-admin

* username: admin2
* password: admin

joshcanhelp:
Hi @S0c5,

Thanks for the detailed report here and apologies for the trouble on this. 

Unfortunately, your replication steps aren't going to work on our end since that's the very least we test before new versions get released. I've walked through a new installation 3 times already today and many more before the last release. Not to say this is not a problem but I'm wondering if it's a config issue on your end. 

So, a few things:

* Are you on the latest version of the plugin (3.5.1)? Latest version of WordPress? Tried deactivating other plugins first?
* Sounds like you're using the default settings but, just to make sure ... anything changed after you installed? 
* Your title is "Failed cross origin authentication" but the error you pasted here is different. Can you clarify? 
* Can you walk through your Auth0 configuration and make sure it matches here:
https://auth0.com/docs/cms/wordpress/configuration

Thanks!

S0c5:
Hi @joshcanhelp,

Answers:

1) yes I'm using the 3.5.1 version of the plugin.
2) Yeah, I didn't do any other change into auth0 configurations.
3) Yeah,  I figure out that error name into the dashboard auth0 in logs section:
 

**Dashboard**:

Type | Failed cross origin authentication
Description  | No verifier returned from client.

 **response server***
```json
{"auth0":"1","error":"invalid_request","error_description":"No verifier returned from client.","state":"HIDDEN_STATE"} 
```
4) yes I walked around the guide and I did all of those steps, but the error continue appearing. 



[...]

PR #390, opened 2018-01-31
added configurationBaseUrl for allow Cross-Origin Authentication

S0c5:
## Cross-Origin Authentication 
in order to allow the cross origin authentication in SSO, I added the configuration BaseUrl with the url of CDN.

related to issue #386 where the third party cookies cause problems.

### DIFF
--- a/lib/WP_Auth0_Lock10_Options.php
+++ b/lib/WP_Auth0_Lock10_Options.php
@@ -264,7 +264,7 @@ class WP_Auth0_Lock10_Options {
     if ( ! $this->is_registration_enabled() ) {
       $options_obj['disableSignupAction'] = true;
     }
-
+    $options_obj['configurationBaseUrl'] =  'https://cdn.auth0.com';
     return $options_obj;
   }
 }
--- a/lib/WP_Auth0_Lock_Options.php
+++ b/lib/WP_Auth0_Lock_Options.php
@@ -240,7 +240,7 @@ class WP_Auth0_Lock_Options {
 		if ( ! $this->is_registration_enabled() ) {
 			$options_obj['disableSignupAction'] = true;
 		}
-
+		$options_obj['configurationBaseUrl'] =  'https://cdn.auth0.com';
 		return $options_obj;
 	}
 }


cocojoe:
@joshcanhelp can you review this please, I don't think it should be inserted in this manner.  If this param has value then it should be populated from the Admin panel options.

joshcanhelp:
@S0c5 - I'm not totally clear what this is solving. From the Lock readme:

https://github.com/auth0/lock#other-options

> configurationBaseUrl {String}: Overrides client settings base url. By default it uses Auth0's CDN url when domain has the format *.auth0.com. Otherwise, it uses the provided domain.

Can you help me understand what this would solve? 

Also, as Walsh mentioned, this probably isn't something we want to hardcode as it's possible to change, if needed, right in the wp-admin under **Auth0 > Settings > Advanced > Extra settings**, just drop in an object like:

`{"configurationBaseUrl": "https://cdn.auth0.com"}`

If you `var_dump`/`print_r` right where you made your change and load wp-login.php, you'll see this setting come through:

screenshot 2018-02-01 07 45 40

But, again, that value you're adding is the default.

S0c5:
hi @joshcanhelp  and @cocojoe,

In order to avoid the [limitations-of-cross-origin-authentication](https://auth0.com/docs/cross-origin-authentication#limitations-of-cross-origin-authentication) due the third party cookies, Auth0 recommends implement custom domains because in this way the cookies are no longer third-party and are not blocked by browsers, but when you are using your custom domain the auth0-lock library try to gets the client script by your custom domain and that's raise a error. 

*If you're using Embedded Lock (Lock v11), you need to use your custom domain when initializing Lock. You will also need to set the configurationBaseUrl to the appropriate CDN URL*

the value of **configurationBaseUrl**  can be changed by the region or zone: 

*The CDN URL varies by region. For regions outside of the US, use https://cdn.{region}.auth0.com.*, but by default we can use the base region.

joshcanhelp:
@S0c5 - I guess I'm still confused as to what this PR is trying to achieve. The hard-coded value is the default that's already used and a custom value can be implemented with the settings I described.

S0c5:
Hi @joshcanhelp  I just seen that you are right I can add the parameter by extra settings. Sorry for you time, I Hope contribute some day to this project,!

best regards guys.

joshcanhelp:
No problem, we appreciate the effort!

Issue #380, opened 2018-01-29
Non-static method WP_Auth0_Api_Client::convertCertToPem() should not be called statically


joshcanhelp:
WP_Auth0_Api_Client.php L:1042

PR #391, opened 2018-02-05
Cleanup PR for 3.5.2

joshcanhelp:
- Fixing convertCertToPem static error (#380) 
- Fixing admin notices when saving settings
- readme.txt spelling and complete changelog with reference to main one maintained here
- Incorrect closing tag in the Setup Wizard


lib/WP_Auth0_Api_Client.php
```diff
@@ -1039,7 +1039,16 @@ public static function update_guardian($domain, $app_token, $factor, $enabled) {
...
+	/**
+	 * Convert a certificate to PEM format
+	 *
+	 * @see https://en.wikipedia.org/wiki/Privacy-enhanced_Electronic_Mail
+	 *
+	 * @param string $cert - certificate, like from .well-known/jwks.json
+	 *
+	 * @return string
+	 */
+  protected static function convert_cert_to_pem( $cert ) {
```
cocojoe: Not sure we need to bother retiring this method and adding an alternative, it does not say public so we shouldn't have to worry about it. Also it is a very new addition so doubt anyone is using it.

cocojoe: Why does WP coding guide go against PSR-2? 😢

readme.txt
```diff
@@ -151,88 +148,148 @@ To change the form_title in this case, you need to add the following attribute t
...
-- Fixed migration for older plugins that use base64 secret [Github PR 324](https://github.com/auth0/wp-auth0/pull/324) - [cocojoe](https://github.com/cocojoe)
+[Details](https://github.com/auth0/wp-auth0/blob/master/CHANGELOG.md#3222-2017-06-26)
 
 = 3.2.21 =
 
-**Added**
-- Improve redirect_login error logging, JWT leeway [Github PR 317](https://github.com/auth0/wp-auth0/pull/317) - [cocojoe](https://github.com/cocojoe)
+[Details](https://github.com/auth0/wp-auth0/blob/master/CHANGELOG.md#3221-2017-06-14)
+
+= 3.2.5 =
```
cocojoe: It's nice to have a complete log. However, I'm not sure it adds any value going back before 3.0. Typically you only care about latest and greatest or at least back a few minors when you upgrade.
It can link off to changelog? for historical <3.x ?

readme.txt
```diff
@@ -13,9 +13,6 @@ PLUGIN_DESCRIPTION
 
 This plugin gives WordPress a new Login Widget (powered by [Auth0](https://auth0.com)) that enables:
 
-- Easy setup
-    + 5 minutes installation
```
cocojoe: 😆

lib/WP_Auth0_Api_Client.php
joshcanhelp: @cocojoe - Reverted.

readme.txt
joshcanhelp: @cocojoe - removed

PR #394, opened 2018-02-12
Fixed SSO auto-login in Lock

### LINKED ISSUE #363: SSO login failing when not using implicit flow
When you enable the **Single Sign On (SSO)** toggle in the **Features** settings of the plugin, [some client javascript is rendered](https://github.com/auth0/wp-auth0/blob/dev/templates/auth0-sso-handler-lock10.php) to `checkSession` and see if a valid SSO session is available at Auth0, to log the user in automatically without prompting for credentials.

The above flow is failing when there's a valid SSO session at Auth0, causing the plugin code to display:

> There was a problem with your log in

and logging a "Wrong number of segments" error in the internal plugin logs:

image

The `checkSession` callback does not have an `idToken` in the `authResult`, which causes the `Wrong number of segments` error described above. 

image

This is because the call to `checkSession` uses `responseType: 'code'` when implicit flow is turned off, and that response type is not supported by `checkSession`.

```javascript
  var options = <?php echo json_encode( $lock_options->get_sso_options() ); ?>;
```

results into:

```javascript
  var options = {"scope":"openid ","responseType":"code","redirectUri":"http:\/\/dev-nicotest1.pantheonsite.io\/index.php?auth0=1","state":"xxxx","nonce":"nonce"};
```

`responseType: 'code'` is not valid in `checkSession`:

https://github.com/auth0/auth0.js/blob/308d55f9f0a8c6fa2f46cd61cf7be2917a8fa647/src/web-auth/index.js#L389-L391

To reproduce:

- Make sure the WP site is working with the Auth0 plugin
- Enable the **Single Sign On (SSO)** toggle in the plugin settings
- Disable the **Implicit flow** toggle.
- Log in with the Auth0 plugin into the WP site. This will create a session both in the WP site and in Auth0.
- Clear the cookies for the WP site (but leave the ones in Auth0). This will log you out of the WP site.
- Go to `/wp-login.php` into the WP site. The rendered HTML will do a `checkSession` call (which will find a session in Auth0). You'll get "There was a problem with your log in"

joshcanhelp:
Fixes #363

* Fixed Lock options sending an incorrect responseType, preventing SSO login
* Fixing incorrect redirect URL used for SSO
* Removing related but unused template


lib/WP_Auth0_LoginManager.php
```diff
@@ -265,11 +265,11 @@ public function redirect_login() {
 
       $userinfo = json_decode( $response['body'] );
       if ( $this->login_user( $userinfo, $data->id_token, $data->access_token ) ) {
-        if ( null !== $stateFromGet && isset( $stateFromGet->interim ) && $stateFromGet->interim ) {
+        if ( ! empty( $stateFromGet->interim ) ) {
```
lbalmaceda: I guess this is behaving like a dictionary, right?

```js
stateFromGet = { 
   interim: "some value",
   //....
}
```

In that case, could $stateFromGet be null / undefined and the call to obtain the value fail/throw? `$stateFromGet->interim`

lib/WP_Auth0_LoginManager.php
```diff
@@ -265,11 +265,11 @@ public function redirect_login() {
...
           include WPA0_PLUGIN_DIR . 'templates/login-interim.php';
           exit();
         } else {
-          if ( null !== $stateFromGet && isset( $stateFromGet->redirect_to ) ) {
+          if ( ! empty( $stateFromGet->redirect_to ) && wp_login_url() !== $stateFromGet->redirect_to ) {
             $redirectURL = $stateFromGet->redirect_to;
           } else {
             $redirectURL = $this->a0_options->get( 'default_login_redirection' );
```
lbalmaceda: Is the next line equal? (without the use of `get`)

```php
$redirectURL = $this->a0_options->default_login_redirection;
```

templates/auth0-sso-handler-lock10.php
```diff
@@ -13,19 +13,20 @@
...
 
   var options = <?php echo json_encode( $lock_options->get_sso_options() ); ?>;
+  options.responseType = 'token id_token';
   webAuth.checkSession(options
   , function (err, authResult) {
       if (typeof(authResult) === 'undefined') {
         return;
       }
 
       if (typeof(authResult.code) !== 'undefined') {
```
lbalmaceda: will this check fail if `authResult.code` is `null`?

templates/auth0-sso-handler-lock10.php
```diff
@@ -13,19 +13,20 @@
...
-        window.location = '<?php echo add_query_arg( 'auth0', 1, site_url() ); ?>&code=' + authResult.code +
-            '&state=' + authResult.state;
+        window.location = '<?php echo WP_Auth0_Options::Instance()->get_wp_auth0_url(); ?>&code='
+            + authResult.code + '&state=' + authResult.state;
       } else if (typeof(authResult.idToken) !== 'undefined') {
         jQuery(document).ready(function($){
           var $form=$(document.createElement('form')).css({display:'none'}).attr("method","POST").attr("action","<?php
-            echo add_query_arg( 'auth0', 'implicit', site_url() ); ?>");
+            echo add_query_arg( 'auth0', 'implicit', site_url( 'index.php' ) ); ?>");
```
lbalmaceda: what's `index.php`, and why you need to hardcode it here?

lib/WP_Auth0_LoginManager.php
joshcanhelp: @lbalmaceda - `empty()` works in the case where the base object/array is undefined so:

screenshot 2018-02-15 08 40 03

lib/WP_Auth0_LoginManager.php
joshcanhelp: @lbalmaceda - I'm not sure what you're asking ...

lbalmaceda: I'm asking if 

```php
$redirectURL = $this->a0_options->get( 'default_login_redirection' );
```

is the same as saying

```php
$redirectURL = $this->a0_options->default_login_redirection;
```

lib/WP_Auth0_LoginManager.php
lbalmaceda: Ok, that's for `empty()`. But my question is: 
What if `$stateFromGet` is null/undefined. If you do `$stateFromGet->interim` will that throw an unhandled exception? (this operation is done before actually calling `empty()` as you need to resolve the parameter for the function)

lib/WP_Auth0_LoginManager.php
joshcanhelp: No magic getters, no

lib/WP_Auth0_LoginManager.php
joshcanhelp: Nope, `empty()` handles it:

screenshot 2018-02-15 09 15 20

templates/auth0-sso-handler-lock10.php
joshcanhelp: The main WP entry point. I've tested without and it had issues

templates/auth0-sso-handler-lock10.php
joshcanhelp: Potentially but there is an `idToken` fallback:

https://github.com/auth0/wp-auth0/blob/master/templates/auth0-sso-handler-lock10.php#L22

lbalmaceda: And what if none of the `if` branches are fulfilled ? Shouldn't you handle this gracefully?

joshcanhelp: Probably? But this is trying to auto-login with SSO so falling through = login form. Technically graceful? :)

cocojoe:
Can you add test setup overview please.

joshcanhelp:
@cocojoe - I tested using Nico's instructions in #363 - login, clear WP cookies, visit wp-login.php - and that works well. Then I setup a second client for a second, public WP site, set it to use the same database, and tested went to wp-login.php and it worked as expected.

PR #406, opened 2018-03-02
State handling during login process for both types

joshcanhelp:
The WordPress plugin doesn't not currently handle state checking at all. A value is set in Lock and sent back but not stored and checked. This PR:

- Adds a `WP_Auth0_State_Handler` class used to generate, store, and validate a state nonce using cookies
- Adjusts Lock to use the new nonce value and stores it as a cookie
- Checks for this value during login for both flows, redirect and implicit
- Adds a `WP_Auth0_LoginManager->die_on_login()` method to handling error message display for login errors
- Adjusts the `auth0-singlelogout-handler.php` to not pull Lock options and skip anything related to state

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -162,6 +162,12 @@ class WP_Auth0_LoginManager {
       return;
     }
 
+    // Check for valid state UUID
+    $state_decoded = json_decode( base64_decode( $this->query_vars( 'state' ) ), TRUE );
+    if ( empty( $state_decoded[ 'nonce' ] ) || ! WP_Auth0_State_Handler::validate( $state_decoded[ 'nonce' ] ) ) {
+      $this->die_on_login( __( 'Invalid state', 'wp-auth0' ) );
+    }
+
     try {
       if ( $this->query_vars( 'auth0' ) === 'implicit' ) {
         $this->implicit_login();
@@ -170,19 +176,11 @@ class WP_Auth0_LoginManager {
       }
     } catch (WP_Auth0_LoginFlowValidationException $e) {
 
-      $msg = __( 'There was a problem with your log in. ', 'wp-auth0' );
-      $msg .= ' '. $e->getMessage();
-      $msg .= '<br/><br/>';
-      $msg .= '<a href="' . wp_login_url() . '">' . __( '← Login', 'wp-auth0' ) . '</a>';
-      wp_die( $msg );
+      $this->die_on_login( $e->getMessage(), $e->getCode() );
 
     } catch (WP_Auth0_BeforeLoginException $e) {
 
-      $msg = __( 'You have logged in successfully, but there is a problem accessing this site. ', 'wp-auth0' );
-      $msg .= ' '. $e->getMessage();
-      $msg .= '<br/><br/>';
-      $msg .= '<a href="' . wp_logout_url() . '">' . __( '← Logout', 'wp-auth0' ) . '</a>';
-      wp_die( $msg );
+      $this->die_on_login( $e->getMessage(), $e->getCode(), FALSE );
 
     } catch (Exception $e) {
 
@@ -206,9 +204,7 @@ class WP_Auth0_LoginManager {
     }
 
     $code = $this->query_vars( 'code' );
-    $state = $this->query_vars( 'state' );
-
-    $stateFromGet = json_decode( base64_decode( $state ) );
+    $state_decoded = json_decode( base64_decode( $this->query_vars( 'state' ) ), TRUE );
 
     $domain = $this->a0_options->get( 'domain' );
 
@@ -265,12 +261,12 @@ class WP_Auth0_LoginManager {
 
       $userinfo = json_decode( $response['body'] );
       if ( $this->login_user( $userinfo, $data->id_token, $data->access_token ) ) {
-        if ( ! empty( $stateFromGet->interim ) ) {
+        if ( ! empty( $state_decoded->interim ) ) {
           include WPA0_PLUGIN_DIR . 'templates/login-interim.php';
           exit();
         } else {
-          if ( ! empty( $stateFromGet->redirect_to ) && wp_login_url() !== $stateFromGet->redirect_to ) {
-            $redirectURL = $stateFromGet->redirect_to;
+          if ( ! empty( $state_decoded->redirect_to ) && wp_login_url() !== $state_decoded->redirect_to ) {
+            $redirectURL = $state_decoded->redirect_to;
           } else {
             $redirectURL = $this->a0_options->get( 'default_login_redirection' );
           }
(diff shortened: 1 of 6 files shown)


lib/WP_Auth0_LoginManager.php
```diff
@@ -66,19 +66,6 @@ public function auth0_singlelogout_footer( $previous_html ) {
...
-      return;
-    }
-
-    $lock_options = new WP_Auth0_Lock10_Options();
-    $cdn = $this->a0_options->get('auth0js-cdn');
-    $client_id = $this->a0_options->get( 'client_id' );
-    $domain = $this->a0_options->get( 'domain' );
-    $logout_url = wp_logout_url( get_permalink() ) . '&SLO=1';
-
     include WPA0_PLUGIN_DIR . 'templates/auth0-singlelogout-handler.php';
```
joshcanhelp: Removed vars above were moved into `templates/auth0-singlelogout-handler.php` down below

templates/auth0-singlelogout-handler.php
```diff
@@ -1,28 +1,29 @@
...
+      domain:'<?php echo sanitize_text_field( $this->a0_options->get( 'domain' ) ); ?>'
     });
 
-    var options = <?php echo json_encode( $lock_options->get_sso_options() ); ?>;
-    options.responseType = 'token id_token';
-    webAuth.checkSession(options, function (err, authResult) {
-      if (err !== null) {
-        if(err.error ==='login_required') {
-          window.location = '<?php echo html_entity_decode( $logout_url ); ?>';
+    webAuth.checkSession( { 'responseType' : 'token', 'redirectUri' : window.location.href }, function ( err ) {
```
joshcanhelp: This change might seem a little out-of-scope but this template file was pulling Lock options which was, in turn, generating a state nonce and trying to store it in cookies. `checkSession` only needs a couple of options and pulling all the Lock ones and [then overriding some](https://github.com/auth0/wp-auth0/blob/master/templates/auth0-singlelogout-handler.php#L17) was confusing.

lib/WP_Auth0_State_Handler.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+class WP_Auth0_State_Handler {
+
+  /**
+   * @var string
+   */
+  protected $uniqid = '';
```
lbalmaceda: If this is being set in the `__construct()` method (I guess the class instance constructor) there shouldn't be a need to set it to `''` here.

templates/auth0-singlelogout-handler.php
```diff
@@ -1,28 +1,29 @@
...
+            if ( err && err.error ==='login_required' ) {
+                window.location = '<?php echo esc_url( wp_logout_url( get_permalink() ) . '&SLO=1' ); ?>';
```
lbalmaceda: What is the `SLO=1` query param added at the end of the url?

joshcanhelp: Takes the `redirect_to` query param (the argument in `wp_logout_url()`) and redirects after running the logout process:

https://github.com/auth0/wp-auth0/blob/master/lib/WP_Auth0_LoginManager.php#L93

lib/WP_Auth0_LoginManager.php
```diff
@@ -162,6 +149,12 @@ public function init_auth0() {
       return;
     }
 
+    // Check for valid state UUID
+    $state_decoded = json_decode( base64_decode( $this->query_vars( 'state' ) ), TRUE );
```
lbalmaceda: what if `state` is missing from `query_vars`. How will the code behave?

joshcanhelp: Then `empty( $state_decoded[ 'nonce' ] )` will be true and the login process will fail

lib/WP_Auth0_LoginManager.php
```diff
@@ -170,19 +163,11 @@ public function init_auth0() {
       }
     } catch (WP_Auth0_LoginFlowValidationException $e) {
 
-      $msg = __( 'There was a problem with your log in. ', 'wp-auth0' );
-      $msg .= ' '. $e->getMessage();
-      $msg .= '<br/><br/>';
-      $msg .= '<a href="' . wp_login_url() . '">' . __( '← Login', 'wp-auth0' ) . '</a>';
-      wp_die( $msg );
+      $this->die_on_login( $e->getMessage(), $e->getCode() );
```
lbalmaceda: This 2 `die_on_login` calls only differ on the boolean you pass at the end.. This uses the default TRUE, meaning is "LOGIN_LINK" (idk what that means). The other passes FALSE, so its "LOGOUT_LINK" (again idk). How is that an exception type on the same "try logic" could tell you that?

lbalmaceda: Do you have an internal exception from when the user needs to log in / log out? In that case I think I understand now, and depending on the bool you are adding the link to log in / log out on the message.

lib/WP_Auth0_State_Handler.php
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * @var string
+   */
+  const COOKIE_NAME = 'auth0_uniqid';
+
+  /**
+   * @var int
+   */
+  protected $cookieExpiresIn = MINUTE_IN_SECONDS;
```
lbalmaceda: Where does this value comes from?

lib/WP_Auth0_LoginManager.php
joshcanhelp: That whole `try` block needs attention in another PR but, as I understand the intended flow, there are exceptions that are thrown during the login process, `WP_Auth0_LoginFlowValidationException`, and exceptions thrown once the login flow is complete but before you're actually logged in, `WP_Auth0_BeforeLoginException`. That `bool` determines what the messaging is.

templates/auth0-singlelogout-handler.php
lbalmaceda: To me this looks as a good call. 👏

templates/auth0-singlelogout-handler.php
lbalmaceda: I don't follow. This is an "internal" query param that you later use to know if you have to redirect?

lib/WP_Auth0_LoginManager.php
lbalmaceda: So you are saying this calls `json_decode( base64_decode( $this->query_vars( 'state' ) ), TRUE );` don't fail when the `query_vars.state` is missing/null/undefined, right?

lib/WP_Auth0_State_Handler.php
joshcanhelp: WP core:

https://github.com/WordPress/WordPress/blob/master/wp-includes/default-constants.php#L129

lib/WP_Auth0_LoginManager.php
joshcanhelp: Exactly. `$this->query_vars( 'state' )` returns `null` if not found, `base64_decode()` returns `false` in that case, and `json_decode()` returns `false`. PHP's `empty()` returns `true` even if that key does not exist (no errors).

templates/auth0-singlelogout-handler.php
joshcanhelp: Basically, yes. It sends it through the regular logout process, then that param tells the process to look for the redirect param.

lbalmaceda: uhm Ok. Sounds hacky and I'm not going to block you for this if it works. But it should probably be reviewed in the future 🐔

lib/WP_Auth0_LoginManager.php
lbalmaceda: Perfect, thanks.

templates/auth0-singlelogout-handler.php
joshcanhelp: Logout process is getting a once-over in another PR fir this same release. You'll be able to sleep through the night soon enough 💤

lib/WP_Auth0_LoginManager.php
```diff
@@ -537,6 +522,45 @@ protected function query_vars( $key ) {
     return null;
   }
 
+  /**
+   * Get and store state returned from Auth0
+   *
+   * @return object|null
+   */
+  protected function get_state() {
```
joshcanhelp: @lbalmaceda - FYI, I just added this method and it's usage above (using this instead of always parsing the state parameter). It didn't invalidate your review but thought I would point it out.

PR #409, opened 2018-03-09
Fix Login Process Error Handling

### LINKED ISSUE #305: "Invalid authorization code": Access token is requested twice in a row, breaking the login flow
I am using Wordpress Version 4.7.3, trying to install the official Auth0 plugin on it for the first time.

The problem is related to logging in as an Auth0 user. After a successful login Auth0 returns the user agent to https://site.com/index.php?auth0=1, which receives the auth code and uses it to get an access token - that's the usual pattern. In my case, however, the Wordpress site automatically redirects from /index.php to / (root), but only AFTER running the plugin token logic. So what actually happened was:
1. User logs in, Auth0 redirects to /index.php?auth0=1
2. Wordpress receives the code at /index.php?auth0=1, gets an access token, THEN it redirects to / (root).
3. Wordpress receives the code again, at / (root) this time, tries to get a new access token, but fails because the auth code was used already.

I don't know if this is a WP bug, perhaps related to a newer version of WP, but it is what it is. I fixed this by replacing every occurance of "index.php" in the source of the plugin with emptiness, and fixing the callbacks in Auth0 to redirect to /?auth0=1 instead of /index.php?auth0=1. 

Maybe you can find a cleaner fix, perhaps a switch or an input field in the dashboard that lets the site owner choose the callback destination.

joshcanhelp:
WP_Auth0_LoginManager was not processing errors well, was poorly
documented, and might have been improperly exposing error messages.
Incoming URL param errors from Auth0 and configuration issues are caught
earlier in the login process. Error message are not exposed to the
user; instead they are logged for an admin. Thrown errors are
standarized and listed in docblocks.

Fixes #305

### DIFF
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -124,13 +124,13 @@ class WP_Auth0 {
 	}
 
   /**
-   * Is the Auth0 plugin ready to proc
+   * Is the Auth0 plugin ready to process logins?
    *
    * @return bool
    */
   public static function ready() {
     $options = WP_Auth0_Options::Instance();
-    if ( ! $options->get( 'domain' ) || ! $options->get( 'client_id' ) || ! $options->get( 'client_id' ) ) {
+    if ( ! $options->get( 'domain' ) || ! $options->get( 'client_id' ) || ! $options->get( 'client_secret' ) ) {
       return FALSE;
     }
     return TRUE;
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -123,6 +123,19 @@ class WP_Auth0 {
 		WP_Auth0_Email_Verification::init();
 	}
 
+  /**
+   * Is the Auth0 plugin ready to proc
+   *
+   * @return bool
+   */
+  public static function ready() {
+    $options = WP_Auth0_Options::Instance();
+    if ( ! $options->get( 'domain' ) || ! $options->get( 'client_id' ) || ! $options->get( 'client_id' ) ) {
+      return FALSE;
+    }
+    return TRUE;
+  }
+
 	/**
 	 * Checks it it should update the database connection no enable or disable signups and create or delete
 	 * the rule that will disable social signups.
@@ -240,9 +253,11 @@ class WP_Auth0 {
 	}
 
 	public function a0_register_query_vars( $qvars ) {
+		$qvars[] = 'error';
 		$qvars[] = 'error_description';
 		$qvars[] = 'a0_action';
 		$qvars[] = 'auth0';
+		$qvars[] = 'state';
 		$qvars[] = 'code';
 		$qvars[] = 'state';
 		return $qvars;
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -231,10 +231,9 @@ class WP_Auth0_LoginManager {
     if ( empty( $data->access_token ) ) {
 
       // Look for clues as to what went wrong
-      throw new WP_Auth0_LoginFlowValidationException(
-        ! empty( $data->error_description ) ? $data->error_description : __( 'Unknown error', 'wp-auth0' ),
-        ! empty( $data->error ) ? $data->error : $exchange_resp_code
-      );
+      $e_message = ! empty( $data->error_description ) ? $data->error_description : __( 'Unknown error', 'wp-auth0' );
+      $e_code = ! empty( $data->error ) ? $data->error : $exchange_resp_code;
+      throw new WP_Auth0_LoginFlowValidationException( $e_message, $e_code );
     }
 
     // Decode our incoming ID token for the Auth0 user_id
@@ -266,7 +265,10 @@ class WP_Auth0_LoginManager {
       // TODO: fallback to /userinfo with access token
 
       WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__ . ' => WP_Auth0_Api_Client::get_user()', $userinfo_resp );
-      throw new WP_Auth0_LoginFlowValidationException( __( 'Error getting user information', 'wp-auth0' ) );
+      throw new WP_Auth0_LoginFlowValidationException(
+        __( 'Error getting user information', 'wp-auth0' ),
+        $userinfo_resp_code
+      );
     }
 
     $userinfo = json_decode( $userinfo_resp_body );
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -211,10 +211,10 @@ class WP_Auth0_LoginManager {
       // Not authorized
       WP_Auth0_ErrorManager::insert_auth0_error(
         __METHOD__ . ' L:' . __LINE__,
-        __( 'An oauth/token call triggered a 401 response from Auth0. ', 'wp-auth0' ) .
+        __( 'An /oauth/token call triggered a 401 response from Auth0. ', 'wp-auth0' ) .
           __( 'Please check the Client Secret saved in the Auth0 plugin settings. ', 'wp-auth0' )
       );
-      throw new WP_Auth0_LoginFlowValidationException( __( 'Not Authorized', 'wp-auth0' ), $exchange_resp_code );
+      throw new WP_Auth0_LoginFlowValidationException( __( 'Not Authorized', 'wp-auth0' ), 401 );
 
     } else if ( empty( $exchange_resp_body ) ) {
 
@@ -233,31 +233,39 @@ class WP_Auth0_LoginManager {
       // Look for clues as to what went wrong
       throw new WP_Auth0_LoginFlowValidationException(
         ! empty( $data->error_description ) ? $data->error_description : __( 'Unknown error', 'wp-auth0' ),
-        ( ! empty( $data->error ) ? $data->error : $exchange_resp_code )
+        ! empty( $data->error ) ? $data->error : $exchange_resp_code
       );
     }
 
+    // Decode our incoming ID token for the Auth0 user_id
     $decoded_token = JWT::decode(
       $data->id_token,
       $this->a0_options->get_client_secret_as_key(),
       array( $this->a0_options->get_client_signing_algorithm() )
     );
 
-    $userinfo_resp = WP_Auth0_Api_Client::get_user(
-      $this->a0_options->get( 'domain' ),
-      WP_Auth0_Api_Client::get_client_token(),
-      $decoded_token->sub
-    );
+    // Attempt to authenticate with the Management API
+    $client_credentials_token = WP_Auth0_Api_Client::get_client_token();
+    $userinfo_resp = null;
+
+    if ( $client_credentials_token ) {
+
+        $userinfo_resp = WP_Auth0_Api_Client::get_user(
+            $this->a0_options->get( 'domain' ),
+            $client_credentials_token,
+            $decoded_token->sub
+        );
+    }
 
     $userinfo_resp_code = (int) wp_remote_retrieve_response_code( $userinfo_resp );
     $userinfo_resp_body = wp_remote_retrieve_body( $userinfo_resp );
 
-    if ( $userinfo_resp instanceof WP_Error || 200 !== $userinfo_resp_code || empty( $userinfo_resp_body ) ) {
+    // Management API call failed
+    if ( 200 !== $userinfo_resp_code || empty( $userinfo_resp_body ) ) {
 
-      // Management API call failed
-      // TODO: fallback to /userinfo
+      // TODO: fallback to /userinfo with access token
+
       WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__ . ' => WP_Auth0_Api_Client::get_user()', $userinfo_resp );
-      error_log( $userinfo_resp->get_error_message() );
       throw new WP_Auth0_LoginFlowValidationException( __( 'Error getting user information', 'wp-auth0' ) );
     }
 
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -144,10 +144,17 @@ class WP_Auth0_LoginManager {
   }
 
   public function init_auth0() {
-    global $wp_query;
 
-    if ( $this->query_vars( 'auth0' ) === null ) {
+    // Not an Auth0 login process or settings are not configured to allow logins
+    if ( ! $this->query_vars( 'auth0' ) || ! WP_Auth0::ready() ) {
       return;
+    }
+
+    // Catch any incoming errors and stop the login process
+    // See https://auth0.com/docs/libraries/error-messages
+    if ( $this->query_vars( 'error_description' ) ) {
+      $error_msg = sanitize_text_field( $this->query_vars( 'error_description' ) );
+      $this->die_on_login( $error_msg ? $error_msg : sanitize_text_field( $this->query_vars( 'error' ) ) );
     }
 
     // Check for valid state nonce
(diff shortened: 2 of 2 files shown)


lib/WP_Auth0_LoginManager.php
```diff
@@ -144,12 +144,19 @@ public function login_auto() {
   }
 
   public function init_auth0() {
-    global $wp_query;
```
joshcanhelp: Not used

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
+      // Errors during the OAuth login flow
       $this->die_on_login( $e->getMessage(), $e->getCode() );
 
-    } catch (WP_Auth0_BeforeLoginException $e) {
+    } catch ( WP_Auth0_BeforeLoginException $e ) {
 
+      // Errors during the WordPress login flow
       $this->die_on_login( $e->getMessage(), $e->getCode(), FALSE );
-
-    } catch (Exception $e) {
```
joshcanhelp: Empty catch block was catching errors from other plugins

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
+  /**
+   * Main login flow, Authorization Code Grant
+   *
+   * @throws WP_Auth0_BeforeLoginException
+   * @throws WP_Auth0_LoginFlowValidationException
+   *
+   * @see https://auth0.com/docs/api-auth/tutorials/authorization-code-grant
+   */
   public function redirect_login() {
-    global $wp_query;
```
joshcanhelp: Not used

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
-
-    if ( $this->query_vars( 'auth0' ) === null ) {
```
joshcanhelp: These 3 `if` blocks were moved to `init_auth0()` to be caught earlier and for both redirect and implicit flows.

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
-    if ( $this->query_vars( 'error_description' ) !== null && $this->query_vars( 'error_description' ) !== '' ) {
-      throw new WP_Auth0_LoginFlowValidationException( sanitize_text_field( $this->query_vars( 'error_description' ) ) );
-    }
-
-    if ( $this->query_vars( 'error' ) !== null && trim( $this->query_vars( 'error' ) ) !== '' ) {
-      throw new WP_Auth0_LoginFlowValidationException( sanitize_text_field( $this->query_vars( 'error' ) ) );
-    }
-
-    $code = $this->query_vars( 'code' );
-    $state_decoded = $this->get_state();
```
joshcanhelp: Moved down

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
-
-    $response = WP_Auth0_Api_Client::get_token( $domain, $client_id, $client_secret, 'authorization_code', array(
+    // Exchange authorization code for token
+    $exchange_resp = WP_Auth0_Api_Client::get_token( $domain, $client_id, $client_secret, 'authorization_code', array(
         'redirect_uri' => home_url(),
-        'code' => $code,
+        'code' => $this->query_vars( 'code' ),
       ) );
 
-    if ( $response instanceof WP_Error ) {
```
joshcanhelp: Diff gets a little mangled from here down because of switching around how the errors are handled. This was removed in favor of checking a specific error code we're looking for, followed by a general check if we have the info we need (`access_token`)

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
 
-      throw new WP_Auth0_LoginFlowValidationException( $response->get_error_message() );
-    }
-
-    $data = json_decode( $response['body'] );
+    if ( 401 === $exchange_resp_code ) {
 
-    if ( isset( $data->access_token ) || isset( $data->id_token ) ) {
+      // Not authorized
+      WP_Auth0_ErrorManager::insert_auth0_error(
```
joshcanhelp: Log but don't show to the end user

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
+        __METHOD__ . ' L:' . __LINE__,
+        __( 'An oauth/token call triggered a 401 response from Auth0. ', 'wp-auth0' ) .
+          __( 'Please check the Client Secret saved in the Auth0 plugin settings. ', 'wp-auth0' )
+      );
+      throw new WP_Auth0_LoginFlowValidationException( __( 'Not Authorized', 'wp-auth0' ), $exchange_resp_code );
 
-	    $decoded_token = JWT::decode(
```
joshcanhelp: Moved down

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
-		    $data->id_token,
-		    $this->a0_options->get_client_secret_as_key(),
-		    array( $this->a0_options->get_client_signing_algorithm() )
-	    );
+    } else if ( empty( $exchange_resp_body ) ) {
 
-	    $data->id_token = null;
-	    $response = WP_Auth0_Api_Client::get_user(
```
joshcanhelp: Moved down

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
-		    $this->a0_options->get( 'domain' ),
-		    WP_Auth0_Api_Client::get_client_token(),
-		    $decoded_token->sub
-	    );
+      // Unsuccessful for another reason
```
joshcanhelp: Log the error if we've got something but don't show to end user

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
-        WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__ . ' => WP_Auth0_Api_Client::get_user()', $response );
+      throw new WP_Auth0_LoginFlowValidationException( __( 'Unknown error', 'wp-auth0' ), $exchange_resp_code );
+    }
 
-        error_log( $response->get_error_message() );
+    $data = json_decode( $exchange_resp_body );
 
-        throw new WP_Auth0_LoginFlowValidationException( );
-      }
+    if ( empty( $data->access_token ) ) {
```
joshcanhelp: Only need the access token here

lib/WP_Auth0_LoginManager.php
joshcanhelp: Blank general `Exception` catching lead to issues like #305

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
-
-    $code = $this->query_vars( 'code' );
-    $state_decoded = $this->get_state();
 
     $domain = $this->a0_options->get( 'domain' );
 
     $client_id = $this->a0_options->get( 'client_id' );
     $client_secret = $this->a0_options->get( 'client_secret' );
 
-    if ( empty( $client_id ) ) {
```
joshcanhelp: Handled earlier in `init_auth0()` by `WP_Auth0::ready()`

lib/WP_Auth0_LoginManager.php
joshcanhelp: Moved down

lib/WP_Auth0_LoginManager.php
joshcanhelp: Moved down

WP_Auth0.php
```diff
@@ -123,6 +123,19 @@ public function init() {
 		WP_Auth0_Email_Verification::init();
 	}
 
+  /**
+   * Is the Auth0 plugin ready to proc
```
lbalmaceda: if "proc" is the short for something I didn't get it. To process, maybe?

WP_Auth0.php
```diff
@@ -123,6 +123,19 @@ public function init() {
...
+   *
+   * @return bool
+   */
+  public static function ready() {
+    $options = WP_Auth0_Options::Instance();
+    if ( ! $options->get( 'domain' ) || ! $options->get( 'client_id' ) || ! $options->get( 'client_id' ) ) {
```
lbalmaceda: duplicated `client_id` condition

WP_Auth0.php
```diff
@@ -240,9 +253,11 @@ public static function get_plugin_dir_url() {
 	}
 
 	public function a0_register_query_vars( $qvars ) {
```
lbalmaceda: what's the purpose of this function?

lib/WP_Auth0_LoginManager.php
```diff
@@ -144,12 +144,19 @@ public function login_auto() {
   }
 
   public function init_auth0() {
-    global $wp_query;
 
-    if ( $this->query_vars( 'auth0' ) === null ) {
+    // Not an Auth0 login process or settings are not configured to allow logins
+    if ( ! $this->query_vars( 'auth0' ) || ! WP_Auth0::ready() ) {
```
lbalmaceda: `auth0` was the query param you use to track context, right?

lib/WP_Auth0_LoginManager.php
```diff
@@ -144,12 +144,19 @@ public function login_auto() {
...
       return;
     }
 
+    // Catch any incoming errors and stop the login process
+    // See https://auth0.com/docs/libraries/error-messages
+    if ( $this->query_vars( 'error_description' ) ) {
+      $error_msg = sanitize_text_field( $this->query_vars( 'error_description' ) );
+      $this->die_on_login( $error_msg ? $error_msg : sanitize_text_field( $this->query_vars( 'error' ) ) );
```
lbalmaceda: the `if` already checks if `error_description` is present on the query. Is there any case where the `sanitize_test_field(error_description)` will fail or give you something null/undefined? if not, it doesn't make sense to have this line's `if` that calls `sanitize_text_field(error)`

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
-    if ( empty( $domain ) ) {
-      throw new WP_Auth0_LoginFlowValidationException( __( 'Error: No Domain defined in Wordpress Administration!', 'wp-auth0' ) );
-    }
-
-    $response = WP_Auth0_Api_Client::get_token( $domain, $client_id, $client_secret, 'authorization_code', array(
+    // Exchange authorization code for token
+    $exchange_resp = WP_Auth0_Api_Client::get_token( $domain, $client_id, $client_secret, 'authorization_code', array(
         'redirect_uri' => home_url(),
-        'code' => $code,
+        'code' => $this->query_vars( 'code' ),
```
lbalmaceda: is there a case where `code` might be missing from the query_vars and this `get_token` call fails/leads to an unhandled exception? Or even then if the code is missing there's no use on calling the method at all.

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
-    }
-
-    $data = json_decode( $response['body'] );
+    if ( 401 === $exchange_resp_code ) {
 
-    if ( isset( $data->access_token ) || isset( $data->id_token ) ) {
+      // Not authorized
+      WP_Auth0_ErrorManager::insert_auth0_error(
+        __METHOD__ . ' L:' . __LINE__,
+        __( 'An oauth/token call triggered a 401 response from Auth0. ', 'wp-auth0' ) .
```
lbalmaceda: `/oauth/token`

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
+          __( 'Please check the Client Secret saved in the Auth0 plugin settings. ', 'wp-auth0' )
+      );
+      throw new WP_Auth0_LoginFlowValidationException( __( 'Not Authorized', 'wp-auth0' ), $exchange_resp_code );
```
lbalmaceda: Since `$exchange_resp_code` is always `401` inside this if, I'd hardcode it. But it's ok either way for me 👍

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
-          exit();
-        } else {
-          if ( ! empty( $state_decoded->redirect_to ) && wp_login_url() !== $state_decoded->redirect_to ) {
-            $redirectURL = $state_decoded->redirect_to;
-          } else {
-            $redirectURL = $this->a0_options->get( 'default_login_redirection' );
-          }
+      // Look for clues as to what went wrong
+      throw new WP_Auth0_LoginFlowValidationException(
+        ! empty( $data->error_description ) ? $data->error_description : __( 'Unknown error', 'wp-auth0' ),
```
lbalmaceda: Can you explain the line below this one please? I'm lost after the comma at the end.

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
+      );
+    }
 
-          wp_safe_redirect( $redirectURL );
-        }
-      }
-    } elseif ( is_array( $response['response'] ) &&  401 === (int) $response['response']['code'] ) {
+    $decoded_token = JWT::decode(
+      $data->id_token,
+      $this->a0_options->get_client_secret_as_key(),
```
lbalmaceda: it's confusing calling it "client secret" when on RSA you only have public/private keys. "There's no secret". I'm OK with the name if the method purpose and output is well documented. Even if this is not public to the user is good to have it documented so maintainers can follow up the code 👍

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
+      array( $this->a0_options->get_client_signing_algorithm() )
+    );
 
-      $error = new WP_Error( '401', 'auth/token response code: 401 Unauthorized' );
+    $userinfo_resp = WP_Auth0_Api_Client::get_user(
+      $this->a0_options->get( 'domain' ),
+      WP_Auth0_Api_Client::get_client_token(),
+      $decoded_token->sub
```
lbalmaceda: Same as a comment I made previously. What happens if `$decoded_token` is null/undefined because the decode function failed. Will this create a "request" anyway and execute it? if true, is not resource friendly since we know in advance it failed so the request will fail as well.

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
+    );
 
-      WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__ . ' => $this->login_user() = 401', $error );
+    $userinfo_resp_code = (int) wp_remote_retrieve_response_code( $userinfo_resp );
+    $userinfo_resp_body = wp_remote_retrieve_body( $userinfo_resp );
 
-      $msg = __( 'Error: the Client Secret configured on the Auth0 plugin is wrong. Make sure to copy the right one from the Auth0 dashboard.', 'wp-auth0' );
+    if ( $userinfo_resp instanceof WP_Error || 200 !== $userinfo_resp_code || empty( $userinfo_resp_body ) ) {
```
lbalmaceda: To be honest I think the "error handler" should catch the last 2 conditions already and create/raise a WP_ERROR. You should only be checking for the error existence, if not you are "extending" the error handler logic to every call.

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
 
-      if ( ! empty( $error ) || ! empty( $description ) ) {
-        $error = new WP_Error( $error, $description );
-        WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__ . ' => $this->login_user()', $error );
+    if ( $this->login_user( $userinfo, $data->id_token, $data->access_token ) ) {
+      $state_decoded = $this->get_state();
+      if ( ! empty( $state_decoded->interim ) ) {
+        include WPA0_PLUGIN_DIR . 'templates/login-interim.php';
+      } else {
+        if ( ! empty( $state_decoded->redirect_to ) && wp_login_url() !== $state_decoded->redirect_to ) {
```
lbalmaceda: what about

```php
     $redirectURL= $state_decoded->redirect_to
      if ( empty( $redirectURL ) || wp_login_url() !==  $redirectURL ) {
         $redirectURL = $this->a0_options->get( 'default_login_redirection' );
       }
     wp_safe_redirect( $redirectURL );
```

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
-        $description = $data->error_description;
-      }
+    $userinfo = json_decode( $userinfo_resp_body );
 
-      if ( ! empty( $error ) || ! empty( $description ) ) {
-        $error = new WP_Error( $error, $description );
-        WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__ . ' => $this->login_user()', $error );
+    if ( $this->login_user( $userinfo, $data->id_token, $data->access_token ) ) {
+      $state_decoded = $this->get_state();
+      if ( ! empty( $state_decoded->interim ) ) {
```
lbalmaceda: what's `interim`? and what is the structure of the object `state_decoded`?

lib/WP_Auth0_LoginManager.php
```diff
@@ -336,12 +331,17 @@ public function implicit_login() {
...
-  // $id_token (string): the user's JWT
-  // $access_token (string): the user's access token.  It is not provided when using the **Implicit flow**.
+  /**
+   * Does all actions required to log the user in to wordpress, invoking hooks as necessary
+   *
+   * @param object $user - the WP user object, such as returned by get_user_by()
+   * @param object $userinfo - the Auth0 profile of the user
+   * @param bool $is_new - `true` if the user was created in the WordPress database, `false` if not
+   * @param string $id_token - user's ID token returned from Auth0
+   * @param string $access_token - user's access token returned from Auth0; not provided when using implicit_login()
```
lbalmaceda: AFAIK the AcT is always returned. The one that might be missing is the IdT.

lib/WP_Auth0_LoginManager.php
```diff
@@ -356,35 +356,29 @@ private function do_login( $user, $userinfo, $is_new, $id_token, $access_token )
...
-     * @param array $credentials {
-     *     Array of entered sign-on data.
-     *
-     *     @type string $user_login    Username.
-     *     @type string $user_password Password entered.
-     *     @type bool   $remember      Whether to 'remember' the user. Increases the time
-     *                                 that the cookie will be kept. Default false.
-     * }
-     */
+    // See wp_signon() for documentation on this filter
```
lbalmaceda: Ah that's lazy 😛 I don't see it in this diff. Was this a copy of that method documentation? What's the reason behind removing this one?

WP_Auth0.php
joshcanhelp: Maybe I just fell asleep in the middle of that docblock or something ...

WP_Auth0.php
joshcanhelp: It registers query variables (URL params) with the main `WP_Query` instance that basically runs the whole site.

lib/WP_Auth0_LoginManager.php
joshcanhelp: Yes, can be `1` for a "regular" login or `implicit`

lib/WP_Auth0_LoginManager.php
joshcanhelp: We'll only ever get a scalar back from `query_vars()` and `sanitize_text_field()` is just there to sanitize. We're expecting error descriptions from the Auth0 server so if something comes through that's wonky then it's probably someone screwing around. If `sanitize_text_field()` returns an empty value then it was empty to begin with or invalid symbols:

https://developer.wordpress.org/reference/functions/sanitize_text_field/

lib/WP_Auth0_LoginManager.php
joshcanhelp: If the `code` is missing then we probably have bigger problems. This is reacting to a auth code grant so we either have a `code` or we have an error (in my understanding)

lib/WP_Auth0_LoginManager.php
joshcanhelp: `WP_Auth0_LoginFlowValidationException` extends `Exception` which accepts a message as the first argument and a code as the second:

https://secure.php.net/manual/en/exception.construct.php

So this is trying to get a message to use on the first line and a code on the second ... tertiaries as arguments. I think the extemporaneous parentheses on that second arg are confusing so I'm removing them.

lib/WP_Auth0_LoginManager.php
joshcanhelp: It's a public method defined before my time ... that process, name-wise, does not make a lot of sense, I agree. It's more like "get the decoded secret if HS256, get the required jwks value if not."

lib/WP_Auth0_LoginManager.php
joshcanhelp: Good call, a bit of a shortcut on my part. Fix OTW.

lib/WP_Auth0_LoginManager.php
joshcanhelp: Also a good call. If `$userinfo_resp` is a `WP_Error` then this will also be true (resolves the code to 0).

lib/WP_Auth0_LoginManager.php
joshcanhelp: What if `$state_decoded->redirect_to` is not set?

joshcanhelp: I guess that's not necessary to consider since I'm setting the state ...

lib/WP_Auth0_LoginManager.php
joshcanhelp: TBH, I'm not sure about `interim` at the moment, has never come up so I left it in-tact. 

State is an object with `interim` (bool), `nonce` (state validation), and `state` (always set to "nonce" ... which can probably go). Sometimes `redirect_to` is set to a URL but it can be blank.

lib/WP_Auth0_LoginManager.php
joshcanhelp: That's true ... but the `$id_token` param was already there

lib/WP_Auth0_LoginManager.php
joshcanhelp: Not really lazy, more like an unnecessary duplicate of WP core documentation. Seems silly to have that here and in the core location and try to keep the two updated.

joshcanhelp:
@lbalmaceda - Fixes up and ready for re-review!

lib/WP_Auth0_LoginManager.php
lbalmaceda: Hmm.. That's confusing, but as long as it's "internal" I don't care. 👍

lib/WP_Auth0_LoginManager.php
lbalmaceda: I find the code easier to follow if you fetch the values before creating the exception, unless you have some performance impact for doing so.

```php
message = $data->error_description ) ? $data->error_description : __( 'Unknown error', 'wp-auth0' )
code = ! empty( $data->error ) ? $data->error : $exchange_resp_code
throw new WP_Auth0_LoginFlowValidationException(message, code)
```

lib/WP_Auth0_LoginManager.php
joshcanhelp: I agree but it's legacy and dangerous to unwire at this point. I'd like to just deprecate implicit altogether ...

lib/WP_Auth0_LoginManager.php
```diff
@@ -163,136 +170,124 @@ public function init_auth0() {
...
-      }
+    $userinfo = json_decode( $userinfo_resp_body );
 
-      if ( ! empty( $error ) || ! empty( $description ) ) {
-        $error = new WP_Error( $error, $description );
-        WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__ . ' => $this->login_user()', $error );
+    if ( $this->login_user( $userinfo, $data->id_token, $data->access_token ) ) {
+      $state_decoded = $this->get_state();
+      if ( ! empty( $state_decoded->interim ) ) {
+        include WPA0_PLUGIN_DIR . 'templates/login-interim.php';
```
lbalmaceda: there was an `exit()` before right below this line. Are you sure it's no longer required?

lib/WP_Auth0_LoginManager.php
joshcanhelp: @lbalmaceda - Agree and fixed!

lib/WP_Auth0_LoginManager.php
lbalmaceda: @joshcanhelp are you saying that `state` is the string `"nonce"` or that `state==nonce` ? In any case, remember this are individual values and should be different, secure random strings.

lib/WP_Auth0_LoginManager.php
joshcanhelp: `exit()` was moved to catch this and `wp_safe_redirect()` (which requires an `exit`/`die`). [Here](https://github.com/auth0/wp-auth0/blob/46e20ecf53029b81a0ee82dd664123afa5d6bca4/lib/WP_Auth0_LoginManager.php#L278)

lib/WP_Auth0_LoginManager.php
joshcanhelp: Currently (in JSON for clarity):
```json
{
  "interim" : true || false,
  "nonce" : "2309u29j23o28j2je8",
  "state" : "nonce"
}
```

lib/WP_Auth0_LoginManager.php
lbalmaceda: I don't know, I guess from your answer that it will throw an unhandled exception. If that's the case then fine 👍. I only suggested this as a way to avoid calling `$state_decoded->redirect_to` 3 times

lib/WP_Auth0_LoginManager.php
lbalmaceda: So why isn't the access token provided "when using `implicit_login()`"?

lib/WP_Auth0_LoginManager.php
lbalmaceda: That's fair. 👍

lib/WP_Auth0_LoginManager.php
```diff
@@ -231,34 +231,44 @@ public function redirect_login() {
...
+            $decoded_token->sub
+        );
+    }
 
     $userinfo_resp_code = (int) wp_remote_retrieve_response_code( $userinfo_resp );
     $userinfo_resp_body = wp_remote_retrieve_body( $userinfo_resp );
 
-    if ( $userinfo_resp instanceof WP_Error || 200 !== $userinfo_resp_code || empty( $userinfo_resp_body ) ) {
+    // Management API call failed
+    if ( 200 !== $userinfo_resp_code || empty( $userinfo_resp_body ) ) {
```
lbalmaceda: What I meant is remove this 2 conditions and only keep the `$userinfo_resp instanceof WP_Error` one. Why did you choose to remove the other?

joshcanhelp: `$userinfo_resp` is only a `WP_Error` if something went wrong during the POST call. If it succeeds but you get a 401 or 404, the response code is what needs to be checked.

PR #411, opened 2018-03-19
[WIP] Change SSO to use HLP

joshcanhelp:
This commit is a working alpha for using the hosted login page for SSO
logins rather than checkSession on the login page.


joshcanhelp:
Removing this from the next release as it's getting too big and this does not fix anything critical at the moment.

Issue #410, opened 2018-03-13
Deprecate `oauth/ro` endpoint


joshcanhelp:
This is unlikely to affect anyone but the methods that use this endpoint are public so we'll need to properly deprecate these (`E_DEPRECATION` error and documentation in the code). Since these are not used in any of the core processes, this will only affect developers who have extended the plugin to use this endpoint. 

Affected methods:

`WP_Auth0_Api_Client::ro()` - posts to the `oauth/ro` endpoint and returns a decoded response body. More information [here](https://auth0.com/docs/api/authentication#resource-owner), replace with a [Password Grant](https://auth0.com/docs/api-auth/grant/password). 

`WP_Auth0_LoginManager::login_with_credentials()` - calls `WP_Auth0_Api_Client::ro()`. 

These will be deprecated in the next version and removed sometime this year. If you have any questions or need any assistance migrating to the new flow, please let us know in the comments below or on our [Community site](https://community.auth0.com/).

joshcanhelp:
Fixed with #413

PR #413, opened 2018-03-20
Deprecate oauth/ro endpoint

joshcanhelp:
Deprecation notices on 2 public methods that are not used anywhere in the plugin.


joshcanhelp:
Fixes #410

PR #417, opened 2018-03-26
Fix improper state handling

joshcanhelp:
State generation, specifically cookie storage, needs to happen before
any output. The previous arch was ok for login page generation but
failed with a "headers already set" error if used in the shortcode.
This addresses the issue by storing state earlier and getting the
value later.

### DIFF
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -13,6 +13,7 @@ define( 'WPA0_LANG', 'wp-auth0' ); // deprecated; do not use for translations
 define( 'AUTH0_DB_VERSION', 17 );
 define( 'WPA0_VERSION', '3.5.2' );
 define( 'WPA0_CACHE_GROUP', 'wp_auth0' );
+define( 'WPA0_STATE_COOKIE_NAME', 'auth0_state' );
 
 /**
  * Main plugin class
@@ -119,9 +120,6 @@ class WP_Auth0 {
 		$edit_profile->init();
 
 		$this->check_signup_status();
-
-		// Store nonce value before headers are sent
-		WP_Auth0_State_Handler::getInstance()->setCookie();
 
 		WP_Auth0_Email_Verification::init();
 	}
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -120,6 +120,9 @@ class WP_Auth0 {
 
 		$this->check_signup_status();
 
+		// Store nonce value before headers are sent
+		WP_Auth0_State_Handler::getInstance()->setCookie();
+
 		WP_Auth0_Email_Verification::init();
 	}
 
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -7,6 +7,7 @@ class WP_Auth0_LoginManager {
   protected $ignore_unverified_email;
   protected $users_repo;
   protected $state;
+  protected $state_decoded;
 
   public function __construct( WP_Auth0_UsersRepo $users_repo, $a0_options = null, $default_role = null, $ignore_unverified_email = false ) {
 
@@ -157,11 +158,9 @@ class WP_Auth0_LoginManager {
       $this->die_on_login( $error_msg ? $error_msg : sanitize_text_field( $this->query_vars( 'error' ) ) );
     }
 
-    // Check for valid state nonce
+    // Check for valid state nonce, set in WP_Auth0_Lock10_Options::get_state_obj()
     // See https://auth0.com/docs/protocols/oauth2/oauth-state
-    $state_decoded = $this->get_state();
-    $state_nonce = isset( $state_decoded->nonce ) ? $state_decoded->nonce : '';
-    if ( ! WP_Auth0_State_Handler::getInstance()->validate( $state_nonce ) ) {
+    if ( ! $this->validate_state() ) {
       $this->die_on_login( __( 'Invalid state', 'wp-auth0' ) );
     }
 
@@ -275,7 +274,7 @@ class WP_Auth0_LoginManager {
     $userinfo = json_decode( $userinfo_resp_body );
 
     if ( $this->login_user( $userinfo, $data->id_token, $data->access_token ) ) {
-      $state_decoded = $this->get_state();
+      $state_decoded = $this->get_state( TRUE );
       if ( ! empty( $state_decoded->interim ) ) {
         include WPA0_PLUGIN_DIR . 'templates/login-interim.php';
       } else {
@@ -302,7 +301,7 @@ class WP_Auth0_LoginManager {
   public function implicit_login() {
 
     $token = $_POST['token'];
-    $state_decoded = $this->get_state();
+    $state_decoded = $this->get_state( TRUE );
 
     $secret = $this->a0_options->get_client_secret_as_key();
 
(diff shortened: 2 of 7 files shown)


WP_Auth0.php
```diff
@@ -120,6 +120,9 @@ public function init() {
 
 		$this->check_signup_status();
 
+		// Store nonce value before headers are sent
+		new WP_Auth0_State_Handler();
```
lbalmaceda: why isn't this being assigned anywhere? If this is some kind of "static helper" maybe you'd want to have a singleton to access it instead of creating new instances.

lib/WP_Auth0_State_Handler.php
```diff
@@ -21,16 +21,20 @@ class WP_Auth0_State_Handler {
...
+    }
   }
 
   /**
-   * Set the unique ID for state and return
+   * Return the unique ID used for state validation
    *
    * @return string
    */
-  public function issue() {
```
lbalmaceda: isn't this a breaking change?

lib/WP_Auth0_State_Handler.php
```diff
@@ -72,7 +77,7 @@ public static function reset() {
    *
    * @return string
    */
-  public static function generateNonce() {
+  protected function generateNonce() {
```
lbalmaceda: isn't this a breaking change?

lib/WP_Auth0_State_Handler.php
```diff
@@ -40,6 +44,7 @@ public function issue() {
    * @return bool
    */
   protected function store() {
+    $_COOKIE[ self::COOKIE_NAME ] = $this->uniqid;
```
lbalmaceda: who was setting this value to `$_COOKIE` previously?

WP_Auth0.php
joshcanhelp: 🤔

lib/WP_Auth0_State_Handler.php
joshcanhelp: Has not gone out yet

lib/WP_Auth0_State_Handler.php
joshcanhelp: `setcookie()` (PHP core) does the work sending the header but it doesn't set the global for use later. I need to check for that cookie later on during processing

lib/WP_Auth0_LoginManager.php
```diff
@@ -543,16 +543,41 @@ protected function query_vars( $key ) {
...
-    return is_object( $this->state ) ? $this->state : null;
+  }
+
+  /**
+   * Check the state send back from Auth0 with the one stored in the user's browser
+   *
+   * @return bool
+   */
+  protected function validate_state() {
+    $valid = isset( $_COOKIE[ WPA0_STATE_COOKIE_NAME ] )
```
lbalmaceda: Are we going to force people to set the state then? Or better put.. is this library setting a new state on every request?

lib/WP_Auth0_Nonce_Handler.php
```diff
@@ -0,0 +1,131 @@
+<?php
+
+final class WP_Auth0_Nonce_Handler {
+
+  /**
+   * Cookie name used to store unique state value
+   *
+   * @var string
+   */
+  const COOKIE_NAME = 'auth0_uniqid';
```
lbalmaceda: maybe you want to use "auth0_nonce" here?

lib/WP_Auth0_Nonce_Handler.php
```diff
@@ -0,0 +1,131 @@
+<?php
+
+final class WP_Auth0_Nonce_Handler {
+
+  /**
+   * Cookie name used to store unique state value
```
lbalmaceda: "unique state value" shouldn't be "nonce"?

lib/WP_Auth0_Nonce_Handler.php
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * State nonce stored in a cookie
+   *
+   * @var string
+   */
+  private $_uniqid;
+
+  /**
+   * WP_Auth0_State_Handler constructor
```
lbalmaceda: nonce handler

lib/WP_Auth0_Nonce_Handler.php
```diff
@@ -0,0 +1,131 @@
...
+
+  /**
+   * Start-up process to make sure we have a unique ID stored
+   */
+  private function init() {
+    if ( isset( $_COOKIE[ self::COOKIE_NAME ] ) ) {
+      // Have a state cookie, don't want to generate a new one
+      $this->_uniqid = $_COOKIE[ self::COOKIE_NAME ];
+    } else {
+      // No state cookie, need to create one
```
lbalmaceda: nonce

lib/WP_Auth0_Nonce_Handler.php
```diff
@@ -0,0 +1,131 @@
...
+   */
+  public static final function getInstance() {
+    if ( null === self::$_instance ) {
+      self::$_instance = new WP_Auth0_Nonce_Handler();
+    }
+    return self::$_instance;
+  }
+
+  /**
+   * Return the unique ID used for state validation
```
lbalmaceda: nonce>

lib/WP_Auth0_Nonce_Handler.php
```diff
@@ -0,0 +1,131 @@
...
+   *
+   * @return string
+   */
+  public function get() {
+    return $this->_uniqid;
+  }
+
+  /**
+   * Check if the stored state matches a specific value
```
lbalmaceda: nonce

lib/WP_Auth0_Nonce_Handler.php
```diff
@@ -0,0 +1,131 @@
...
+   *
+   * @param $state
```
lbalmaceda: nonce

lib/WP_Auth0_Nonce_Handler.php
```diff
@@ -0,0 +1,131 @@
...
+   *
+   * @return bool
+   */
+  public function validate( $state ) {
```
lbalmaceda: nonce

lib/WP_Auth0_LoginManager.php
```diff
@@ -543,16 +543,41 @@ protected function query_vars( $key ) {
   /**
    * Get and store state returned from Auth0
    *
-   * @return object|null
+   * @param bool $decoded - `true` to return decoded state
+   *
+   * @return string|object|null
    */
-  protected function get_state() {
+  protected function get_state( $decoded = FALSE ) {
```
lbalmaceda: why would I need the "encoded" state? URL encoding a state shouldn't be responsibility of this class, but the class who constructs the requests.

lib/WP_Auth0_LoginManager.php
joshcanhelp: Setting a new state when initializing Lock, yes

lib/WP_Auth0_LoginManager.php
joshcanhelp: References base64 encoded, not URL. URL decoding has to happen, this is just basically "do you want the raw value or the object"

lbalmaceda: Why would I need it to be base64 encoded? The form of the `nonce` or `state` values should be obscure. I don't care how you generate it internally as long as it's "safe and secure". (guarantying that it's too hard to repeat itself in a short period of time, etc). 

And TBH, the server doesn't care either. It's a plain text string that it needs to put somewhere so the client can later obtain it and validate it with the one it sent before. So unless you have a reason **why** to have a method supporting both encoded/decoded scenarios, I think it's useless.

joshcanhelp: It's base64 encoded when it's created in `WP_Auth0_Lock10_Options->get_state_obj()` ... more specifically, it's a JSON encoded object that's then base64 encoded. That's what I store in a cookie then validate during the login process, still base64 encoded. Once someone is logged in, then I need that object I created before to get a redirect URL. 

If you look for where `get_state` is used in this class, you'll see what I mean

PR #418, opened 2018-03-27
Improve WP_Auth0_Options

joshcanhelp:
Adding get_lock_connections and add_lock_connection for working with
separated options field. reordering option defaults, remove comments,
add docblocks

### DIFF
--- a/lib/WP_Auth0_Options.php
+++ b/lib/WP_Auth0_Options.php
@@ -2,15 +2,15 @@
 
 class WP_Auth0_Options extends WP_Auth0_Options_Generic {
 
-	protected static $instance = null;
-	public static function Instance() {
-		if ( self::$instance === null ) {
-			self::$instance = new WP_Auth0_Options;
-		}
-		return self::$instance;
-	}
+	protected static $_instance = null;
+	protected $_options_name = 'wp_auth0_settings';
 
-	protected $options_name = 'wp_auth0_settings';
+	public static function Instance() {
+		if ( null=== self::$_instance ) {
+			self::$_instance = new WP_Auth0_Options;
+		}
+		return self::$_instance;
+	}
 
 	public function is_wp_registration_enabled() {
 		if ( is_multisite() ) {
@@ -19,6 +19,7 @@ class WP_Auth0_Options extends WP_Auth0_Options_Generic {
 		return get_site_option( 'users_can_register', 0 ) == 1;
 	}
 
+	// TODO: Deprecate, not used
 	public function get_enabled_connections() {
 		return array( 'facebook', 'twitter', 'google-oauth2' );
 	}
@@ -43,9 +44,13 @@ class WP_Auth0_Options extends WP_Auth0_Options_Generic {
 		return $defaults[$key];
 	}
 
+	/**
+	 * Get the stored token signing algorithm
+	 *
+	 * @return string
+	 */
 	public function get_client_signing_algorithm() {
-			$client_signing_algorithm = $this->get('client_signing_algorithm', WP_Auth0_Api_Client::DEFAULT_CLIENT_ALG);
-			return $client_signing_algorithm;
+		return $this->get( 'client_signing_algorithm', WP_Auth0_Api_Client::DEFAULT_CLIENT_ALG );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


lib/WP_Auth0_Options.php
```diff
@@ -129,92 +134,124 @@ public function get_logout_url() {
 		return add_query_arg( 'action', 'logout', site_url( 'wp-login.php', 'login' ) );
 	}
 
+	/**
+	 * Get as a string-separated string and parse to array
```
lbalmaceda: No. That's what the method does internally. What devs should care is "obtain the connections as an array of strings".

lib/WP_Auth0_Options.php
```diff
@@ -129,92 +134,124 @@ public function get_logout_url() {
...
+	 *
+	 * @return array
+	 */
+	public function get_lock_connections() {
+		$connections = $this->get( 'lock_connections' );
```
lbalmaceda: why is the `lock_connections` a "string,of,connections"? It seems you are always manipulating arrays. Can it be defined as an array to avoid all this coming and going from one structure to the other?

lib/WP_Auth0_Options.php
```diff
@@ -129,92 +134,124 @@ public function get_logout_url() {
...
+		$connections = empty( $connections ) ? array() : explode( ',', $connections );
+		return array_map( 'trim', $connections );
```
lbalmaceda: is the trim needed because it's a user input or because you don't trust the source? (i don't know where it comes from)

lib/WP_Auth0_Options.php
```diff
@@ -129,92 +134,124 @@ public function get_logout_url() {
...
 			'last_step' => 1,
-			'auto_login' => 0,
-			'auto_login_method' => '',
+			'migration_token_id' => null,
+			'use_lock_10' => true,
+			'jwt_auth_integration' => false,
+			'amplificator_title' => '',
+			'amplificator_subtitle' => '',
+			'connections' => array(),
+			'auth0js-cdn' => '//cdn.auth0.com/js/auth0/9.1/auth0.min.js',
```
lbalmaceda: you are using `https`, right?

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -556,8 +556,6 @@ public function connections_validation( $old_options, $input ) {
 
     if ($input['passwordless_enabled'] && $input['passwordless_enabled'] != $old_options['passwordless_enabled']) {
 
-      // $check_if_enabled = explode(',', $input['lock_connections']);
```
lbalmaceda: what happened to all these comments that are now removed? was it implemented somewhere?

lib/WP_Auth0_Options.php
```diff
@@ -43,9 +44,13 @@ public function get_default($key) {
...
 
+	/**
+	 * Get the stored token signing algorithm
+	 *
+	 * @return string
+	 */
 	public function get_client_signing_algorithm() {
-			$client_signing_algorithm = $this->get('client_signing_algorithm', WP_Auth0_Api_Client::DEFAULT_CLIENT_ALG);
-			return $client_signing_algorithm;
+		return $this->get( 'client_signing_algorithm', WP_Auth0_Api_Client::DEFAULT_CLIENT_ALG );
```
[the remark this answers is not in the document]
joshcanhelp: Yes

lib/WP_Auth0_Options.php
joshcanhelp: User input

lib/WP_Auth0_Options.php
joshcanhelp: Some folks are, some are not ... though this URL will always be `https` so I guess that could be explicit here? This is just a reordering, FYI

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: They've been commented for a while

lib/WP_Auth0_Options.php
joshcanhelp: The field in on the settings page is just a text field so comma-separated (or line break or pipe or whatever) is about the best we can do without a UI to force multiple, separate values. This is basically to handle user input

lib/WP_Auth0_Options.php
lbalmaceda: it's hosted by auth0 so I don't see a reason why they shouldn't be using `https`. They don't need to setup anything on their side. It's not like a custom domain where the certificate needs to be handled by your or something like that.. I'd enforce https on all auth0 urls. Double check with luis, but i think that's the right choice.

lib/admin/WP_Auth0_Admin_Advanced.php
lbalmaceda: but is it something you want to do someday? are you tracking it somewhere else or are they just gone for good? 😂

lib/WP_Auth0_Options.php
luisrudge: this should be always https

lib/WP_Auth0_Options.php
```diff
@@ -129,92 +134,124 @@ public function get_logout_url() {
...
+		$connections = empty( $connections ) ? array() : explode( ',', $connections );
+		return array_map( 'trim', $connections );
+	}
+
+	/**
+	 * Add a new connection to the lock_connections setting
+	 *
+	 * @param string $add_conn - connection name to add
+	 */
+	public function add_lock_connection( $add_conn ) {
```
lbalmaceda: I'd rename this parameter to "connection" or "conn" since the method name already describes what you are doing with it.

lib/WP_Auth0_Options.php
```diff
@@ -129,92 +134,124 @@ public function get_logout_url() {
...
+		$connections = $this->get_lock_connections();
+
+		// Add if it doesn't exist already
+		if ( ! array_key_exists( $add_conn, $connections ) ) {
```
lbalmaceda: This is a suggestion:
It could be more performant if you FIRST check that the string from `$this->get( 'lock_connections' );` "contains" the trimmed `$add_conn` parameter.
- If it DOES contain it, return (skip). This way, you avoid converting it to an array in the first place.
- if not, then you can call the `get_lock_connections()` method that prepares the array for you.

lib/WP_Auth0_Options.php
joshcanhelp: "string-separated string" 😆

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: I'm not sure what it's for, it's obviously not doing anything, and it's related to the settings I'm working with so 👋 

:)

lib/WP_Auth0_Options.php
joshcanhelp: You can't really do that can you? What if I want to add the connection `google-oauth` so I search for the string `google` in `twitter,facebook,google-oauth2` and it's there so I bail out?

PR #420, opened 2018-03-28
add WPML compatibility for frontend

bjoly-nexurium:
Hi, 

There a few update of the plugin to add the WPML compatibility. 
A boolean selector is added in the appereance setting. If true th language is override by the language given by WPML.

Best regards


lib/WP_Auth0_Lock10_Options.php
```diff
@@ -132,6 +132,15 @@ protected function build_settings( $settings ) {
     if (isset($settings['language']) && !empty($settings['language'])) {
       $options_obj['language'] = $settings['language'];
     }
+
```
joshcanhelp: Whitespace is off by a space it looks like

lib/WP_Auth0_Lock10_Options.php
```diff
@@ -132,6 +132,15 @@ protected function build_settings( $settings ) {
...
+   if ( isset( $settings['wpml_support'] ) && $settings['wpml_support'] == '0' ) {
+      $options_obj['wpml_support'] = null;
```
joshcanhelp: What's the reason for setting `null` explicitly?

lib/WP_Auth0_Lock10_Options.php
```diff
@@ -132,6 +132,15 @@ protected function build_settings( $settings ) {
...
+   }
+
+   if ( isset( $settings['wpml_support'] ) && $settings['wpml_support'] == '1' ) {
```
joshcanhelp: I think just a `! empty( $settings['wpml_support'] )` would work here

lib/WP_Auth0_Lock10_Options.php
```diff
@@ -132,6 +132,15 @@ protected function build_settings( $settings ) {
...
+	$options_obj['language'] =  apply_filters( 'wpml_current_language', NULL );
```
joshcanhelp: I assume that WPML uses the `wpml_current_language` filter to set the language? Or is there a default?

lib/WP_Auth0_Lock_Options.php
```diff
@@ -144,7 +144,12 @@ protected function build_settings( $settings ) {
 			$options_obj['dict'] = $settings['language'];
 		}
 
-		
+		if ( $this->_is_valid( $settings, 'wpml_support' ) ) {
```
joshcanhelp: This class is actually being deprecated in the next release. It's currently used for passwordless but that implementation is changing. You can leave this change out.

lib/admin/WP_Auth0_Admin_Appearance.php
```diff
@@ -172,5 +174,19 @@ public function basic_validation( $old_options, $input ) {
 		return $input;
 	}
 
+	public function render_wpml_support() {
```
joshcanhelp: Can we check for a core WPML function or class with `function_exists()` or `class_exists()` to determine whether this setting should render?

lib/WP_Auth0_Lock10_Options.php
joshcanhelp: My main concern here is passing a language straight from WPML to Lock. Do you have a list of the languages WPML uses? Here's the list of the [languages we support](https://github.com/auth0/lock/tree/master/src/i18n).

lib/admin/WP_Auth0_Admin_Appearance.php
bjoly-nexurium: Yes it's possible with `if ( function_exists('icl_object_id') ) { }`

lib/WP_Auth0_Lock10_Options.php
bjoly-nexurium: The filter wpml_current_language is the reference to get the current language of the user. 
The doc : https://wpml.org/wpml-hook/wpml_current_language/

What happens if the language does not exist in Lock ?
Here you can find the préconfigurated language of WPML : https://wpml.org/documentation/getting-started-guide/language-setup/pre-configured-languages/

In the Lock's code do we have an array of the available language ? if yes, may be we can use something like this : 
` if (in_array(apply_filters( 'wpml_current_language', 'en' ), $available_lang))) { 
....
}
`

lib/WP_Auth0_Lock10_Options.php
bjoly-nexurium: Nop i do the same that the option gravatar

lib/WP_Auth0_Lock10_Options.php
bjoly-nexurium: Yes that's right it's just a double check

lib/WP_Auth0_Lock10_Options.php
bjoly-nexurium: I'm not sure to understand.

lib/WP_Auth0_Lock_Options.php
joshcanhelp: @bjoly-nexurium - I think you got my feedback backwards. This block should be removed, the other block in `lib/WP_Auth0_Lock10_Options.php` should remain but be simplified.

lib/WP_Auth0_Lock10_Options.php
joshcanhelp: > What happens if the language does not exist in Lock ?

Likely falls back to English but I haven't tested that. Probably good to try out. 

> Here you can find the préconfigurated language of WPML

Do the codes match up with what Lock is expecting?

> In the Lock's code do we have an array of the available language ?

Possibly but Lock is in JS, not an easy/possible task to pull that out and use in PHP. Might need to maintain an array to make that happen (though it might be a good idea anyways to support WP core language).

bjoly-nexurium:
Everything is done ?

joshcanhelp:
@bjoly-nexurium  - Our conversation is not in the Files Changed tab, looks like it's just on your first commit, but I'm still not clear the language codes that WPML is providing and what happens if that code is not allowed by Lock.

joshcanhelp:
I'm going to close this one for now. I have a backlog task to better integrate the plugin with language settings and would rather align to the core WP language setting (or a custom one that defaults to that) than rely on another plugin's functionality. 

I appreciate the contribution here and getting us started down this road @bjoly-nexurium!

PR #422, opened 2018-03-28
Remove api_audience settings field

joshcanhelp:
This field was for an audience value used with the Management API. This
value should never change and should not be user-configurable.

The render function is `public` but has only been around for a short 
period of time so I don't think it's necessary to deprecate officially.


cocojoe:
@lbalmaceda any issue with this?

Issue #401, opened 2018-02-19
Fallback /api/v2/users/{id} to /userinfo


joshcanhelp:
Make sure that login flows can get user data when Management API is not accessible (due to misconfiguration)

joshcanhelp:
Needs a source

joshcanhelp:
Added in #423 and merged into `dev`

PR #423, opened 2018-03-29
Add /userinfo fallback during login

joshcanhelp:
Some customers are having trouble with the upgrade process in 3.5.2 while others made make changes to their account disabling management API access. This fallback allows user data to be pulled for logging-in users. Also adds requested scopes for auth code login to make this possible and updates the error manager class to handle a common error case.

### DIFF
--- a/lib/WP_Auth0_ErrorManager.php
+++ b/lib/WP_Auth0_ErrorManager.php
@@ -12,6 +12,8 @@ class WP_Auth0_ErrorManager {
 	 * @param string|WP_Error|Exception $error - error message string or discoverable error type
 	 */
 	public static function insert_auth0_error( $section, $error ) {
+		$code = 'unknown_code';
+		$message = __( 'Unknown error message', 'wp-auth0' );
 
 		if ( $error instanceof WP_Error ) {
 			$code = $error->get_error_code();
@@ -20,10 +22,13 @@ class WP_Auth0_ErrorManager {
 			$code = $error->getCode();
 			$message = $error->getMessage();
 		} elseif ( is_array( $error ) && ! empty( $error['response'] ) ) {
-			$code = ! empty( $error['response']['code'] ) ? $error['response']['code'] : 'N/A';
-			$message = ! empty( $error['response']['message'] ) ? $error['response']['message'] : 'N/A';
+			if ( ! empty( $error['response']['code'] ) ) {
+				$code = sanitize_text_field( $error['response']['code'] );
+			}
+			if ( ! empty( $error['response']['message'] ) ) {
+				$message = sanitize_text_field( $error['response']['message'] );
+			}
 		} else {
-			$code = 'N/A';
 			$message = is_object( $error ) || is_array( $error ) ? serialize( $error ) : $error;
 		}
 
--- a/lib/WP_Auth0_ErrorManager.php
+++ b/lib/WP_Auth0_ErrorManager.php
@@ -19,9 +19,12 @@ class WP_Auth0_ErrorManager {
 		} elseif ( $error instanceof Exception ) {
 			$code = $error->getCode();
 			$message = $error->getMessage();
+		} elseif ( is_array( $error ) && ! empty( $error['response'] ) ) {
+			$code = ! empty( $error['response']['code'] ) ? $error['response']['code'] : 'N/A';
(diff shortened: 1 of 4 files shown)


joshcanhelp:
Addresses #401

lib/WP_Auth0_Lock10_Options.php
```diff
@@ -4,8 +4,8 @@ class WP_Auth0_Lock10_Options {
 
   protected $wp_options;
   protected $extended_settings;
-
   protected $signup_mode = false;
+  protected $_id_token_scopes = 'openid email email_verified name nickname picture';
```
cocojoe: why id_token_scopes btw and not just scopes?

lib/WP_Auth0_LoginManager.php
```diff
@@ -245,30 +244,29 @@ public function redirect_login() {
...
+      $userinfo_resp = WP_Auth0_Api_Client::get_user( $domain, $client_credentials_token, $decoded_token->sub );
+      $userinfo_resp_code = (int) wp_remote_retrieve_response_code( $userinfo_resp );
+      $userinfo_resp_body = wp_remote_retrieve_body( $userinfo_resp );
     }
 
-    $userinfo_resp_code = (int) wp_remote_retrieve_response_code( $userinfo_resp );
-    $userinfo_resp_body = wp_remote_retrieve_body( $userinfo_resp );
-
-    // Management API call failed
+    // Management API call failed, fallback to userinfo
```
cocojoe: So why does Management API fail?

joshcanhelp: This is something we should have had in 3.5.0. It will fail if the Client/Application is not authorized on the Management API or if Client Grant has not been turned on. Has been a common support theme and wanted to get it in there for anyone else upgrading.

lib/WP_Auth0_ErrorManager.php
```diff
@@ -19,9 +19,12 @@ public static function insert_auth0_error( $section, $error ) {
 		} elseif ( $error instanceof Exception ) {
 			$code = $error->getCode();
 			$message = $error->getMessage();
+		} elseif ( is_array( $error ) && ! empty( $error['response'] ) ) {
+			$code = ! empty( $error['response']['code'] ) ? $error['response']['code'] : 'N/A';
```
lbalmaceda: Instead of `N/A` use something like `a0_unknown_code` or if you received a "plain text response" then use that  as description and use some code like `a0_plaintext_error`

lib/WP_Auth0_ErrorManager.php
```diff
@@ -19,9 +19,12 @@ public static function insert_auth0_error( $section, $error ) {
...
+			$message = ! empty( $error['response']['message'] ) ? $error['response']['message'] : 'N/A';
```
lbalmaceda: this one is fine, it makes sense.

lib/WP_Auth0_Lock10_Options.php
```diff
@@ -4,8 +4,8 @@ class WP_Auth0_Lock10_Options {
 
   protected $wp_options;
   protected $extended_settings;
-
   protected $signup_mode = false;
+  protected $_scopes = 'openid email email_verified name nickname picture';
```
lbalmaceda: `email_verified` included in `email`

lib/WP_Auth0_Lock_Options.php
```diff
@@ -4,8 +4,8 @@ class WP_Auth0_Lock_Options {
 
 	protected $wp_options;
 	protected $extended_settings;
-
 	protected $signup_mode = false;
+	protected $_id_token_scopes = 'openid email email_verified name nickname picture';
```
lbalmaceda: Same scope comment

lib/WP_Auth0_LoginManager.php
```diff
@@ -245,30 +244,29 @@ public function redirect_login() {
 
     // Attempt to authenticate with the Management API
     $client_credentials_token = WP_Auth0_Api_Client::get_client_token();
-    $userinfo_resp = null;
+    $userinfo_resp_code = $userinfo_resp_body = null;
```
lbalmaceda: what's this 2 `=` assignment?

joshcanhelp: Equivalent to:

```
$userinfo_resp_code = null;
$userinfo_resp_body = null;
```

PR #426, opened 2018-04-03
Fix Implicit login handling

joshcanhelp:
- Fixed error handling for implicit login
- Moved implicit handling JS to own file; refactored
- Added nonce handling for auto-login
- Added state handling for auto-login
- Better nonce generator if < PHP 7 (`openssl_random_pseudo_bytes()`)

### DIFF
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -122,6 +122,10 @@ class WP_Auth0 {
 
 		$this->check_signup_status();
 
+		if ( $this->a0_options->get( 'auto_login' ) ) {
+			WP_Auth0_Nonce_Handler::getInstance()->setCookie();
+		}
+
 		WP_Auth0_Email_Verification::init();
 	}
 
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -345,26 +345,49 @@ class WP_Auth0 {
 
 	}
 
+	/**
+	 * Enqueue styles and scripts on the wp-login.php page if the plugin has been configured
+	 */
 	public function render_auth0_login_css() {
-		$client_id = WP_Auth0_Options::Instance()->get( 'client_id' );
-
-		if ( trim( $client_id ) === '' ) {
+		if ( ! WP_Auth0::ready() ) {
 			return;
 		}
-?>
-		<link rel='stylesheet' href='<?php echo plugins_url( 'assets/css/login.css', __FILE__ ); ?>' type='text/css' />
-	<?php
+
+		wp_enqueue_style( 'auth0', WPA0_PLUGIN_CSS_URL . 'login.css', FALSE, WPA0_VERSION );
+
+		if ( $this->a0_options->get( 'auth0_implicit_workflow' ) ) {
+			wp_enqueue_script( 'auth0-implicit', WPA0_PLUGIN_JS_URL . 'implicit-login.js', FALSE, WPA0_VERSION );
+			wp_localize_script( 'auth0-implicit', 'wpAuth0ImplicitGlobal', array(
+				'postUrl' => add_query_arg( 'auth0', 'implicit', site_url( 'index.php' ) ),
+			) );
+		}
 	}
 
+	/**
+	 * Output the Auth0 form on wp-login.php
+	 *
+	 * @hook filter:login_message
+	 *
+	 * @param $html
+	 *
+	 * @return string
+	 */
 	public function render_form( $html ) {
-		if ( ( isset( $_GET['action'] ) && $_GET['action'] == 'lostpassword' ) || ! self::ready() ) {
+		// Do not show Auth0 form when ...
+		if (
+			// .. processing lost password
+			( isset( $_GET['action'] ) && $_GET['action'] == 'lostpassword' )
+			// ... handling an Auth0 callback
+			|| ! empty( $_GET[ 'auth0' ] )
+			// ... plugin is not configured
+			|| ! self::ready()
+		) {
 			return $html;
 		}
 
 		ob_start();
 		require_once WPA0_PLUGIN_DIR . 'templates/login-form.php';
 		renderAuth0Form();
-
 		return ob_get_clean();
 	}
 
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -359,7 +359,7 @@ class WP_Auth0 {
 
 		wp_enqueue_style( 'auth0', WPA0_PLUGIN_CSS_URL . 'login.css', FALSE, WPA0_VERSION );
 
-		if ( $this->a0_options->get( 'auth0_implicit_workflow' ) ) {
+		if ( $this->a0_options->get( 'auth0_implicit_workflow' ) && ! empty( $_GET[ 'auth0' ] ) ) {
 			wp_enqueue_script( 'auth0-implicit', WPA0_PLUGIN_JS_URL . 'implicit-login.js', FALSE, WPA0_VERSION );
 			wp_localize_script( 'auth0-implicit', 'wpAuth0ImplicitGlobal', array(
 				'postUrl' => add_query_arg( 'auth0', 'implicit', site_url( 'index.php' ) ),
--- a/assets/js/implicit-login.js
+++ b/assets/js/implicit-login.js
@@ -1,0 +1,35 @@
+document.addEventListener( 'DOMContentLoaded', function() {
+    if ( ! window.location.hash || window.location.hash.indexOf('id_token') < 0 ) {
+        return;
+    }
+
+    var hash = window.location.hash;
+    if (hash[0] === '#') {
+        hash = hash.slice(1);
+    }
+    var data = hash.split('&').reduce(function(p,c) {
+        var parts = c.split('=');
+        p[parts[0]] = parts[1];
+        return p;
+    }, {});
+
+    var form = document.createElement('form');
+    form.setAttribute('method', 'post');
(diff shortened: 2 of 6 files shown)


joshcanhelp:
@cocojoe - as usual 😄 working on a Docker setup now to make life a little easier.

WP_Auth0.php
```diff
@@ -352,19 +358,26 @@ public static function render_back_to_auth0() {
...
+			wp_enqueue_script( 'auth0-implicit', WPA0_PLUGIN_JS_URL . 'implicit-login.js', FALSE, WPA0_VERSION );
+			wp_localize_script( 'auth0-implicit', 'wpAuth0ImplicitGlobal', array(
+				'postUrl' => add_query_arg( 'auth0', 'implicit', site_url( 'index.php' ) ),
+			) );
+		}
 	}
 
 	public function render_form( $html ) {
-		if ( isset( $_GET['action'] ) && $_GET['action'] == 'lostpassword' ) {
+		if ( ( isset( $_GET['action'] ) && $_GET['action'] == 'lostpassword' ) || ! empty( $_GET[ 'auth0' ] ) ) {
```
lbalmaceda: what does this change represent?

assets/js/implicit-login.js
```diff
@@ -0,0 +1,35 @@
+document.addEventListener( 'DOMContentLoaded', function() {
+    if ( ! window.location.hash || window.location.hash.indexOf('id_token') < 0 ) {
+        return;
```
lbalmaceda: If the token is missing you return. Does that mean that you are ignoring a possible auth error?

assets/js/implicit-login.js
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    var hash = window.location.hash;
```
lbalmaceda: I don't know how much resources you are saving, but this line could be the first one and `hash` used on the `if`

assets/js/implicit-login.js
```diff
@@ -0,0 +1,35 @@
...
+    if (hash[0] === '#') {
```
lbalmaceda: will this work fine if `hash=""`?

lib/WP_Auth0_LoginManager.php
```diff
@@ -297,47 +294,60 @@ public function redirect_login() {
...
+
+      // Validated in $this->init_auth0()
+      $state_decoded = $this->get_state( TRUE );
 
-          wp_safe_redirect( $redirectURL );
-          exit;
+      if ( ! empty( $state_decoded->interim ) ) {
+        include WPA0_PLUGIN_DIR . 'templates/login-interim.php';
+      } else {
+        $redirect_to = ! empty( $state_decoded->redirect_to ) ? $state_decoded->redirect_to : null;
```
lbalmaceda: Isn't this the same, or you need an explicit `null` later?

```php
$redirect_to = $state_decoded->redirect_to
```

lib/WP_Auth0_LoginManager.php
```diff
@@ -551,7 +561,7 @@ protected function get_state( $decoded = FALSE ) {
 
     if ( empty( $this->state ) ) {
       // Get and store base64 encoded state
-      $state_val = $this->query_vars( 'state' );
+      $state_val = $_REQUEST[ 'state' ];
```
lbalmaceda: Why, did the previous stop working?

lib/WP_Auth0_Nonce_Handler.php
```diff
@@ -123,9 +123,12 @@ public function reset() {
...
    * @see https://secure.php.net/manual/en/function.random-bytes.php
    *
+   * @param int $bytes - number of bytes to generate
+   *
    * @return string
    */
-  public function generateNonce() {
-    return function_exists( 'random_bytes' ) ? bin2hex( random_bytes( 32 ) ) : md5( uniqid( rand(), true ) );
+  public function generateNonce( $bytes = 32 ) {
+    $nonce_bytes = function_exists( 'random_bytes' ) ? random_bytes( $bytes ) : openssl_random_pseudo_bytes( $bytes );
```
lbalmaceda: will `openssl_random_pseudo_bytes` always exist?

joshcanhelp: This plugin is tagged PHP 5.3, which is when this function was introduced:

https://secure.php.net/manual/en/function.openssl-random-pseudo-bytes.php

lib/WP_Auth0_LoginManager.php
joshcanhelp: `$this->query_vars()` is not explicit enough, makes for confusion around how these vars are handled. I want `state` from either POST or GET (aka `$_REQUEST`), I don't want anything else.

lib/WP_Auth0_LoginManager.php
joshcanhelp: `$state_decoded->redirect_to` will throw an error if it does not exist, so I'm checking whether it's there or not with `empty()`

assets/js/implicit-login.js
joshcanhelp: screenshot 2018-04-04 11 57 59

joshcanhelp: But it can't be empty because we've escaped already if that's the case

assets/js/implicit-login.js
joshcanhelp: Good question and I don't know if I have a good answer. The core login page processing does not use anchors at all and I don't want this to run unless I have something to do here. I'm not sure of another situation where we would have a hash and want to do something with it. I don't see anything else in the [implicit grant doc](https://auth0.com/docs/api-auth/tutorials/implicit-grant). 

Is there an error message that I might get back? `error` or `error_description`?

WP_Auth0.php
joshcanhelp: If the `auth0` URL param is set, don't load the form. It's only sent when coming back from an implicit grant. Auth code grant goes straight to `index.php?auth0=1`

joshcanhelp: If I load the form, I get a new `state` set so I don't want to do that ... also, no need to load the form at all since we're just passing through.

lbalmaceda: Alright, I remember the `auth0` param now. Are you planning to refactor this "auth0 hack" anytime soon?

WP_Auth0.php
```diff
@@ -123,6 +123,10 @@ public function init() {
 
 		$this->check_signup_status();
 
+		if ( $this->a0_options->get( 'auto_login' ) ) {
+			WP_Auth0_Nonce_Handler::getInstance()->setCookie();
```
lbalmaceda: Is this well indented?

joshcanhelp: Like, correct whitespace? Yes. This project mixes tabs and spaces, as well as how many spaces for each. I'll do an automated whitespace cleanup before I start the next release.

WP_Auth0.php
joshcanhelp: The param can show up on the login page and on the `index.php` entry point.

joshcanhelp: Not loading the form in this case does solve the problem of re-setting a state cookie but we don't need to load the login form at all if this is a response from Auth0. So, I would say that the URL parameter check will stay but state checking will evolve to handle this better (by abstracting that `WP_Auth0_Nonce_Handler` class to handle both scenarios)

assets/js/implicit-login.js
```diff
@@ -14,23 +17,44 @@ document.addEventListener('DOMContentLoaded', function() {
...
-    var hiddenField = document.createElement('input');
-    hiddenField.setAttribute('type', 'hidden');
+    if ( data.hasOwnProperty( 'error' ) ) {
+        form.appendChild( createHiddenField( 'error', data.error ) );
+        form.appendChild( createHiddenField( 'error_description', data.error_description ) );
+    } else if ( data.hasOwnProperty( 'id_token' ) && data.hasOwnProperty( 'state' ) ) {
+        form.appendChild( createHiddenField( 'token', data.id_token ) );
+        form.appendChild( createHiddenField( 'state', data.state ) );
+    } else {
+        redirectWithoutQuery();
```
lbalmaceda: Is this a valid path? When will it happen? Do you have an example? If you don't have one but still want to catch it, would be nice to log that query somewhere to debug and understand it I guess. If the authentication was initiated for "implicit flow" and the token is not there, there must be an error.

lib/WP_Auth0_LoginManager.php
```diff
@@ -150,9 +150,10 @@ public function init_auth0() {
 
     // Catch any incoming errors and stop the login process
     // See https://auth0.com/docs/libraries/error-messages
-    if ( $this->query_vars( 'error_description' ) ) {
+    if ( $this->query_vars( 'error' ) || $this->query_vars( 'error_description' ) ) {
```
lbalmaceda: I think the `error_description` might be missing sometimes (AFAIK is optional). In any case, because `error` is always present that condition should suffice.

assets/js/implicit-login.js
joshcanhelp: It's valid as in it exists but without a hash we don't have much to do. The param triggers this JS to load and for the form to be excluded (param tells us that we need to process something from Auth0). If we don't have the information we need, then we need to cease processing, hence the redirect. 

If we have the Auth0 param but no `error` or no `state`, that could be an intermittent problem, in which case showing the login form to try again seems reasonable. I'm not sure I want to catch and log every possible outcome here, I think it's best to get someone back to a place where they can try again.

PR #429, opened 2018-04-10
Change how Advanced admin settings fields are output

joshcanhelp:
This commit will:

- Switch Advanced admin settings to use new render functions
- Reword confusing and incorrect descriptions
- Rename incorrect protected vars
- Add check for function_exists to use hooks for settings callbacks
- Fix settings order and option names

### DIFF
--- a/lib/admin/WP_Auth0_Admin_Advanced.php
+++ b/lib/admin/WP_Auth0_Admin_Advanced.php
@@ -44,12 +44,12 @@ class WP_Auth0_Admin_Advanced extends WP_Auth0_Admin_Generic {
         'id' => 'wpa0_default_login_redirection', 'function' => 'render_default_login_redirection' ),
       array( 'name' => __( 'Passwordless Login', 'wp-auth0' ), 'opt' => 'passwordless_enabled',
         'id' => 'wpa0_passwordless_enabled', 'function' => 'render_passwordless_enabled' ),
-      array( 'name' => __( 'Force HTTPS Callback', 'wp-auth0' ), 'opt' => 'force_https_callback',
-        'id' => 'wpa0_force_https_callback', 'function' => 'render_force_https_callback' ),
-      array( 'name' => __( 'Lock JS CDN URL', 'wp-auth0' ), 'opt' => 'passwordless_enabled',
-        'id' => 'wpa0_cdn_url', 'function' => 'render_cdn_url' ),
       array( 'name' => __( 'Connections to Show', 'wp-auth0' ), 'opt' => 'lock_connections',
         'id' => 'wpa0_connections', 'function' => 'render_connections' ),
+      array( 'name' => __( 'Force HTTPS Callback', 'wp-auth0' ), 'opt' => 'force_https_callback',
+        'id' => 'wpa0_force_https_callback', 'function' => 'render_force_https_callback' ),
+      array( 'name' => __( 'Lock JS CDN URL', 'wp-auth0' ), 'opt' => 'cdn_url',
+        'id' => 'wpa0_cdn_url', 'function' => 'render_cdn_url' ),
       array( 'name' => __( 'Link Users with Same Email', 'wp-auth0' ), 'opt' => 'link_auth0_users',
         'id' => 'wpa0_link_auth0_users', 'function' => 'render_link_auth0_users' ),
       array( 'name' => __( 'Auto Provisioning', 'wp-auth0' ), 'opt' => 'auto_provisioning',
@@ -58,7 +58,7 @@ class WP_Auth0_Admin_Advanced extends WP_Auth0_Admin_Generic {
         'id' => 'wpa0_migration_ws', 'function' => 'render_migration_ws' ),
       array( 'name' => __( 'Migration IPs Whitelist', 'wp-auth0' ), 'opt' => 'migration_ips_filter',
         'id' => 'wpa0_migration_ws_ips_filter', 'function' => 'render_migration_ws_ips_filter' ),
-      array( 'name' => __( 'IP Addresses', 'wp-auth0' ), 'opt' => 'migration_ips_filter',
+      array( 'name' => __( 'IP Addresses', 'wp-auth0' ), 'opt' => 'migration_ips',
         'id' => 'wpa0_migration_ws_ips', 'function' => 'render_migration_ws_ips' ),
       array( 'name' => __( 'Auto Login', 'wp-auth0' ), 'opt' => 'auto_login',
         'id' => 'wpa0_auto_login', 'function' => 'render_auto_login' ),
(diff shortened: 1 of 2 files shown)


lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -44,12 +44,12 @@ public function init() {
         'id' => 'wpa0_default_login_redirection', 'function' => 'render_default_login_redirection' ),
       array( 'name' => __( 'Passwordless Login', 'wp-auth0' ), 'opt' => 'passwordless_enabled',
         'id' => 'wpa0_passwordless_enabled', 'function' => 'render_passwordless_enabled' ),
+      array( 'name' => __( 'Connections to Show', 'wp-auth0' ), 'opt' => 'lock_connections',
+        'id' => 'wpa0_connections', 'function' => 'render_connections' ),
       array( 'name' => __( 'Force HTTPS Callback', 'wp-auth0' ), 'opt' => 'force_https_callback',
         'id' => 'wpa0_force_https_callback', 'function' => 'render_force_https_callback' ),
-      array( 'name' => __( 'Lock JS CDN URL', 'wp-auth0' ), 'opt' => 'passwordless_enabled',
+      array( 'name' => __( 'Lock JS CDN URL', 'wp-auth0' ), 'opt' => 'cdn_url',
```
lbalmaceda: can these 2 var renaming changes be considered breaking changes?

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
-
-    echo $this->render_a0_switch( "wpa0_jwt_auth_integration", "jwt_auth_integration", 1, 1 == $v );
-?>
-    <div class="subelement">
-      <span class="description"><?php echo __( 'This will enable the JWT Auth\'s Users Repository override.', 'wp-auth0' ); ?></span>
-    </div>
-  <?php
+  public function render_jwt_auth_integration( $args ) {
+    $this->render_switch( $args[ 'label_for' ], $args[ 'opt_name' ] );
+    $this->render_field_description( __( 'This will enable the JWT Auth Users Repository override', 'wp-auth0' ) );
```
lbalmaceda: Previous string had a `'s` after `JWT`. Is this intentional? and btw, what does this string mean?

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
-    ... // more fields could be specified
-  }
-]</pre></code>
-      </span>
-    </div>
-    <?php
+  public function render_custom_signup_fields( $args ) {
+    $this->render_textarea_field( $args[ 'label_for' ], $args[ 'opt_name' ] );
+    $this->render_field_description(
+      __( 'Valid JSON for custom signup fields in the Auth0 signup form. ', 'wp-auth0' ) .
```
lbalmaceda: I'd use the same name as in the docs 

~custom~ additional signup fields

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
-    echo $this->render_a0_switch( "wpa0_link_auth0_users", "link_auth0_users", 1, ! empty( $v ) );
-?>
-      <div class="subelement">
-        <span class="description"><?php echo __( 'Links accounts with the same e-mail address. It will only occur if both e-mails are previously verified.', 'wp-auth0' ); ?></span>
-      </div>
-    <?php
+  public function render_link_auth0_users( $args ) {
+    $this->render_switch( $args[ 'label_for' ], $args[ 'opt_name' ] );
+    $this->render_field_description(
+      __( 'Links accounts with the same e-mail address (emails must be verified)', 'wp-auth0' )
```
lbalmaceda: If possible, add a small sentence telling when an email is considered verified. Such as: "An email is verified when the user was created using a social provider that shared the account's email or when a user receives a verification email and proceeds with the verify flow."

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
-?>
-
-      <div class="subelement">
-        <span class="description"><?php echo __( 'The plugin will automatically add new users if they do not exist in the WordPress database if the signups are enabled (enabling this setting will enable this behaviour when signups are disabled).', 'wp-auth0' ); ?></span>
-        </div>
-    <?php
+  public function render_auto_provisioning( $args ) {
+    $this->render_switch( $args[ 'label_for' ], $args[ 'opt_name' ] );
+    $this->render_field_description(
+      __( 'Create new users in the WordPress database when signups are off. ', 'wp-auth0' ) .
```
lbalmaceda: if here you refer to the "auth0 database signups" I'd clarify that:


Create new users in the WordPress database when signups are disabled in the auth0 database connection and an existing auth0 user logs in. Or something like that. It's not clearer my suggestion, sorry haha.

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
-?>
-
-      <div class="subelement">
-        <span class="description"><?php echo __( 'This option will replace the login widget with Lock Passwordless (Username and password login will not be enabled).', 'wp-auth0' ); ?></span>
-      </div>
-    <?php
+  public function render_passwordless_enabled( $args ) {
+    $this->render_switch( $args[ 'label_for' ], $args[ 'opt_name' ] );
+    $this->render_field_description(
+      __( 'Turns on passwordless login (email or SMS) in the Auth0 form. ', 'wp-auth0' ) .
```
lbalmaceda: Passwordless (capital P, it's a product name)

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
+      __( 'Passwordless connections are managed in the ', 'wp-auth0' ) .
+      $this->get_dashboard_link( 'connections/passwordless' ) .
+      __( ' and at least one connection must be active for this to work. ', 'wp-auth0' ) .
```
lbalmaceda: active -> enabled (on this client) for this to work

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
+      __( 'Username and password login are not enabled when this is on', 'wp-auth0' )
```
lbalmaceda: "Username &/and Password" login IS not enabled

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
+    if ( $value ) {
 
-      <div class="subelement">
-        <span class="description"><?php echo __( 'Users session by default lives for two days. Enabling this setting will make the sessions be kept for 14 days.', 'wp-auth0' ); ?></span>
-      </div>
-    <?php
-  }
+      $this->render_field_description(
+        __( 'Users migration is enabled. ', 'wp-auth0' ) .
+        __( 'If you disable this setting, it must be re-enabled manually in the ', 'wp-auth0' ) .
```
lbalmaceda: Is there a reason to have this option at all? Can't both enabling and disabling actions be performed directly on the dashboard, to prevent user error? Also think in that case if removing this would be considered breaking.

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
-		  <span class="description">
-	        <?php echo __( 'Only requests from this IPs will be allowed to the migration WS.', 'wp-auth0' ); ?>
-	        <?php echo __( 'Separate multiple IPs with commas.', 'wp-auth0' ); ?>
-        </span>
-	  </div>
-	  <?php
+  public function render_migration_ws_ips( $args ) {
+    $this->render_textarea_field( $args[ 'label_for' ], $args[ 'opt_name' ] );
+    $this->render_field_description(
+      __( 'Only requests from this IPs will be allowed to the migration WS. ', 'wp-auth0' ) .
```
lbalmaceda: "will be allowed to the migration WS." I don't understand this one. Is WS webserver? Don't feel bad for writing the large version.

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
-    <div class="subelement">
-      <span class="description"><?php echo __( 'Activate this option to change the login workflow and allow the plugin to work when the server doesn\'t have internet access.', 'wp-auth0' ); ?></span>
-    </div>
-    <?php
+  public function render_auth0_implicit_workflow( $args ) {
+    $this->render_switch( $args[ 'label_for' ], $args[ 'opt_name' ] );
+    $this->render_field_description(
+      __( 'Turns on implicit login flow, which most sites will not need. ', 'wp-auth0' ) .
+      __( 'Only enable this if outbound connections to auth0.com are disabled on your server. ', 'wp-auth0' ) .
+      __( 'Your Client should be set to "Single Page App" in your ', 'wp-auth0' ) .
```
lbalmaceda: should or must?

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
+      $this->get_dashboard_link( 'clients' ) .
+      __( ' for this setting to work properly. ', 'wp-auth0' ) .
+      __( 'This will limit profile changes and other functionality in the plugin', 'wp-auth0' )
```
lbalmaceda: Is there any link you can provide with more details to this "other functionality" that gets limited?

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
-    <div class="subelement">
-      <span class="description"><?php echo __( 'Only one range per line! Range format should be as follows (spaces will be trimmed):', 'wp-auth0' ); ?></span>
-      <code>xx.xx.xx.xx - yy.yy.yy.yy</code>
-    </div>
-    <?php
+  public function render_ip_ranges( $args ) {
+    $this->render_textarea_field( $args[ 'label_for' ], $args[ 'opt_name' ] );
+    $this->render_field_description(
+      __( 'Only one range per line! Range format should be as follows (spaces ignored): ', 'wp-auth0' ) .
+      __( '<br><code>xx.xx.xx.xx - yy.yy.yy.yy</code>', 'wp-auth0' )
```
lbalmaceda: `<br>` doesn't require a closure tag?

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
+  }
 
-      <div class="subelement">
-        <span class="description"><?php echo __( 'This is used to select which connections should lock show. It is ignored when empty and is mandatory for passwordless with social mode.', 'wp-auth0' ); ?></span>
-      </div>
-    <?php
+  public function render_connections( $args ) {
+    $this->render_text_field( $args[ 'label_for' ], $args[ 'opt_name' ] );
+    $this->render_field_description(
+      __( 'Connections the Auth0 login form should show (separate multiple with commas). ', 'wp-auth0' ) .
```
lbalmaceda: Connections _that_ the .. 

and then

separate multiple connections with commas

(Same as in the IP text)

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -134,294 +134,233 @@ public function render_passwordless_method() {
...
+      __( 'If this is empty, all active connections will be shown. ', 'wp-auth0' ) .
+      __( 'Connections listed here must already be active under Connections in your ', 'wp-auth0' ) .
```
lbalmaceda: dashboard/connections/social would be enabling those for the TENANT scope, but then each CLIENT must enable those too. Make sure this is clearer please.

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -438,16 +377,14 @@ public function render_metrics() {
...
-?>
-      <div class="subelement">
-        <span class="description"><?php echo __( 'Mark this if you require the user to have a verified email to login', 'wp-auth0' ); ?></span>
-      </div>
-    <?php
+  public function render_verified_email( $args ) {
+    $this->render_switch( $args[ 'label_for' ], $args[ 'opt_name' ] );
+    $this->render_field_description(
+      __( 'Require new users to verify their email before logging in. ', 'wp-auth0' ) .
+      __( 'This will disallow logins from social connections that do not provide email (like Twitter)', 'wp-auth0' )
```
lbalmaceda: Are users even created when this case happens? i.e. I sign up using Twitter. I won't be able to log in ever??

lib/admin/WP_Auth0_Admin_Generic.php
```diff
@@ -133,7 +137,7 @@ protected function render_switch( $id, $input_name, $expand_id = '' ) {
 		printf(
 			'<div class="a0-switch"><input type="checkbox" name="%s[%s]" id="%s" data-expand="%s" value="1"%s>
 			<label for="%s"></label></div>',
-			esc_attr( $this->option_name ),
+			esc_attr( $this->_option_name ),
```
lbalmaceda: Line 30 declares this
```
$options_name = $this->_option_name . '_' . strtolower( $id );
```

Why is this getting renamed to `_option_name` ? I'm not following that

lib/admin/WP_Auth0_Admin_Generic.php
```diff
@@ -239,13 +243,15 @@ protected function render_field_description( $text ) {
 	 * Output translated dashboard HTML link
 	 *
 	 * @param string $path - dashboard sub-section, if any
+	 * @param bool $append_period - append a period and space to the output
 	 *
 	 * @return string
 	 */
-	protected function get_dashboard_link( $path = '' ) {
-		return sprintf( '<a href="https://manage.auth0.com/#/%s" target="_blank">%s</a>',
+	protected function get_dashboard_link( $path = '', $append_period = FALSE ) {
```
lbalmaceda: I don't think this is necessary or part of this method logic. The method who is calling it must be completing the format. BTW it's only used 2 times on this doc, both of which have a follow up text that can easily contain those ". " you are adding in here. On the contrary, [this line](https://github.com/auth0/wp-auth0/pull/429/files#diff-3a97bffe53fd0d3ebe9560b32470d1d2R217) would have made sense that it called this method passing true since it doesn't have a follow up text, but it's not. 😂

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: They were incorrect in a previous (unreleased) branch. More testing found the incorrect option name.

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: Yes, intentional. Sounded awkward. 

JWT Auth is an unsupported plugin we put out a while ago. So this is saying if that plugin is installed and this option is on, use the "user repository" from the plugin.

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: Not clear or correct :)

This is ... WordPress registrations are off, user is added to Auth0 and logs in, user is added to WordPress. I think what I have is about as straightforward as I can get (after a few revisions) but open to suggestions.

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: Long conversation but yes, you are correct about how it _should_ be. This is something that needs to be dealt with carefully: 

https://wordpress.org/support/topic/dont-need-user-migration/

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: Should ... and we should remove this at some point as well.

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: Not really. This is a strange setting to begin with but it's in use currently so can't be removed. Basically, instead of pulling the user data from the management API, it's only coming from the `id_token`.

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: [Not since HTML5](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br)

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: Users are created if an email is present in their userinfo. If this is on and you sign in with Twitter, you're rejected, correct.

lib/admin/WP_Auth0_Admin_Generic.php
joshcanhelp: Renamed in a previous PR but these functions were added after that.

lib/admin/WP_Auth0_Admin_Generic.php
joshcanhelp: The description method auto-adds periods. This is mainly to not screw up translation too much. A translate-able text string that just starts with a period and a space is hard to re-use in any context.

joshcanhelp:
@lbalmaceda - I saw your main review comment re: `$args`. Those functions are callbacks that always get that `$args` argument, comes from here:

https://github.com/auth0/wp-auth0/blob/master/lib/admin/WP_Auth0_Admin_Generic.php#L30

So that has always been passed, I just added another key, `opt_name`, and started using it. 👍

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: Added this to the "Require Verified Email" setting and mentioned it here.

lib/admin/WP_Auth0_Admin_Generic.php
joshcanhelp: Slept on it ... this is goofy. Removing.

lib/admin/WP_Auth0_Admin_Advanced.php
lbalmaceda: How can users be "added to wordpress" if "wordpress registrations are off"?

joshcanhelp: You add in Auth0, successfully login, then they are added to WP

lib/admin/WP_Auth0_Admin_Advanced.php
lbalmaceda: @joshcanhelp But is the "wordpress user" created in that case?

joshcanhelp: No. No user created, no login allowed

lib/admin/WP_Auth0_Admin_Generic.php
lbalmaceda: > A translate-able text string that just starts with a period and a space is hard to re-use in any context.
This is true 100%. But then you can split `". The text to be translated" in `". "` and `"The text to be translated"`, fixing both what I pointed out and the translation inconvenience.

joshcanhelp: But then you're not able to change the period, if needed. Tradeoffs 🤷‍♂️

lbalmaceda:
How was that passed, if I see the method's signature is changing??

See line 137 doesn't include the parameter but now after the changes, that same line now receives `$args`

image

What I meant is that whoever was calling that public function in the past, would find that broken after this change.

lib/admin/WP_Auth0_Admin_Advanced.php
lbalmaceda: Is that expected? I understand there are 2 different user databases, with duplicated info IMO. What's the point on keeping a "auth0 user" that would never be able to "create a wordpress account"?

lib/admin/WP_Auth0_Admin_Generic.php
lbalmaceda: How so?

```
"Text with placeholder on my right" . <placeholder var> . ". " . "Other translatable text".
```

Something like this would work fine. Both the first and last texts could be translated just fine. The dot is a dot, it will never change at least not if you want to keep the original redaction of the message.

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: Looking through how this operates ... the scenario here is that WP registrations are turned off so the Lock form hides the **Signup** tab and the WP registration form is off. This means that no one can register themselves in WP (users can still be added by an admin). Now, an Auth0 admin adds a user in Auth0 and provides them their username and password. If the setting is turned:

- **off** (default) then the login is blocked (`throw new WP_Auth0_RegistrationNotEnabledException();` in `WP_Auth0_UsersRepo::create()`)
- **on** then the user is created as usual and life continues

I'm not saying this is *how* it should work, just how it currently does. IMHO, this should at least be turned on by default, maybe just forced to be the case.

PR #433, opened 2018-04-16
Change Basic settings field display; better admin UX

joshcanhelp:
- Refactored Basic settings with new render functions, new descriptions
- Fixed PHP < 5.4 array style for advanced render functions
- Persist tab showing on settings page 
- Universal show/hide JS for switch fields
- Constant for JWKS cache transient name

### DIFF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -85,7 +85,7 @@ jQuery(document).ready(function($) {
 
     function localStorageAvailable() {
         try {
-            var x = '__storage_test__';
+            var x = '__Auth0_localStorage_assertion__';
             window.localStorage.setItem(x, x);
             window.localStorage.removeItem(x);
             return true;


assets/js/admin.js
```diff
@@ -37,24 +38,60 @@ jQuery(document).ready(function($) {
...
 
+        $.post(wpa0.ajax_url, postData, function() {
+            $deleteCacheButton.prop( 'disabled', false ).val( wpa0.clear_cache_done );
+        }, 'json');
+    } );
+
+    function localStorageAvailable() {
+        try {
+            var x = '__storage_test__';
+            window.localStorage.setItem(x, x);
```
lbalmaceda: is this try/catch the only way to check for availability? And, can the `removeItem` fail after the first one succeeded?

lib/admin/WP_Auth0_Admin_Basic.php
```diff
@@ -32,153 +32,120 @@ public function init() {
...
-	?>
-      <input type="text" autocomplete="off" name="<?php echo $this->options->get_options_name(); ?>[client_secret]" id="wpa0_client_secret"  <?php if ( !empty( $v ) ) {?>placeholder="Not visible"<?php } ?> />
-      <div class="subelement">
-        <span class="description"><?php echo __( 'Application secret, copy from your application\'s settings in the', 'wp-auth0' ); ?> <a href="https://manage.auth0.com/#/applications" target="_blank">Auth0 dashboard</a>.</span>
-      </div>
-  <?php
+	public function render_client_secret_b64_encoded( $args = array() ) {
+		$this->render_switch( $args['label_for'], $args['opt_name'] );
+		$this->render_field_description(
+			__( 'Enable this if your Client Secret is base64 enabled. ', 'wp-auth0' ) .
```
lbalmaceda: base64 _encoded_

lib/admin/WP_Auth0_Admin_Basic.php
```diff
@@ -32,153 +32,120 @@ public function init() {
...
+		printf(
+			' <input type="button" id="auth0_delete_cache_transient" value="%s" class="button button-secondary">',
+			__( 'Delete Cache', 'wp-auth0' )
+		);
+		$this->render_field_description( __( 'JWKS cache expiration in minutes (use 0 for no caching)', 'wp-auth0' ) );
+		if ( $domain = $this->options->get( 'domain' ) ) {
+			$this->render_field_description(
+				sprintf( '<a href="https://%s/.well-known/jwks.json" target="_blank">%s</a>',
+					$domain,
+					__( 'View your JWKS here', 'wp-auth0' )
```
lbalmaceda: Would this section show the JWKs contents or what?

assets/js/admin.js
```diff
@@ -37,24 +38,60 @@ jQuery(document).ready(function($) {
...
-        // Hide/Show login method depending on auto login
-        var $loginMethodField = $("#wpa0_auto_login_method").closest("tr");
-        var $autoLoginCheckbox = $("#wpa0_auto_login");
-        if (!$autoLoginCheckbox.prop("checked")) {
-            $loginMethodField.hide();
-        }
+    // Show/hide field for specific switches
+    $('[data-expand!=""]').each( function() {
+        var $thisSwitch = $( this );
+        var $showFieldRow = $( '#' + $thisSwitch.attr( 'data-expand' ) ).closest( 'tr' );
```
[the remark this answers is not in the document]
joshcanhelp: Length is always there for a jQuery selector. Next line checks that

assets/js/admin.js
joshcanhelp: No, but it's the best way:

https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Feature-detecting_localStorage

As for `removeItem`, that would be a weird case, not implementing the webStorage interface. But localStorage persist forever and we don't want to store something we're never going to use.

lib/admin/WP_Auth0_Admin_Basic.php
joshcanhelp: Links to it, yes.

lib/admin/WP_Auth0_Admin_Basic.php
joshcanhelp: Why did I say "enabled" here? 😆 

Fixed!

assets/js/admin.js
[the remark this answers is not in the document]
lbalmaceda: I meant the "data-expand" and then "closest TR".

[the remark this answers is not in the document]
joshcanhelp: Having `data-expand` is what gets you into the `each` here. No `data-expand`, no `each`

assets/js/admin.js
lbalmaceda: is this stored under some domain key or can it collide with existing "x" stored keys? In that case, you might want to use some key like "a0_local_storage_assertion"

joshcanhelp: It's per-domain:

https://stackoverflow.com/q/4201239/728480

But you have a good point about that. WP sites could load a million other JS libraries.

Issue #430, opened 2018-04-11
Widget URL changes don't save when you are using passwordless


rdowinton:
If you have passwordless enabled and try to update the widget URL, the changes aren't saved.

This appears to be due to it using the `wpa0_passwordless_cdn_url` option in on the client side while reading the settings from the `passwordless_cdn_url` option instead.

joshcanhelp:
Fixed with #434 and merged into `dev`

PR #434, opened 2018-04-16
Fix Passwordless handling; update Lock instantiation

### LINKED ISSUE #400: CORS errors
Hi guys

I've got the plugin installed and can authenticate when i start chrome with no security (google-chrome --disable-web-security --user-data-dir)

However, when I run the plugin as normal it errors out when i use either password or passwordless logins.

In the chrome console i see:

`Failed to load https://<my-domain>.auth0.com/passwordless/verify: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://<my-domain>.com' is therefore not allowed access.`

Should I be editing my apache config to fix this? or does wordpress prefer I do something else?

joshcanhelp:
Main focus here was to update Lock Passwordless to use the combined library and move all instantiation JS to an external file, populated via `wpAuth0LockGlobal `

- Combined JS file to initiate Lock; removed JS from PHP template files
- Removing unused template files
- README support block, whitespace

Fixes #400, #430


README.md
```diff
@@ -27,8 +27,7 @@ We're happy to review and approve new filters and actions that help you integrat
 **Thank you in advance!**
```
joshcanhelp: Only line break changes here

templates/auth0-login-form-lock10.php
```diff
@@ -1,98 +0,0 @@
-<?php
```
joshcanhelp: Combined with `templates/auth0-login-form.php`

templates/auth0-login-form.php
```diff
@@ -1,212 +1,27 @@
 <?php
```
joshcanhelp: Modified to handle standard and Passwordless

templates/auth0-login-form.php
```diff
@@ -1,212 +1,27 @@
...
-<?php } ?>
-
-<script type="text/javascript">
-var ignore_sso = false;
-document.addEventListener("DOMContentLoaded", function() {
-
-    var callback = null;
-
-
-    <?php if ( $lock_options->get_auth0_implicit_workflow() ) { ?>
```
joshcanhelp: All this moved into a separate file in #426

templates/error-msg.php
```diff
@@ -1,34 +0,0 @@
-<?php if ( isset( $_GET['message'] ) ): ?>
```
joshcanhelp: Checked with @glena, this is likely not used anymore. No references in the plugin anymore and duplicates the `wp_die()` catches during login.

templates/auth0-login-form.php
```diff
@@ -1,212 +1,27 @@
...
-        a0ShowLoginModal();
-    <?php } else { ?>
-
-        jQuery('#a0LoginButton').click(a0ShowLoginModal);
-
-    <?php } ?>
-
-    if (lock.on) {
-        lock.on('ready', function(){
-            if ( lock.options['$client'].subscription === 'free' ) {
```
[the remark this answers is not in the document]
joshcanhelp: 👍

templates/error-msg.php
joshcanhelp: I believe it can be safely removed, yes. No other references to this and this functionality (minus the URL param) is part of the main login flow. There's nothing in the plugin that uses this functionality and I've never seen it during any testing thus far.

PR #435, opened 2018-04-16
Whitespace, method order, and linebreaks in lib/WP_Auth0_LoginManager.php

joshcanhelp:
No functional changes. Completed using:

```bash
phpcbf ./lib/WP_Auth0_LoginManager.php --standard=WordPress
```

... then manually adjusting where needed. Guidelines are here:

https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -2,618 +2,633 @@
 
 class WP_Auth0_LoginManager {
 
-  protected $a0_options;
-  protected $default_role;
-  protected $ignore_unverified_email;
-  protected $users_repo;
-  protected $state;
-  protected $state_decoded;
-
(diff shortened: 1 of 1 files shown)


lbalmaceda:
> then manually adjusting where needed
Do you remember what things you've changed? It won't escalate well if you run a code formatter and then do manual updates. Better to find how to update or tell the "run formatter command" to do those for you.

Also.. Is there a way to add this as a "pre-commit/release" action? The idea is that every time before a release you run this. Maybe you can add a bash script to run that, commit, and make a release commit.

joshcanhelp:
> Do you remember what things you've changed?

It was just the order of the functions, something that an automated formatter wouldn't change. Also removed several unnecessary empty lines. 

> Also.. Is there a way to add this as a "pre-commit/release" action?

Yes

PR #436, opened 2018-04-17
Change Feature settings output

joshcanhelp:
This commit will refactor `WP_Auth0_Admin_Appearance` to use the new
rendering funtions in `WP_Auth0_Admin_Generic`; change settings
descriptions to be more clear; fix broken, incorrect, or redirected
doc links; move passwordless switch to Features tab; reformat and
test password policy validation.

### DIFF
--- a/lib/admin/WP_Auth0_Admin_Features.php
+++ b/lib/admin/WP_Auth0_Admin_Features.php
@@ -39,12 +39,16 @@ class WP_Auth0_Admin_Features extends WP_Auth0_Admin_Generic {
         'id' => 'wpa0_password_policy', 'function' => 'render_password_policy' ),
       array( 'name' => __( 'Single Sign On (SSO)', 'wp-auth0' ), 'opt' => 'sso',
         'id' => 'wpa0_sso', 'function' => 'render_sso' ),
-      array( 'name' => __( 'Single Logout (SLO)', 'wp-auth0' ), 'opt' => 'singlelogout',
+      array( 'name' => __( 'Single Logout', 'wp-auth0' ), 'opt' => 'singlelogout',
         'id' => 'wpa0_singlelogout', 'function' => 'render_singlelogout' ),
+      array( 'name' => __( 'Passwordless Login', 'wp-auth0' ), 'opt' => 'passwordless_enabled',
+        'id' => 'wpa0_passwordless_enabled', 'function' => 'render_passwordless_enabled' ),
       array( 'name' => __( 'Multifactor Authentication (MFA)', 'wp-auth0' ), 'opt' => 'mfa',
         'id' => 'wpa0_mfa', 'function' => 'render_mfa' ),
       array( 'name' => __( 'FullContact Integration', 'wp-auth0' ), 'opt' => 'fullcontact',
         'id' => 'wpa0_fullcontact', 'function' => 'render_fullcontact' ),
+      array( 'name' => __( 'FullContact API Key', 'wp-auth0' ), 'opt' => 'fullcontact_apikey',
+        'id' => 'wpa0_fullcontact_key', 'function' => 'render_fullcontact_apikey' ),
       array( 'name' => __( 'Store Geolocation', 'wp-auth0' ), 'opt' => 'geo_rule',
         'id' => 'wpa0_geo', 'function' => 'render_geo' ),
       array( 'name' => __( 'Store Zipcode Income', 'wp-auth0' ), 'opt' => 'income_rule',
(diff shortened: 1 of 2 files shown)


lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -55,131 +59,102 @@ public function init() {
...
-      <input type="radio" name="<?php echo $this->options->get_options_name(); ?>[password_policy]" id="wpa0_password_policy_excellent" value="excellent" <?php echo checked( $v, 'excellent', false ); ?>/><label for="wpa0_password_policy_excellent">Excellent</label>
-      <div class="subelement">
-        <span class="description">
-          <?php echo __( 'The difficulty of the user password where \'none\' requires a single character, \'low\' requires six characters and so on. For more details see our', 'wp-auth0' ); ?> <a target="_blank" href="https://auth0.com/docs/password-strength"><?php echo __( 'help page', 'wp-auth0' ); ?></a> <?php echo __( 'on password difficulty.', 'wp-auth0' ); ?>
-        </span>
-      </div>
-    <?php
+  public function render_password_policy( $args = array() ) {
+    $curr_val = $this->options->get( $args['opt_name'] );
+    $this->render_radio_button( $args['label_for'] . '_none', $args['opt_name'], '', 'None', empty( $curr_val ) );
```
lbalmaceda: I'm not sure about this. But check if once the policy is enabled, if turning it OFF and then ON again still removes the policy object from the JSON or if it now does send a "none" value.

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -55,131 +59,102 @@ public function init() {
...
-        <span class="description">
-          <?php echo __( 'Single Sign On (SSO) allows users to sign in once to multiple services. For more details, see our ', 'wp-auth0' ); ?>
-          <a target="_blank" href="https://auth0.com/docs/sso/single-sign-on"><?php echo __( 'help page on SSO', 'wp-auth0' ); ?></a>.
-        </span>
-      </div>
-    <?php
+  public function render_sso( $args = array() ) {
+    $this->render_switch( $args['label_for'], $args['opt_name'] );
+    $this->render_field_description(
+      __( 'SSO allows users to sign in once to multiple Clients in the same tenant. ', 'wp-auth0' ) .
```
lbalmaceda: Clients or Applications?

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -55,131 +59,102 @@ public function init() {
...
-      </div>
-    <?php
+  public function render_mfa( $args = array() ) {
+    $this->render_switch( $args['label_for'], $args['opt_name'] );
+    $this->render_field_description(
+      __( 'Mark this if you want to enable multifactor authentication with Auth0 Guardian. ', 'wp-auth0' ) .
+      sprintf(
+        __( 'You can enable other MFA providers in the %s. ', 'wp-auth0' ),
+        $this->get_dashboard_link( 'multifactor' )
+      ) . __( 'For more information, see our %s. ', 'wp-auth0' ) .
```
lbalmaceda: This `%s` is not inside the `sprintf` scope so I think there's no substitution happening here, am I right?

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -55,131 +59,102 @@ public function init() {
...
-          <?php echo __( 'Mark this if you want to enrich your users\' profiles with the data provided by FullContact. A valid api key is required. ', 'wp-auth0' ); ?>
-          <?php echo __( 'For more information, see our ', 'wp-auth0' ); ?>
-          <a target="_blank" href="https://auth0.com/docs/scenarios/mixpanel-fullcontact-salesforce#2-augment-user-profile-with-fullcontact-"><?php echo __( 'help page on FullContact integration with Auth0', 'wp-auth0' );?></a>
-        </span>
-      </div>
-    <?php
+  public function render_passwordless_enabled( $args = array() ) {
+    $this->render_switch( $args[ 'label_for' ], $args[ 'opt_name' ] );
+    $this->render_field_description(
+      __( 'Turn on Passwordless login (email or SMS) in the Auth0 form. ', 'wp-auth0' ) .
```
lbalmaceda: What about the `MFA` grant? that needs to be enabled for the Application too, at least when calling this from an SDK. Test if creating a new Application leaves that grant enabled/disabled and if you can log in successfully using this wp-auth0

lib/admin/WP_Auth0_Admin_Features.php
joshcanhelp: I tested it back and forth between `none` and other values and everything works 👍

lib/admin/WP_Auth0_Admin_Features.php
joshcanhelp: No special grant needed for Passwordless, uses the Auth Code grant.

PR #437, opened 2018-04-17
Improved code quality in WP_Auth0_LoginManager

joshcanhelp:
Very minimal functional changes, mostly comments. 

- Added docblocks, parameter + error descriptions, and better inline comments
- Renamed internal variables to conform to snake_case and better describe what they're for
- Changed `urlencode` -> `rawurlencode` per `phpcs` suggestion (RFC-compliant encoding, improves interoperability)
- Changed `die()` -> `exit` for standardization (identical functionality)
- Changed `json_encode()` to `wp_ json_encode()` to catch more problematic encoding

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -1,21 +1,78 @@
 <?php
+/**
+ * WP_Auth0_LoginManager class
+ *
+ * @package WordPress
+ * @subpackage WP-Auth0
+ * @since 2.0.0
+ */
 
+/**
+ * Handles login callbacks and auto-login redirecting
+ *
+ * @since 2.0.0
+ */
 class WP_Auth0_LoginManager {
 
+	/**
+	 * Instance of WP_Auth0_Options.
+	 *
+	 * @var null|WP_Auth0_Options
+	 */
 	protected $a0_options;
-	protected $default_role;
+
+	/**
+	 * Should the new user have an administrator role?
+	 *
+	 * @var bool|null
+	 */
+	protected $admin_role;
+
+	/**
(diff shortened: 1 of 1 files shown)


lib/WP_Auth0_LoginManager.php
```diff
@@ -61,38 +126,43 @@ public function login_auto() {
 			// Create the link to log in.
 			$login_url = 'https://' . $this->a0_options->get( 'domain' ) .
 						'/authorize?' .
-						'scope=' . urlencode( $options['auth']['params']['scope'] ) .
```
cocojoe: For others reading: https://stackoverflow.com/questions/996139/urlencode-vs-rawurlencode

lib/WP_Auth0_LoginManager.php
```diff
@@ -481,29 +564,36 @@ public function auth0_sso_footer( $previous_html ) {
...
-
-		if ( ! $singlelogout ) {
-			return;
+	/**
+	 * Outputs JS on all pages to log a user out if no Auth0 session is found.
+	 * Hooked to `wp_footer` action.
+	 *
+	 * @see WP_Auth0_LoginManager::init()
+	 */
+	public function auth0_singlelogout_footer() {
```
cocojoe: you have changed the signature here, sneaky. This looks like an internal method though so fine.

joshcanhelp: It's hooked to the `wp_footer` action, which does not pass any arguments to that function so that wasn't doing anything anyways. I was talking about this with @lbalmaceda, these callbacks have to be public to be hooked but I can't think of a use case where someone would call this directly, let alone have any use for an already non-functioning parameter.

PR #439, opened 2018-04-19
Change Appearance tab settings output

joshcanhelp:
- Refactor Appearance settings with new render functions, new descriptions
- Fix `avatar` Lock setting
- Minor refactor of `WP_Auth0_Admin_Features::render_password_policy()` for consistency

### DIFF
--- a/lib/admin/WP_Auth0_Admin_Appearance.php
+++ b/lib/admin/WP_Auth0_Admin_Appearance.php
@@ -110,10 +110,7 @@ class WP_Auth0_Admin_Appearance extends WP_Auth0_Admin_Generic {
 		$this->render_radio_button( $id_attr . '_un', $opt_name, 'username', '', 'username' === $value );
 		$this->render_field_description(
 			__( 'To allow the user to use either email or username to login, leave this as "Auto." ', 'wp-auth0' ) .
-			$this->get_docs_link(
-				'libraries/lock/customization#usernamestyle-string-',
-				__( 'More information on this setting', 'wp-auth0'	)
-			)
+			__( 'Only database connections that require a username will allow username logins', 'wp-auth0' )
 		);
 	}
 
--- a/lib/admin/WP_Auth0_Admin_Appearance.php
+++ b/lib/admin/WP_Auth0_Admin_Appearance.php
@@ -75,7 +75,7 @@ class WP_Auth0_Admin_Appearance extends WP_Auth0_Admin_Generic {
 			__( 'Primary color for the Auth0 login form in hex format. ', 'wp-auth0' ) .
 			$this->get_docs_link(
 				'libraries/lock/v11/configuration#primarycolor-string-',
-				__( 'More information this settings', 'wp-auth0'	)
+				__( 'More information on this settings', 'wp-auth0'	)
 			)
 		);
 	}
@@ -86,7 +86,7 @@ class WP_Auth0_Admin_Appearance extends WP_Auth0_Admin_Generic {
 			__( 'The languageDictionary parameter for the Auth0 login form. ', 'wp-auth0' ) .
 			sprintf(
 				'<a href="https://github.com/auth0/lock/blob/master/src/i18n/en.js" target="_blank">%s</a>',
-				__( 'List of all modifiable text', 'wp-auth0' )
+				__( 'List of all modifiable options', 'wp-auth0' )
 			)
 		);
 	}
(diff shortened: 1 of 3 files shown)


lib/admin/WP_Auth0_Admin_Appearance.php
```diff
@@ -53,117 +53,91 @@ public function init() {
...
-?>
-      <textarea name="<?php echo $this->options->get_options_name(); ?>[custom_js]" id="wpa0_custom_js"><?php echo esc_attr( $v ); ?></textarea>
-      <div class="subelement">
-        <span class="description"><?php echo __( 'This should be a valid JS to customize the Auth0 login widget to, for example, add custom buttons. ', 'wp-auth0' ); ?><a target="_blank" href="https://auth0.com/docs/hrd#option-3-adding-custom-buttons-to-lock"><?php echo __( 'More info', 'wp-auth0' ); ?></a></span>
-      </div>
-    <?php
+	public function render_primary_color( $args = array() ) {
+		$this->render_text_field( $args[ 'label_for' ], $args[ 'opt_name' ] );
+		$this->render_field_description(
+			__( 'Primary color for the Auth0 login form in hex format. ', 'wp-auth0' ) .
```
lbalmaceda: should hex be uppercase? (I don't know either)

lib/admin/WP_Auth0_Admin_Appearance.php
```diff
@@ -53,117 +53,91 @@ public function init() {
...
+			$this->get_docs_link(
+				'libraries/lock/v11/configuration#primarycolor-string-',
+				__( 'More information this settings', 'wp-auth0'	)
```
lbalmaceda: > More information this settings
????

lib/admin/WP_Auth0_Admin_Appearance.php
```diff
@@ -53,117 +53,91 @@ public function init() {
...
-        </span>
-      </div>
-    <?php
+	public function render_language_dictionary( $args = array() ) {
+		$this->render_textarea_field( $args[ 'label_for' ], $args[ 'opt_name' ] );
+		$this->render_field_description(
+			__( 'The languageDictionary parameter for the Auth0 login form. ', 'wp-auth0' ) .
+			sprintf(
+				'<a href="https://github.com/auth0/lock/blob/master/src/i18n/en.js" target="_blank">%s</a>',
+				__( 'List of all modifiable text', 'wp-auth0' )
```
lbalmaceda: > List of all modifiable text

text ... options? or something that describes that this are items that change how the form is displayed.

lib/admin/WP_Auth0_Admin_Appearance.php
```diff
@@ -53,117 +53,91 @@ public function init() {
...
-		$v = absint( $this->options->get( 'gravatar' ) );
+	public function render_username_style( $args = array() ) {
+	    $opt_name = $args[ 'opt_name' ];
+	    $id_attr = $args[ 'label_for' ];
+		$value = $this->options->get( $opt_name );
+		$this->render_radio_button( $id_attr . '_au', $opt_name, '', 'Auto', empty( $value ) );
+		$this->render_radio_button( $id_attr . '_em', $opt_name, 'email', '', 'email' === $value );
+		$this->render_radio_button( $id_attr . '_un', $opt_name, 'username', '', 'username' === $value );
+		$this->render_field_description(
+			__( 'To allow the user to use either email or username to login, leave this as "Auto." ', 'wp-auth0' ) .
```
lbalmaceda: Because by reading this I get that "auto" would mean either email or username, but if the db connection has not "requires username = true" then this would only accept email.

lib/admin/WP_Auth0_Admin_Appearance.php
```diff
@@ -53,117 +53,91 @@ public function init() {
...
-      <div class="subelement">
-        <span class="description">
-          <?php echo __( 'Read more about the gravatar integration ', 'wp-auth0' ); ?>
-          <a target="_blank" href="https://auth0.com/docs/libraries/lock/customization#gravatar-boolean-"><?php echo __( 'HERE', 'wp-auth0' ); ?></a></span>
-      </div>
-    <?php
+	public function render_gravatar( $args = array() ) {
+		$this->render_switch( $args[ 'label_for' ], $args[ 'opt_name' ] );
+		$this->render_field_description(
+			__( 'Automatically display avatar on login form when email address is entered', 'wp-auth0' )
```
lbalmaceda: explain this are obtained from gravatar

lib/admin/WP_Auth0_Admin_Appearance.php
joshcanhelp: It's like you read these things over and over and still ...

lib/admin/WP_Auth0_Admin_Appearance.php
joshcanhelp: @lbalmaceda - I was a little confused because the "requires username" setting on the DB connection was taking a while to populate on the Lock side but I see now that Lock does do the right thing even when this is wrong. I'll re-word.

PR #441, opened 2018-04-20
Require telemetry for API calls

joshcanhelp:
Changing API header calls to include telemetry always. Also adding server and mysql information.

### DIFF
--- a/lib/WP_Auth0_Api_Client.php
+++ b/lib/WP_Auth0_Api_Client.php
@@ -147,25 +147,21 @@ class WP_Auth0_Api_Client {
 		return true;
 	}
 
+	/**
+	 * Get required telemetry header
+	 *
+	 * @return array
+	 */
 	public static function get_info_headers() {
-		global $wp_version;
-
-		$a0_options = WP_Auth0_Options::Instance();
-
-		if ( $a0_options->get( 'metrics' ) != 1 ) {
-			return array();
-		}
-
-		return array(
-			'Auth0-Client' => base64_encode( wp_json_encode( array(
-						'name' => 'wp-auth0',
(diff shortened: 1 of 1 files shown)


lib/WP_Auth0_Api_Client.php
```diff
@@ -147,25 +147,37 @@ public static function validate_user_token( $app_token ) {
...
-						'version' => WPA0_VERSION,
-						'environment' => array(
-							'PHP' => phpversion(),
-							'WordPress' => $wp_version,
-						)
-					) ) )
+		$header_value = array(
+			'name' => 'wp-auth0',
+			'version' => WPA0_VERSION,
+			'environment' => array(
```
lbalmaceda: I'm not sure we support a non-plain object (this `environment` dictionary). Looking at the previous implementation it seems we do, but may be worth checking if this is easy to query later on the BI app.

joshcanhelp:
@lbalmaceda - Understood and I agree with everything you said. This PR, though, is only to require this data to be sent, not to refactor the entire telemetry system before we know more about it (I have a request in to the data team now).

lbalmaceda:
I understand that this PR won't refactor that. However, note that I'm not talking about the _contents_ of the telemetry. I'm pointing out how the _method_ is working today and how it should be changed later. Keeping the method like it is today would require too many wrap-unwrap like operations to just obtain a value. I'll approve this PR, but please track this issue somewhere (new gh issue?).

joshcanhelp:
@lbalmaceda - Understood. This will be addressed once we understand how the telemetry is received and what more we want to (and can) add to it. Good call on tracking, I'll add that as a story.

PR #446, opened 2018-04-23
Deprecate 2 lookup methods

joshcanhelp:
- WP_Auth0_Options::get_enabled_connections()
- WP_Auth0::get_plugin_dir_url()

### DIFF
--- a/lib/WP_Auth0_Options.php
+++ b/lib/WP_Auth0_Options.php
@@ -19,11 +19,6 @@ class WP_Auth0_Options extends WP_Auth0_Options_Generic {
 		return get_site_option( 'users_can_register', 0 ) == 1;
 	}
 
-	// TODO: Deprecate, not used
-	public function get_enabled_connections() {
-		return array( 'facebook', 'twitter', 'google-oauth2' );
-	}
-
 	public function set_connection( $key, $value ) {
 		$options = $this->get_options();
 		$options['connections'][$key] = $value;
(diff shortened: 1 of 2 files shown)


lib/WP_Auth0_Options.php
```diff
@@ -19,8 +19,13 @@ public function is_wp_registration_enabled() {
 		return get_site_option( 'users_can_register', 0 ) == 1;
 	}
 
-	// TODO: Deprecate, not used
+	/**
+	 * @deprecated 3.6.0 - Not used
```
lbalmaceda: explain why?

joshcanhelp:
@lbalmaceda - beautification is a bit of a WIP but I will be running the whitespace one once all the PRs are in place.

lib/WP_Auth0_Options.php
```diff
@@ -252,4 +247,15 @@ protected function defaults() {
...
+
+	/**
+	 * @deprecated 3.6.0 - Social connections are no longer set during initial setup so this data is no longer needed. 
+	 *
+	 * @return array
+	 */
+	public function get_enabled_connections() {
+		// phpcs:ignore
+		trigger_error( sprintf( __( 'Method %s is deprecated.', 'wp-auth0' ), __METHOD__ ), E_USER_DEPRECATED );
+		return array( 'facebook', 'twitter', 'google-oauth2' );
```
lbalmaceda: is weird that this was (and is) hardcoding an array of enabled connections. I've read the deprecation message. However, are you currently calling this method somewhere in your plugin?

joshcanhelp: I didn't remove it so I'm not entirely sure when or where this was used. I checked through all the functions here and did not see it used anywhere in the code. I think this was part of the initial setup at some point? 

This is the case with a number of this deprecations. I didn't remove them and didn't see them used anywhere. If we're not using the method internally, I don't want to maintain it 👍

PR #447, opened 2018-04-23
Only show email verification resend for DB connections

### LINKED ISSUE #345: Provide Resend verification email only for DB connections
AD/LDAP connections (and maybe others) can have unverified email addresses, but the verify email action only works for DB connections. So the [link to **Resend verification email**](https://github.com/auth0/wp-auth0/blob/7f6c43dad23e3703d63555febbc309c49a53c4b8/templates/verify-email.php) should only be shown to users coming from a DB connection. 

API v1's `send_verification_email` does not check that, but the link sent by email will cause an error when the user tries to use it.

joshcanhelp:
Looks for `auth0` strategy on the user ID and only shows the "Resend verification email" link if there. 

Fixes #345

### DIFF
--- a/lib/WP_Auth0_Email_Verification.php
+++ b/lib/WP_Auth0_Email_Verification.php
@@ -20,30 +20,36 @@ class WP_Auth0_Email_Verification {
 	 * @param object $userinfo
 	 */
 	public static function render_die( $userinfo ) {
+		$user_id = isset( $userinfo->user_id ) ? $userinfo->user_id : $userinfo->sub;
 
-		$html = sprintf(
-			'<p>%s<br><a id="js-a0-resend-verification" href="#">%s</a></p>
+		$html = sprintf( '<p>%s</p>', __( 'This site requires a verified email address. ', 'wp-auth0' ) );
+
+		// Only provide resend verification link for DB connection users
+		if ( 0 === strpos( $user_id, 'auth0|' ) ) {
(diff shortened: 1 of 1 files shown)


lib/WP_Auth0_Email_Verification.php
```diff
@@ -20,30 +20,36 @@ public static function init() {
...
-			'//code.jquery.com/jquery-1.12.4.js',
-			WPA0_PLUGIN_URL . 'assets/js/die-with-verify-email.js?ver=' . WPA0_VERSION
-		);
+				__( 'Resend verification email.', 'wp-auth0' ),
+				wp_login_url(),
+				time(),
+				__( '← Login', 'wp-auth0' ),
+				esc_url( admin_url( 'admin-ajax.php' ) ),
+				esc_js( $user_id ),
+				esc_js( wp_create_nonce( self::RESEND_NONCE_ACTION ) ),
```
lbalmaceda: I guess you're using the [jobs re-send email endpoint](https://auth0.com/docs/api/management/v2#!/Jobs/post_verification_email). The request only requires `user_id` and `client_id` properties. What is the nonce for?

joshcanhelp: @lbalmaceda - WP thing to make sure that POSTed data is coming from the site. It's validated in the AJAX callback.

PR #448, opened 2018-04-24
Encode auth0Client in login redirect URL

pnu:
Apply same encoding as in logout(). Not encoding will result this parameter to have value "Array".

Not encoding also outputs a PHP warning: `Notice: Array to string conversion in …/WP_Auth0_LoginManager.php on line 151`.

Outputting this notice to the output buffer causes another warning: `Warning: Cannot modify header information - headers already sent …`.

This means that the login does not work, unless the warning are suppressed (as usually is the case for production.) However, this is an issue for development environments when warnings are enabled.

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -96,7 +96,8 @@ class WP_Auth0_LoginManager {
     }
 
     if ( $sso ) {
-      wp_redirect( 'https://' . $this->a0_options->get( 'domain' ) . '/v2/logout?federated&returnTo=' . urlencode( home_url() ) . '&client_id='.$client_id.'&auth0Client=' . base64_encode( json_encode( WP_Auth0_Api_Client::get_info_headers() ) ) );
+      $telemetry = WP_Auth0_Api_Client::get_info_headers();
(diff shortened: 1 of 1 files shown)


joshcanhelp:
@pnu - Thanks so much for the contribution here!

I was actually just in the process of refactoring this a bit and found this issue as well. It's actually one more step to get this to work right as that static method returns the correct information but in the wrong format. You want the value for `Auth0-Client` array that's returned, which is already encoded. 

You can see that in the branch I'm working on here (code is not reviewed or thoroughly tested quite yet):

https://github.com/auth0/wp-auth0/blob/change-auto-login-refactor/lib/WP_Auth0_LoginManager.php#L616

pnu:
Great, thanks! I was also wondering if double-encoding is really what's wanted, but that was already done that way in `logout()` (same file), so I ended up copying it from there. I'm assuming both (login/logout) should use the same format for this parameter?

I'm not sure how soon your branch is landing, so in case you want to merge this PR before that, here are the corrected changes (now including the same change also for `logout()`).

This would be otherwise mostly harmless issue (except auth0 not getting correct stats), but in a development environment the emitted PHP warning (`Notice: Array to string conversion …`) stops the login, because one can't modify headers (ie. do the redirect) after outputting the body (ie. the warning notice).

PR #449, opened 2018-04-24
Change auto-login action

joshcanhelp:
- Change auto-login logic to be more clear and explicit
- Change implicit to handle nonce correctly
- Add authorize URL param builder
- Fix callback error handling for implicit and redirect

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -102,43 +102,34 @@ class WP_Auth0_LoginManager {
 	 * Redirect to a specific connection designated in Settings > Advanced
 	 */
 	public function login_auto() {
-		$auto_login = absint( $this->a0_options->get( 'auto_login' ) );
-
-		if ( $auto_login && ( ! isset( $_GET['action'] ) || 'logout' !== $_GET['action'] ) && ! isset( $_GET['wle'] ) ) {
-
-			if ( strtolower( $_SERVER['REQUEST_METHOD'] ) !== 'get' ) {
-				return;
-			}
-
-			if ( $this->query_vars( 'auth0' ) !== null ) {
-				return;
-			}
-
-			$lock_options = new WP_Auth0_Lock10_Options();
-			$options      = $lock_options->get_lock_options();
-
-			$connection = apply_filters( 'auth0_get_auto_login_connection', $this->a0_options->get( 'auto_login_method' ) );
-
-			$response_type = $lock_options->get_auth0_implicit_workflow() ? 'id_token' : 'code';
-
-			$state = $lock_options->get_state_obj();
-
-			// Create the link to log in.
-			$login_url = 'https://' . $this->a0_options->get( 'domain' ) .
-						'/authorize?' .
-						'scope=' . rawurlencode( $options['auth']['params']['scope'] ) .
-						'&response_type=' . $response_type .
-						'&client_id=' . $this->a0_options->get( 'client_id' ) .
-						'&redirect_uri=' . $options['auth']['redirectUrl'] .
-						'&state=' . rawurlencode( $state ) .
-						'&nonce=' . WP_Auth0_Nonce_Handler::getInstance()->get() .
-						'&connection=' . trim( $connection ) .
-						'&auth0Client=' . WP_Auth0_Api_Client::get_info_headers();
-
-			setcookie( WPA0_STATE_COOKIE_NAME, $state, time() + WP_Auth0_Nonce_Handler::COOKIE_EXPIRES, '/' );
-			wp_redirect( $login_url );
-			exit;
+		if (
+			// Nothing to do
+			( ! $this->a0_options->get( 'auto_login', FALSE ) )
+			// Auth0 is not ready to process logins
+			|| ! WP_Auth0::ready()
+			// Do not redirect POST requests
+			|| strtolower( $_SERVER['REQUEST_METHOD'] ) !== 'get'
+			// Do not redirect login page override
+			|| isset( $_GET['wle'] )
+			// Do not redirect log out action
+			|| ( isset( $_GET['action'] ) && 'logout' === $_GET['action'] )
+			// Do not redirect Auth0 login processing
+			|| null !== $this->query_vars( 'auth0' )
+			// Do not redirect if already authenticated
+			|| is_user_logged_in()
+		) {
+			return;
 		}
+
+		$connection = apply_filters( 'auth0_get_auto_login_connection', $this->a0_options->get( 'auto_login_method' ) );
+		$auth_params = self::get_authorize_params( $connection );
+
+		$auth_url = 'https://' . $this->a0_options->get( 'domain' ) . '/authorize';
+		$auth_url = add_query_arg( array_map( 'rawurlencode', $auth_params ), $auth_url );
+
+		setcookie( WPA0_STATE_COOKIE_NAME, $auth_params['state'], time() + WP_Auth0_Nonce_Handler::COOKIE_EXPIRES, '/' );
+		wp_redirect( $auth_url );
+		exit;
 	}
 
 	/**
@@ -156,8 +147,8 @@ class WP_Auth0_LoginManager {
 		// Catch any incoming errors and stop the login process.
 		// See https://auth0.com/docs/libraries/error-messages for more info.
 		if ( $this->query_vars( 'error' ) || $this->query_vars( 'error_description' ) ) {
-			$error_msg  = sanitize_text_field( $this->query_vars( 'error_description' ) );
-			$error_code = sanitize_text_field( $this->query_vars( 'error' ) );
+			$error_msg  = sanitize_text_field( rawurldecode( $_REQUEST[ 'error_description' ] ) );
+			$error_code = sanitize_text_field( $_REQUEST[ 'error' ]  );
 			$this->die_on_login( $error_msg, $error_code );
 		}
 
@@ -332,13 +323,11 @@ class WP_Auth0_LoginManager {
 		}
 
 		// Validate the nonce if one was included in the request if using auto-login.
-		if ( $this->a0_options->get( 'auto_login' ) ) {
-			$nonce = isset( $decoded_token->nonce ) ? $decoded_token->nonce : null;
-			if ( ! WP_Auth0_Nonce_Handler::getInstance()->validate( $nonce ) ) {
-				throw new WP_Auth0_LoginFlowValidationException(
-					__( 'Invalid nonce', 'wp-auth0' )
-				);
-			}
+		$nonce = isset( $decoded_token->nonce ) ? $decoded_token->nonce : null;
+		if ( ! WP_Auth0_Nonce_Handler::getInstance()->validate( $nonce ) ) {
+			throw new WP_Auth0_LoginFlowValidationException(
+				__( 'Invalid nonce', 'wp-auth0' )
+			);
 		}
 
 		// Populate legacy userinfo property.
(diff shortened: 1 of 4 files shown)


lib/WP_Auth0_LoginManager.php
```diff
@@ -156,8 +147,8 @@ public function init_auth0() {
 		// Catch any incoming errors and stop the login process.
 		// See https://auth0.com/docs/libraries/error-messages for more info.
 		if ( $this->query_vars( 'error' ) || $this->query_vars( 'error_description' ) ) {
-			$error_msg  = sanitize_text_field( $this->query_vars( 'error_description' ) );
-			$error_code = sanitize_text_field( $this->query_vars( 'error' ) );
+			$error_msg  = sanitize_text_field( rawurldecode( $_REQUEST[ 'error_description' ] ) );
```
joshcanhelp: `$this->query_vars()` checks for WP query vars first, which was leaving out the `error` parameter for some reason. Switched to a more explicit check and now pulls `error` more consistently.

lib/WP_Auth0_LoginManager.php
```diff
@@ -332,13 +323,11 @@ public function implicit_login() {
...
 
 		// Validate the nonce if one was included in the request if using auto-login.
-		if ( $this->a0_options->get( 'auto_login' ) ) {
-			$nonce = isset( $decoded_token->nonce ) ? $decoded_token->nonce : null;
-			if ( ! WP_Auth0_Nonce_Handler::getInstance()->validate( $nonce ) ) {
-				throw new WP_Auth0_LoginFlowValidationException(
-					__( 'Invalid nonce', 'wp-auth0' )
-				);
-			}
+		$nonce = isset( $decoded_token->nonce ) ? $decoded_token->nonce : null;
```
joshcanhelp: Nonce now works for any implicit login 🎉

lib/WP_Auth0_LoginManager.php
```diff
@@ -587,6 +576,68 @@ public function end_session() {
...
 	}
 
+	/**
+	 * Get and filter the scope used for access and ID tokens
+	 *
+	 * @param string $context - how are the scopes being used?
+	 *
+	 * @return string
+	 */
+	public static function get_userinfo_scope( $context = '' ) {
```
joshcanhelp: Moved default scopes to here so they can be filtered

lib/WP_Auth0_Nonce_Handler.php
```diff
@@ -12,7 +12,7 @@
   /**
    *
    */
-  const COOKIE_EXPIRES = MINUTE_IN_SECONDS;
+  const COOKIE_EXPIRES = HOUR_IN_SECONDS;
```
joshcanhelp: Longer state cookie, was failing if you left the login page open for more than a minute.

lib/WP_Auth0_LoginManager.php
joshcanhelp: Also `$context` is only for that filter. I pass in a context when I request those scopes so the filter function can decide to do something different based on what's being asked.

lib/WP_Auth0_LoginManager.php
lbalmaceda: Why does this call `rawurldecode` first when the line below doesn't?

lib/WP_Auth0_LoginManager.php
```diff
@@ -332,13 +323,11 @@ public function implicit_login() {
...
 		// Validate the nonce if one was included in the request if using auto-login.
-		if ( $this->a0_options->get( 'auto_login' ) ) {
-			$nonce = isset( $decoded_token->nonce ) ? $decoded_token->nonce : null;
-			if ( ! WP_Auth0_Nonce_Handler::getInstance()->validate( $nonce ) ) {
-				throw new WP_Auth0_LoginFlowValidationException(
-					__( 'Invalid nonce', 'wp-auth0' )
-				);
-			}
+		$nonce = isset( $decoded_token->nonce ) ? $decoded_token->nonce : null;
+		if ( ! WP_Auth0_Nonce_Handler::getInstance()->validate( $nonce ) ) {
```
lbalmaceda: Will this work fine (fail) if the nonce was not found (null) in the id token and the flow was implicit?

lib/WP_Auth0_LoginManager.php
lbalmaceda: how does this method work? I saw you call it passing `sso` and `lock`. The method should have a single known behavior. What's the input/output?

lib/WP_Auth0_LoginManager.php
```diff
@@ -587,6 +576,68 @@ public function end_session() {
...
+	 * @param string $context - how are the scopes being used?
+	 *
+	 * @return string
+	 */
+	public static function get_userinfo_scope( $context = '' ) {
+		return apply_filters( 'auth0_auth_token_scope', 'openid email name nickname picture', $context );
+	}
+
+	/**
+	 * Get a base authorize URL for handling HLP redirects.
```
lbalmaceda: HLP -> Universal Login Page

lib/WP_Auth0_LoginManager.php
```diff
@@ -587,6 +576,68 @@ public function end_session() {
...
+	 *
+	 * @param null|string $connection - specific connection to use, pass null to use all enabled.
```
lbalmaceda: I think it reads better this way:

```
specific connection to use, pass null to use all the enabled connections.
```

lib/WP_Auth0_LoginManager.php
```diff
@@ -587,6 +576,68 @@ public function end_session() {
...
+	 * @param null|string $redirect_to - URL to redirect upon successful authentication.
```
lbalmaceda: shouldn't this be required? where is the result taken to if not?

lib/WP_Auth0_LoginManager.php
```diff
@@ -587,6 +576,68 @@ public function end_session() {
...
+	public static function get_authorize_params( $connection = null, $redirect_to = null ) {
+		$params = array();
+		$options = WP_Auth0_Options::Instance();
+		$lock_options = new WP_Auth0_Lock10_Options();
+		$is_implicit = (bool) $options->get( 'auth0_implicit_workflow', FALSE );
+		$nonce = WP_Auth0_Nonce_Handler::getInstance()->get();
+
+		$params[ 'client_id' ] = $options->get( 'client_id' );
+		$params[ 'scope' ] = self::get_userinfo_scope( 'authorize_url' );
+		$params[ 'response_type' ] = $is_implicit ? 'id_token': 'code';
```
lbalmaceda: I think this value should be changeable by the user. You can also [accept combinations](https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#MultiValueResponseTypes) like `id_token code`

lib/WP_Auth0_LoginManager.php
```diff
@@ -587,6 +576,68 @@ public function end_session() {
...
+		$params[ 'redirect_uri' ] = $is_implicit
```
lbalmaceda: why does the `redirect_uri` change when using implicit flow?

lib/WP_Auth0_LoginManager.php
```diff
@@ -587,6 +576,68 @@ public function end_session() {
...
+			? $lock_options->get_implicit_callback_url()
+			: $options->get_wp_auth0_url( null, $is_implicit );
```
lbalmaceda: why use `$is_implicit` if you already know the value?

lib/WP_Auth0_LoginManager.php
```diff
@@ -587,6 +576,68 @@ public function end_session() {
...
+		if ( $is_implicit ) {
+			$params[ 'nonce' ] = $nonce;
+		}
+
+		if ( ! empty( $connection ) ) {
+			$params[ 'connection' ] = $connection;
+		}
+
+		// Get the telemetry header.
+		$telemetry = WP_Auth0_Api_Client::get_info_headers();
```
lbalmaceda: this object generates headers with lots of information, right? why use only Auth0-Client? what about the rest?

lib/WP_Auth0_LoginManager.php
```diff
@@ -587,6 +576,68 @@ public function end_session() {
...
+		$params[ 'auth0Client' ] = $telemetry[ 'Auth0-Client' ];
+
+		// Where should the user be redirected after logging in?
+		if ( empty( $redirect_to ) && ! empty( $_GET['redirect_to'] ) ) {
```
lbalmaceda: Line 608 talks about a `redirect_uri`. What would be the difference with this `redirect_to` parameter, and shouldn't they be merged or at least be closer? (lines distance in this file)

lib/WP_Auth0_LoginManager.php
```diff
@@ -587,6 +576,68 @@ public function end_session() {
...
+			$redirect_to = $_GET['redirect_to'];
+		} elseif ( empty( $redirect_to ) ) {
+			$redirect_to = $options->get( 'default_login_redirection' );
+		}
+
+		// State parameter, checked during login callback.
+		$params[ 'state' ] = base64_encode( json_encode( array(
+			'interim' => false,
+			'nonce' => $nonce,
+			'redirect_to' => filter_var( $redirect_to, FILTER_SANITIZE_URL ),
```
lbalmaceda: why do you need to manipulate the `redirect_to` value? can't you use directly the one defined just a few lines above this?

lib/WP_Auth0_Options.php
```diff
@@ -108,12 +108,13 @@ public function get_web_origins() {
...
 	 * @param string|null $protocol - forced URL protocol, use default if empty
+	 * @param bool $implicit - use the implicit flow in the callback
 	 *
 	 * @return string
 	 */
-	public function get_wp_auth0_url( $protocol = null ) {
+	public function get_wp_auth0_url( $protocol = null, $implicit = false ) {
 		$site_url = site_url( 'index.php', $protocol );
-		return add_query_arg( 'auth0', '1', $site_url );
+		return add_query_arg( 'auth0', ( $implicit ? 'implicit' : '1' ), $site_url );
```
lbalmaceda: how are you handling this with a query param?

lib/WP_Auth0_LoginManager.php
joshcanhelp: My thought was that the error code is usually a machine name without any characters that need decoding but that's a little bit of wishful thinking ... and no harm decoding that as well.

lib/WP_Auth0_LoginManager.php
joshcanhelp: Yes. Regardless of whether we have a value or not, that `validate()` method checks against what's stored. Since `null` is not equal to whatever string we stored, this will return `false`.

lib/WP_Auth0_LoginManager.php
joshcanhelp: On a basic install with no customizations, 'openid email name nickname picture' will always be returned. If someone wants to modify the scopes that are being requested, they can use the `auth0_auth_token_scope` filter to add/remove. The `$context` is there to provide ... context for where those scopes are used. This is a pattern used in i18n in WordPress where a word might be different depending on context.  

Adjusted this to use an array for filtering ... better data structure for what's being done here.

lib/WP_Auth0_LoginManager.php
joshcanhelp: `$options->get( 'default_login_redirection' );`

lib/WP_Auth0_LoginManager.php
joshcanhelp: True but the plugin doesn't handle that so someone would have to build their own entire auth callback to handle.

lib/WP_Auth0_LoginManager.php
joshcanhelp: Implicit comes back with all the required data in an anchor link so it has to be processed by JS. The callback in that case is the WP login page which grabs the anchor data and POSTs that to the main callback URL:

https://github.com/auth0/wp-auth0/blob/dev/assets/js/implicit-login.js

We handle the rest on the backend.

lib/WP_Auth0_LoginManager.php
joshcanhelp: You know, when I create a variable, I don't want it to just do a job and go away ... I want that variable to _experience_ it's short time in memory.

lib/WP_Auth0_LoginManager.php
joshcanhelp: If you look at what `WP_Auth0_Api_Client::get_info_headers()` returns, you'll see. That's meant to built telemetry headers for API calls but the value of the array it returns can be used here. That was part of my comment about it being a good candidate for refactoring but beyond the scope of this release.

lib/WP_Auth0_LoginManager.php
joshcanhelp: The `redirect_uri` is the one that /authorize is looking for whereas `redirect_to` is where the plugin sends them after they authenticate (sent with `state` below).

lib/WP_Auth0_LoginManager.php
joshcanhelp: Since this could come from a URL param or a settings field, I want to make sure that it's an actual URL.

lib/WP_Auth0_Options.php
joshcanhelp: Determines the login flow to use:

https://github.com/auth0/wp-auth0/blob/dev/lib/WP_Auth0_LoginManager.php#L171

lib/WP_Auth0_LoginManager.php
lbalmaceda: and why you don't bother about that in the lines above?

lib/WP_Auth0_LoginManager.php
lbalmaceda: Honestly, the errors returned by the server should be used directly. No need to parse nor decode them at all (besides JSON if you need to translate them to an Object structure). Some error descriptions might include URLs to help articles. I wouldn't do any urldecoding here.

joshcanhelp: Error descriptions typically include at least spaces, if not other characters. Without urldecoding it, you'll get "This%20is%20an%20error" and such. URLs in a URL definitely need to be encoded.  

I've never seen an error code have decode-able characters, though, which was why I left it out.

lib/WP_Auth0_LoginManager.php
joshcanhelp: I see what you're saying here but I don't think we need to include functionality now that we're not sure we're going to use later. It would just be a matter of adding a parameter now or later.

lib/WP_Auth0_LoginManager.php
joshcanhelp: Can you clarify what your question is here? Do what to which lines?

lbalmaceda: Line 628 or 630 define what `$redirect_to` is going to be. Then in this line (637) you set as the `redirect_to` property (inside the state object) `filter_var( $redirect_to, FILTER_SANITIZE_URL )` as value. What's the difference and why do you need to call `filter_var` first this time?

joshcanhelp: Line 628 and 630 pull a raw value from the URL or from the database, could be anything. So if a redirect URL is not passed and there's one in the URL, get that. Otherwise if it's not passed but nothing in the URL, get the default. Then take that value, get rid of non-URL-y characters, and pack it up into `state` to be sent to Auth0 and returned. If that's null/empty in the callback then it will go to the homepage.

Hopefully that answers your question? If not, ask another way and I'll try again 😀

PR #453, opened 2018-05-01
Fix logout process

joshcanhelp:
- Remove federated logout for SSO (heavy-handed and was causing [issues](https://community.auth0.com/t/linkedin-logout-not-working-in-wordpress/10997/10) with some users) 
- Remove `end_session` method call (we're not setting any sessions so we should not be deleting any either)
- Fix telemetry header issues (fixes #448)
- Fix SLO link empty issue (`get_permalink()` is empty on non-`post` pages)
- Fix improper filter handling in `auth0_sso_footer` (cannot `echo` in filters, must return modified data)
- Add `no_sso` check for SSO footer on wp-login.php to skip SSO processing if there was a login error (can cause redirect loops)
- Minor code quality fixes (calling out internal methods

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -409,11 +409,13 @@ class WP_Auth0_LoginManager {
 					}
 				}
 
-				wp_update_user( array(
-					'ID'          => $user->data->ID,
-					'user_email'  => $userinfo->email,
-					'description' => $description,
-				) );
+				wp_update_user(
+					array(
+						'ID'          => $user->data->ID,
+						'user_email'  => $userinfo->email,
+						'description' => $description,
+					)
+				);
 			}
 
 			$this->users_repo->update_auth0_object( $user->data->ID, $userinfo );
@@ -498,8 +500,8 @@ class WP_Auth0_LoginManager {
 	 * @link https://codex.wordpress.org/Plugin_API/Action_Reference/wp_logout
 	 */
 	public function logout() {
-		$is_sso = (bool) $this->a0_options->get( 'sso' );
-		$is_slo = (bool) $this->a0_options->get( 'singlelogout' );
+		$is_sso        = (bool) $this->a0_options->get( 'sso' );
+		$is_slo        = (bool) $this->a0_options->get( 'singlelogout' );
 		$is_auto_login = (bool) $this->a0_options->get( 'auto_login' );
 
 		// Redirected here after checkSession in the footer (templates/auth0-singlelogout-handler.php).
@@ -516,18 +518,18 @@ class WP_Auth0_LoginManager {
 		// If SSO is in use, redirect to Auth0 to logout there as well.
 		if ( $is_sso ) {
 			$telemetry_headers = WP_Auth0_Api_Client::get_info_headers();
-			$redirect_url = sprintf(
+			$redirect_url      = sprintf(
 				'https://%s/v2/logout?returnTo=%s&client_id=%s&auth0Client=%s',
 				$this->a0_options->get( 'domain' ),
 				rawurlencode( home_url() ),
 				$this->a0_options->get( 'client_id' ),
-				$telemetry_headers[ 'Auth0-Client' ]
+				$telemetry_headers['Auth0-Client']
 			);
 			wp_redirect( $redirect_url );
 			exit;
 		}
 
-		// If auto-login is in use, cannot redirect back to login page
+		// If auto-login is in use, cannot redirect back to login page.
 		if ( $is_auto_login ) {
 			wp_redirect( home_url() );
 			exit;
@@ -550,7 +552,7 @@ class WP_Auth0_LoginManager {
 
 		// No need to checkSession if already logged in.
 		// URL parameter `no_sso` is set to skip checkSession.
-		if ( is_user_logged_in() || ! empty( $_GET[ 'no_sso' ] ) || ! $this->a0_options->get( 'sso' ) ) {
+		if ( is_user_logged_in() || ! empty( $_GET['no_sso'] ) || ! $this->a0_options->get( 'sso' ) ) {
 			return $previous_html;
 		}
 
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -551,8 +551,8 @@ class WP_Auth0_LoginManager {
 	public function auth0_sso_footer( $previous_html ) {
 
 		// No need to checkSession if already logged in.
-		// URL parameter `no_sso` is set to skip checkSession.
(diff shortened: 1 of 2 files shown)


lib/WP_Auth0_LoginManager.php
```diff
@@ -502,71 +504,79 @@ private function do_login( $user, $userinfo, $is_new, $id_token, $access_token )
...
-				'&auth0Client=' . base64_encode( wp_json_encode( WP_Auth0_Api_Client::get_info_headers() ) )
+		// If SSO is in use, redirect to Auth0 to logout there as well.
+		if ( $is_sso ) {
+			$telemetry_headers = WP_Auth0_Api_Client::get_info_headers();
+			$redirect_url      = sprintf(
+				'https://%s/v2/logout?returnTo=%s&client_id=%s&auth0Client=%s',
+				$this->a0_options->get( 'domain' ),
+				rawurlencode( home_url() ),
+				$this->a0_options->get( 'client_id' ),
+				$telemetry_headers['Auth0-Client']
```
lbalmaceda: I can't find any GH issues created for refactoring this usage. Are you tracking this?

lib/WP_Auth0_LoginManager.php
```diff
@@ -502,71 +504,79 @@ private function do_login( $user, $userinfo, $is_new, $id_token, $access_token )
...
+
+		// Redirected here after checkSession in the footer (templates/auth0-singlelogout-handler.php).
+		if ( $is_slo && isset( $_REQUEST['SLO'] ) ) {
+			if ( ! empty( $_REQUEST['redirect_to'] ) && filter_var( $_REQUEST['redirect_to'], FILTER_VALIDATE_URL ) ) {
+				$redirect_url = $_REQUEST['redirect_to'];
+			} else {
+				$redirect_url = home_url();
+			}
+			wp_redirect( $redirect_url );
 			exit;
```
lbalmaceda: at this line `$is_sso` and `$is_auto_login` were not used and the function is already returning. Would it make sense to define those variables right above the line that uses it? Would that help this code perform better?

lib/WP_Auth0_LoginManager.php
```diff
@@ -502,71 +504,79 @@ private function do_login( $user, $userinfo, $is_new, $id_token, $access_token )
...
+				rawurlencode( home_url() ),
+				$this->a0_options->get( 'client_id' ),
+				$telemetry_headers['Auth0-Client']
 			);
+			wp_redirect( $redirect_url );
 			exit;
 		}
 
-		if ( $auto_login ) {
+		// If auto-login is in use, cannot redirect back to login page.
```
lbalmaceda: "login page" would be the "authorize page" / HLP?

lib/WP_Auth0_LoginManager.php
```diff
@@ -502,71 +504,79 @@ private function do_login( $user, $userinfo, $is_new, $id_token, $access_token )
...
 	public function auth0_sso_footer( $previous_html ) {
 
-		// TODO: Filter handling is incorrect here!
-		echo $previous_html;
-
-		if ( is_user_logged_in() ) {
-			return;
+		// No need to checkSession if already logged in.
+		// URL parameter `no_sso` is set to skip checkSession.
+		if ( is_user_logged_in() || ! empty( $_GET['no_sso'] ) || ! $this->a0_options->get( 'sso' ) ) {
```
lbalmaceda: is like the negative (!) of the negative (empty) of the negative (no_). Should this be extracted to a variable for readability ?

lib/WP_Auth0_LoginManager.php
```diff
@@ -502,71 +504,79 @@ private function do_login( $user, $userinfo, $is_new, $id_token, $access_token )
...
-		if ( $sso ) {
-			$client_id = $lock_options->get_client_id();
-			$domain    = $lock_options->get_domain();
-			$cdn       = $this->a0_options->get( 'auth0js-cdn' );
-
-			wp_enqueue_script( 'wpa0_auth0js', $cdn );
-			include WPA0_PLUGIN_DIR . 'templates/auth0-sso-handler-lock10.php';
-		}
+		wp_enqueue_script( 'wpa0_auth0js', $this->a0_options->get( 'auth0js-cdn' ) );
+		ob_start();
```
lbalmaceda: what's this? Can't find it in this diff. Also the one `ob_get_clean()`

lib/WP_Auth0_LoginManager.php
```diff
@@ -578,8 +588,6 @@ public function auth0_singlelogout_footer() {
 
 	/**
 	 * End the PHP session.
-	 *
-	 * TODO: Remove, not necessary!
```
lbalmaceda: this method is necessary now or what?

lib/WP_Auth0_LoginManager.php
```diff
@@ -664,7 +672,7 @@ protected function die_on_login( $msg = '', $code = 0, $login_link = true ) {
 					: __( 'Please see the site administrator', 'wp-auth0' ),
 				__( 'error code', 'wp-auth0' ),
 				$code ? sanitize_text_field( $code ) : __( 'unknown', 'wp-auth0' ),
-				$login_link ? wp_login_url() : wp_logout_url(),
+				$login_link ? add_query_arg( 'no_sso', 1, wp_login_url() ) : wp_logout_url(),
```
lbalmaceda: so `add_query_arg` will return the modified URL?

lib/WP_Auth0_LoginManager.php
joshcanhelp: Yes, it's in Jira

lib/WP_Auth0_LoginManager.php
joshcanhelp: No, this is the WP login page. Auto-login forwards to /authorize with a `connection` param so we need to drop folks logging out somewhere else.

lib/WP_Auth0_LoginManager.php
joshcanhelp: I think that's less clear and `$this->a0_options->get()` is just returning an array value so I don't think optimization there is helpful.

lib/WP_Auth0_LoginManager.php
joshcanhelp: `ob_start()` is output buffer, `ob_get_clean()` returns the buffer and clears it out. This is used to "capture" an `echo` basically

lib/WP_Auth0_LoginManager.php
joshcanhelp: Should be deprecated, added that comment below 👇

lib/WP_Auth0_LoginManager.php
lbalmaceda: So you're creating a new buffer and putting in the contents of the `'templates/auth0-sso-handler-lock10.php'` file, and returning `$previous_html` plus the buffer contents?

PR #455, opened 2018-05-02
Adding back additional signup fields

joshcanhelp:
Additional signup fields were overlooked when refactoring Lock output. This PR will add them back.

### DIFF
--- a/templates/auth0-login-form.php
+++ b/templates/auth0-login-form.php
@@ -20,8 +20,17 @@ $wle = (bool) $auth0_options->get( 'wordpress_login_enabled', true );
         <?php echo $auth0_options->get( 'custom_css' ) ?>
     </style>
 
-<?php if ( $custom_js = $auth0_options->get( 'custom_js' ) ) : ?>
-    <script type="text/javascript">
-        document.addEventListener('DOMContentLoaded', function() { <?php echo $custom_js ?> });
-    </script>
-<?php endif ?>
+<?php
+$custom_js = (string) trim( $auth0_options->get( 'custom_js' ) );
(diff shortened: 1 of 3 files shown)


luisrudge:
LGTM but I can't approve

templates/auth0-login-form.php
```diff
@@ -20,8 +20,17 @@
...
-    </script>
-<?php endif ?>
\ No newline at end of file
+<?php
+$custom_js = (string) trim( $auth0_options->get( 'custom_js' ) );
+$custom_signup_fields = (string) trim( $auth0_options->get( 'custom_signup_fields' ) );
+
+if ( $custom_js || $custom_signup_fields ) {
+  printf(
+    '<script type="text/javascript">document.addEventListener("DOMContentLoaded", function() {%s %s});</script>',
```
lbalmaceda: 1. can injecting a JS file _just like that_ be considered secure?? Can someone replace this script and put some phishing stuff?
2. Previously you used `echo $custom_js` and now it's just using `$custom_js` directly. Is that OK?
3. If I got this right.. you end up having 

```
function() {
 [Content of $custom_js]
 wpAuth0LockGlobal.settings.additionalSignUpFields=[Content of $custom_signup_fields]
}
```

So the question is, because the IF is accepting both a `$custom_js OR/AND $custom_signup_fields`, if the former is not null/empty.. would it make sense to append the custom fields instruction???? because it's based on a "wpAuth0LockGlobal" variable. Are you ensuring somehow that that variable is the very same that the user is using in it's `$custom_js` content?

4. Is any of this contents (custom code snippets) being validated somehow? I'm thinking of a JSON like object parser, with a whitelist function that keeps only the known variables. Probably this is more for the (other) SDK to deal with.

PR #456, opened 2018-05-04
Adding refresh token support; adjusting default scope

### LINKED ISSUE #296: Is there a way to use Refresh Tokens and Wordpress?
Right now I'm using a users JWT Token to call my custom APIs and it's working nice.  The only issue is that when the token expires I have to kick the user out of the site and have them login again.   Which is not the best user experience.

I've looked at options to use the refresh token but it doesn't look like there is a vehicle or documentation of it for Wordpress.

Any help?
Thanks

joshcanhelp:
Adds refresh token support. Developers would add an `offline_access` key via the `auth0_auth_token_scope` filter, then store that token in the `auth0_user_login` action.

Also adds a `profile` scope to address some missing data in implicit and the /userinfo fallback. Since this PR is addressing scope and I found it while testing this, I added it here even though it's not directly related. 

Closes #296

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -232,9 +232,13 @@ class WP_Auth0_LoginManager {
 			throw new WP_Auth0_LoginFlowValidationException( $e_message, $e_code );
 		}
 
+		$access_token = $data->access_token;
+		$id_token = $data->id_token;
+		$refresh_token = isset( $data->refresh_token ) ? $data->refresh_token : null;
+
 		// Decode the incoming ID token for the Auth0 user.
 		$decoded_token = JWT::decode(
-			$data->id_token,
+			$id_token,
 			$this->a0_options->get_client_secret_as_key(),
 			array( $this->a0_options->get_client_signing_algorithm() )
 		);
@@ -251,7 +255,7 @@ class WP_Auth0_LoginManager {
 		// Management API call failed, fallback to userinfo.
 		if ( 200 !== $userinfo_resp_code || empty( $userinfo_resp_body ) ) {
 
-			$userinfo_resp      = WP_Auth0_Api_Client::get_user_info( $domain, $data->access_token );
+			$userinfo_resp      = WP_Auth0_Api_Client::get_user_info( $domain, $access_token );
 			$userinfo_resp_code = (int) wp_remote_retrieve_response_code( $userinfo_resp );
 			$userinfo_resp_body = wp_remote_retrieve_body( $userinfo_resp );
 
@@ -267,7 +271,7 @@ class WP_Auth0_LoginManager {
 
 		$userinfo = json_decode( $userinfo_resp_body );
 
-		if ( $this->login_user( $userinfo, $data->id_token, $data->access_token ) ) {
+		if ( $this->login_user( $userinfo, $id_token, $access_token, $refresh_token ) ) {
 			$state_decoded = $this->get_state( true );
 			if ( ! empty( $state_decoded->interim ) ) {
 				include WPA0_PLUGIN_DIR . 'templates/login-interim.php';
@@ -293,8 +297,6 @@ class WP_Auth0_LoginManager {
 	 * @link https://auth0.com/docs/api-auth/tutorials/implicit-grant
 	 *
 	 * @see /wp-content/plugins/auth0/assets/js/implicit-login.js
-	 *
-	 * TODO: Add a WP nonce!
 	 */
 	public function implicit_login() {
 		if ( empty( $_POST['token'] ) ) {
@@ -302,11 +304,11 @@ class WP_Auth0_LoginManager {
 		}
 
 		// Posted from the login page to the callback URL.
-		$token = sanitize_text_field( wp_unslash( $_POST['token'] ) );
+		$id_token = sanitize_text_field( wp_unslash( $_POST['token'] ) );
 
 		try {
 			$decoded_token = JWT::decode(
-				$token,
+				$id_token,
 				$this->a0_options->get_client_secret_as_key(),
 				array( $this->a0_options->get_client_signing_algorithm() )
 			);
@@ -333,7 +335,7 @@ class WP_Auth0_LoginManager {
 		// Populate legacy userinfo property.
 		$decoded_token->user_id = $decoded_token->sub;
 
-		if ( $this->login_user( $decoded_token, $token, null ) ) {
+		if ( $this->login_user( $decoded_token, $id_token ) ) {
 
 			// Validated above in $this->init_auth0().
 			$state_decoded = $this->get_state( true );
@@ -355,15 +357,16 @@ class WP_Auth0_LoginManager {
 	 * Attempts to log the user in and create a new user, if possible/needed.
 	 *
 	 * @param object $userinfo - Auth0 profile of the user.
-	 * @param string $id_token - user's ID token returned from Auth0.
-	 * @param string $access_token - user's access token returned from Auth0; not provided when using implicit_login().
+	 * @param null|string $id_token - user's ID token if returned from Auth0.
+	 * @param null|string $access_token - user's access token if returned from Auth0.
+	 * @param null|string $refresh_token - user's refresh token if returned from Auth0.
 	 *
 	 * @return bool
 	 *
 	 * @throws WP_Auth0_LoginFlowValidationException - OAuth login flow errors.
 	 * @throws WP_Auth0_BeforeLoginException - Errors encountered during the auth0_before_login action.
 	 */
-	public function login_user( $userinfo, $id_token, $access_token ) {
+	public function login_user( $userinfo, $id_token = null, $access_token = null, $refresh_token = null ) {
 		// Check that the user has a verified email, if required.
 		if ( ! $this->ignore_unverified_email && $this->a0_options->get( 'requires_verified_email' ) ) {
 			if ( empty( $userinfo->email ) ) {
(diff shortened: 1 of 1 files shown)


lib/WP_Auth0_LoginManager.php
```diff
@@ -232,9 +232,13 @@ public function redirect_login() {
 			throw new WP_Auth0_LoginFlowValidationException( $e_message, $e_code );
 		}
 
+		$access_token = $data->access_token;
+		$id_token = $data->id_token;
```
lbalmaceda: how can you be so sure that the `id_token` is present?? you might want to null check like you do for refresh tokens.

lib/WP_Auth0_LoginManager.php
```diff
@@ -232,9 +232,13 @@ public function redirect_login() {
...
+		$refresh_token = isset( $data->refresh_token ) ? $data->refresh_token : null;
+
 		// Decode the incoming ID token for the Auth0 user.
 		$decoded_token = JWT::decode(
```
lbalmaceda: does this work / would it make sense for this to be called if the token is missing?

lib/WP_Auth0_LoginManager.php
```diff
@@ -355,15 +359,16 @@ public function implicit_login() {
...
+	 * @param null|string $access_token - user's access token returned from Auth0; not provided when using implicit.
+	 * @param null|string $refresh_token - user's refresh token returned from Auth0; not provided when using implicit.
 	 *
 	 * @return bool
 	 *
 	 * @throws WP_Auth0_LoginFlowValidationException - OAuth login flow errors.
 	 * @throws WP_Auth0_BeforeLoginException - Errors encountered during the auth0_before_login action.
 	 */
-	public function login_user( $userinfo, $id_token, $access_token ) {
+	public function login_user( $userinfo, $id_token = null, $access_token = null, $refresh_token = null ) {
```
lbalmaceda: Since you're not using the values but passing them to the `do_login` function, shouldn't these defaults go in that function?

lib/WP_Auth0_LoginManager.php
```diff
@@ -355,15 +359,16 @@ public function implicit_login() {
 	 * Attempts to log the user in and create a new user, if possible/needed.
 	 *
 	 * @param object $userinfo - Auth0 profile of the user.
-	 * @param string $id_token - user's ID token returned from Auth0.
-	 * @param string $access_token - user's access token returned from Auth0; not provided when using implicit_login().
+	 * @param null|string $id_token - user's ID token returned from Auth0.
+	 * @param null|string $access_token - user's access token returned from Auth0; not provided when using implicit.
```
lbalmaceda: > not provided when using implicit.

how not?

joshcanhelp: We don't request one and don't need one. We're not pinging back to Auth0 in this case, which is the purpose of implicit, the WP server cannot connect to auth0.com for whatever reason.

lib/WP_Auth0_LoginManager.php
joshcanhelp: This is the public method, `do_login` is private. We don't need to call `login_user` with values here so I don't want to require them.

lib/WP_Auth0_LoginManager.php
lbalmaceda: only if the `response_type` includes `id_token` (i.e. `token id_token`) the `id_token` should be present. At least that's how it's supposed to work :D

joshcanhelp: Well, this being more of an app than an SDK, that's something the plugin controls and doesn't change. Both flows currently get an id_token and have no reason to change.

joshcanhelp:
> 1. Keeping in mind that refresh tokens can't be revoked, how safe is this storage?

There is no storage here, just the ability to request and the ability to process. 

> I find sort of confusing the auth0_auth_token_scope name, but I guess it can't be changed now? 

It can, it was introduced for this release ... I'm open to suggestions. 

> Can you please share a small snippet on how would a user pass this customized value?

This will be added to the plugin [extension docs](https://auth0.com/docs/cms/wordpress/extending) after release but, in short:

```php
function auth0_theme_hook_auth0_auth_token_scope( $scope ) {
	$scope[] = 'offline_access';
	return $scope;
}
add_filter( 'auth0_auth_token_scope', 'auth0_theme_hook_auth0_auth_token_scope' );
```

☝️ that's how one would add the `refresh_token` request.

```
function auth0_theme_hook_auth0_user_login( $user_id, $userinfo, $is_new, $id_token, $access_token, $refresh_token ) {
	echo $refresh_token ? $refresh_token : 'not provided';
        die();
}
add_action( 'auth0_user_login', 'auth0_theme_hook_auth0_user_login', 10, 6 );
```

☝️ that's how you would do something with the refresh token. That action runs right after a user is authenticated in WP.

lib/WP_Auth0_LoginManager.php
lbalmaceda: What do you mean by _both_? decode function is only using id_token.

joshcanhelp: Sorry ... both `access_token` and `id_token`. They're both returned at the same time so if we have an `access_token`, we also have an `id_token`.

lib/WP_Auth0_LoginManager.php
lbalmaceda: Reading this again after yesterday's discussion, this method handles code exchange redirection so here we have access_token and id_token. No need to null-check them 👍

lib/WP_Auth0_LoginManager.php
```diff
@@ -333,7 +337,7 @@ public function implicit_login() {
 		// Populate legacy userinfo property.
 		$decoded_token->user_id = $decoded_token->sub;
 
-		if ( $this->login_user( $decoded_token, $token, null ) ) {
+		if ( $this->login_user( $decoded_token, $token ) ) {
```
lbalmaceda: Can the `$_POST['token']` "token" param be renamed before next release? And please rename `$token` to `$id_token` or similar, it's easier to follow up which token is it. 

https://github.com/auth0/wp-auth0/blob/6382a6ca84c1be0ff64a80a4dbfbfc5efeaead39/lib/WP_Auth0_LoginManager.php#L304

lib/WP_Auth0_LoginManager.php
lbalmaceda: I get it that implicit doesn't receive an access_token since you're not requesting it. Fine. But This method doesn't know if the user called implicit flow or code flow. So, the description of the method should just say that null values are accepted for access_token and refresh_token, just that. And if you want, in the method that handles the implicit result parsing you can add in the docblock that you'll never receive an access_token or refresh_token there.

lib/WP_Auth0_LoginManager.php
joshcanhelp: The key in `$_POST` cannot, unfortunately. That's the name it's had for a while so changing would be breaking. Just changed the internal var name, good call.

lib/WP_Auth0_LoginManager.php
joshcanhelp: I confirmed that OIDC and non-OIDC applications both prove `access_token` and `id_token` with `response_type=code`

lib/WP_Auth0_LoginManager.php
```diff
@@ -462,10 +467,11 @@ public function login_user( $userinfo, $id_token, $access_token ) {
 	 * @param bool   $is_new - `true` if the user was created in the WordPress database, `false` if not.
 	 * @param string $id_token - user's ID token returned from Auth0.
 	 * @param string $access_token - user's access token returned from Auth0; not provided when using implicit_login().
+	 * @param string $refresh_token - user's refresh token returned from Auth0; not provided when using implicit_login().
 	 *
 	 * @throws WP_Auth0_BeforeLoginException - Errors encountered during the auth0_before_login action.
 	 */
-	private function do_login( $user, $userinfo, $is_new, $id_token, $access_token ) {
+	private function do_login( $user, $userinfo, $is_new, $id_token, $access_token, $refresh_token ) {
```
lbalmaceda: what's the difference between `login_user` and `do_login` methods?

lib/WP_Auth0_LoginManager.php
```diff
@@ -462,10 +467,11 @@ public function login_user( $userinfo, $id_token, $access_token ) {
 	 * @param bool   $is_new - `true` if the user was created in the WordPress database, `false` if not.
 	 * @param string $id_token - user's ID token returned from Auth0.
 	 * @param string $access_token - user's access token returned from Auth0; not provided when using implicit_login().
```
lbalmaceda: same docblock comment

lbalmaceda:
About `auth0_auth_token_scope` I'd just call it `auth0_auth_scope` or `auth0_authentication_scope`. The the scope is added to the tokens (plural), but in all the platforms I recall I saw that as authentication scope or just "scope" param that you called `setScope` or `addAuthenticationParams({dic with scope param})`

lib/WP_Auth0_LoginManager.php
joshcanhelp: `login_user` is called after auth is successful so it decides if someone can login or if we need to make an account. `do_login` is called if `login_user` is successful, handles core WP login and hooks.

lbalmaceda: can these get renamed to something more clear? 

i.e.:

 `login_user` is processing an authentication and deciding what to do next. Shouldn't that be something like `process_successful_auth`?

and `do_login` well, it's more straight forward. But it could be `do_wp_login`

PR #457, opened 2018-05-04
Fix empty path notice on initial setup

joshcanhelp:


lib/WP_Auth0_Options.php
```diff
@@ -86,10 +86,14 @@ public function convert_client_secret_to_key( $secret, $is_encoded, $is_RS256, $
 	 */
 	public function get_web_origins() {
 		$home_url_parsed = parse_url( home_url() );
-		$home_url_origin = str_replace( $home_url_parsed[ 'path' ], '', home_url() );
+		$home_url_origin = ! empty( $home_url_parsed[ 'path' ] )
+			? str_replace( $home_url_parsed[ 'path' ], '', home_url() )
```
lbalmaceda: `str_replace` with `''`  as the index, would it work as an append? and if so, why not use `.` instead? also if true, if `path` is empty appending something without any characters in the middle would mean having that later string as the final output. 
e.g.: 

- `path=''`
- `separator=''`
- home_url='something'

if !empty(path) -> home_url_origin=str_replace('', '', 'something') -> home_url_origin='something'

or, using php append



home_url_origin='' . 'something' -> home_url_origin='something'

Note the line above uses `$home_url_parsed[ 'path' ]` directly without checking if it's empty first, just as the previous code was doing. Am I right, is that the same?

joshcanhelp: I'm not sure if I understand ... I'm not appending anything here, I'm removing the `path` from the base URL of the site. So if the base URL is:

`http://bananafarmer.com/peels`

... I want:

`http://bananafarmer.com`

Using `str_replace()`, the first param is the needle, second is the replacement, third is the haystack. 

That said, I probably want to do this instead:

```
$home_url_parsed = parse_url( home_url() );
$home_url_origin = $home_url_parsed['scheme'] . '://' . $home_url_parsed['host'] . 
	( ! empty( $home_url_parsed['port'] ) ? ':' . $home_url_parsed['port'] : '' );
```

Then I can put that in a function and not repeat it.

PR #460, opened 2018-05-07
Hide the signup tab if registrations are turned off

joshcanhelp:


lib/WP_Auth0_Lock10_Options.php
```diff
@@ -193,6 +193,10 @@ public function get_lock_options() {
 
     $options_obj = array_replace_recursive( $extraOptions, $options_obj, $extended_settings );
 
+    if ( ! $this->wp_options->is_wp_registration_enabled() && ! isset( $options_obj["allowSignUp"] )) {
+      $options_obj["allowSignUp"] = false;
```
lbalmaceda: what if `is_wp_registration_enabled` is false so that the first condition passes; but then the `allowSignup` is true. You would keep `allowSignUp=true` instead of turning it off (since `is_wp_registration_enabled` is false)

lib/WP_Auth0_Options.php
```diff
@@ -13,10 +13,7 @@ public static function Instance() {
 	}
 
 	public function is_wp_registration_enabled() {
-		if ( is_multisite() ) {
-			return users_can_register_signup_filter();
-		}
-		return get_site_option( 'users_can_register', 0 ) == 1;
+		return is_multisite() ? users_can_register_signup_filter() : get_site_option( 'users_can_register' );
```
lbalmaceda: would that second value be the same to `get_site_option( 'users_can_register', 0 ) == 1` (previous)? Is this tested somewhere?

joshcanhelp: Common practice. `get_site_option( 'users_can_register', 0 ) == 1` means "get the option `users_can_register` and if it doesn't exist use `0` and then loosely compare that to `1`." 

If `users_can_register` is on, it will be a `"1"` (true), if it's off it will be a `"0"` (false), and if it's not set, `get_site_options()` will return an explicit `false`. [Codex](https://codex.wordpress.org/Function_Reference/get_site_option#Return_Values)

lbalmaceda: thanks for that link ⚡️

lib/WP_Auth0_Lock10_Options.php
joshcanhelp: Using the main pathway, yes, but some of that is hook-able so it can be overridden. These extra settings JSON fields say they will override all other settings and I want them to be able to do that, even if it's edge case

PR #465, opened 2018-05-22
Fixed auto-loader to skip non-WP-Auth0 classes

### LINKED ISSUE #461: Autoloader performance issue
It appears that the autoloader slows down the wordpress site significantly.

Running performance monitor I was able to see that for every class that was being searched for, it would search in a bunch of different directories - even for files that were not part of the auth0 plugin. The more plugins I had installed, the worse the problem would get.

image

I believe the issue exists here:

https://github.com/auth0/wp-auth0/blob/9713e6a7cdf4c3446632be5338426dfba102bf3a/WP_Auth0.php#L412

And this is the fix that I did that seems to help things although I don't know if it is a permanent solution:

```
foreach ( $paths as $p ) {
    foreach ( $exts as $ext ) {
        if (strpos($p.$class.$ext, "WP_Auth0") !== FALSE && file_exists( $p.$class.$ext ) ) {
            require_once $p.$class.$ext;
            return true;
        }
    }
}
```

Thanks.

joshcanhelp:
Fixes #461

### DIFF
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -436,27 +436,50 @@ class WP_Auth0 {
 		delete_transient( WPA0_JWKS_CACHE_TRANSIENT_NAME );
 	}
 
+	/**
+	 * Look for a class within a specific set of paths.
+	 *
+	 * @param string $class - Class name to look for.
+	 *
+	 * @return bool
+	 */
 	private function autoloader( $class ) {
-		$path = WPA0_PLUGIN_DIR;
-		$paths = array();
-		$exts = array( '.php', '.class.php' );
+		$source_dir = WPA0_PLUGIN_DIR . 'lib/';
 
-		$paths[] = $path.'lib/';
-		$paths[] = $path.'lib/admin/';
-		$paths[] = $path.'lib/exceptions/';
-		$paths[] = $path.'lib/wizard/';
-		$paths[] = $path.'lib/initial-setup/';
-		$paths[] = $path.'lib/twitter-api-php/';
-		$paths[] = $path.'lib/php-jwt/Exceptions/';
-		$paths[] = $path.'lib/php-jwt/Authentication/';
-		$paths[] = $path;
+		// Catch non-name-spaced classes that still need auto-loading.
+		switch ( $class ) {
+			case 'TwitterAPIExchange':
+				require_once $source_dir . 'twitter-api-php/' . $class . '.php';
+				return true;
 
-		foreach ( $paths as $p ) {
-			foreach ( $exts as $ext ) {
-				if ( file_exists( $p.$class.$ext ) ) {
-					require_once $p.$class.$ext;
-					return true;
-				}
+			case 'JWT':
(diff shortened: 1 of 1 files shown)


WP_Auth0.php
```diff
@@ -436,27 +436,50 @@ public static function uninstall() {
...
+			case 'BeforeValidException':
+			case 'ExpiredException':
+			case 'SignatureInvalidException':
+				require_once $source_dir . 'php-jwt/Exceptions/'  . $class . '.php';
+				return true;
+		}
+
+		// Anything that's not part of the above and not name-spaced can be skipped.
+		if ( 0 !== strpos( $class, 'WP_Auth0' ) ) {
+			return false;
```
lbalmaceda: should you log something on this line?

WP_Auth0.php
```diff
@@ -436,27 +436,50 @@ public static function uninstall() {
...
+		}
+
+		$paths = array(
+			$source_dir,
+			$source_dir . 'admin/',
+			$source_dir . 'exceptions/',
+			$source_dir . 'wizard/',
+			$source_dir . 'initial-setup/',
```
lbalmaceda: I guess this last comma will be ignored.

WP_Auth0.php
```diff
@@ -436,27 +436,50 @@ public static function uninstall() {
...
+		);
+
+		foreach ( $paths as $path ) {
```
lbalmaceda: I don't know how often this method would be called, but I suppose several times per execution. Is there any way of iterating through all the files in a more efficient way? Maybe accepting an array of "classes to require" so you can iterate once and require them all.

WP_Auth0.php
```diff
@@ -436,27 +436,50 @@ public static function uninstall() {
...
+			if ( file_exists( $path . $class . '.php' ) ) {
+				require_once $path . $class . '.php';
+				return true;
 			}
```
lbalmaceda: no need to `return false`?

WP_Auth0.php
joshcanhelp: Definitely not. This is called for all class instantiations. It's not an error if an non-WP-Auth0 class is being called. We want to escape quickly.

WP_Auth0.php
joshcanhelp: Yes. It's a PHP standard to add a comma on the last line of an array

WP_Auth0.php
joshcanhelp: > several times per execution

Correct

> Maybe accepting an array of "classes to require"

That would defeat the purpose of the autoloader. This is here to "magically" include the necessary files for whatever class you might call. There is a more efficient way to do this using [Namespaces](https://secure.php.net/manual/en/language.namespaces.php) but that would require a lot more work (I have it on my list)

WP_Auth0.php
joshcanhelp: Nope!

PR #469, opened 2018-05-25
Save  or  if not provided; remove extemporaneous ID token sttributes

joshcanhelp:
- Make sure we have both a `user_id` and `sub` saved for both login flows
- Unset extra ID token attributes not required for profile data

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -260,6 +260,11 @@ class WP_Auth0_LoginManager {
 
 		$userinfo = json_decode( $userinfo_resp_body );
 
+		// Populate sub property, if not provided.
+		if ( ! isset( $userinfo->sub ) ) {
+			$userinfo->sub = $userinfo->user_id;
+		}
+
 		if ( $this->login_user( $userinfo, $id_token, $access_token, $refresh_token ) ) {
 			$state_decoded = $this->get_state();
 			if ( ! empty( $state_decoded->interim ) ) {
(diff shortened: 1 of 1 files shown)


lib/WP_Auth0_LoginManager.php
```diff
@@ -321,8 +326,15 @@ public function implicit_login() {
...
-		// Populate legacy userinfo property.
-		$decoded_token->user_id = $decoded_token->sub;
+		// Remove unneeded ID token attributes.
+		foreach ( array( 'iss', 'aud', 'iat', 'exp', 'nonce', 'clientID' ) as $attr ) {
+			unset( $decoded_token->$attr );
+		}
+
+		// Populate legacy user_id property, if not provided.
+		if ( ! isset( $decoded_token->user_id ) ) {
+			$decoded_token->user_id = $decoded_token->sub;
```
lbalmaceda: why? if you ensure you always have a `sub` you can make your code rely on that.

lib/WP_Auth0_LoginManager.php
```diff
@@ -321,8 +326,15 @@ public function implicit_login() {
 			);
 		}
 
-		// Populate legacy userinfo property.
-		$decoded_token->user_id = $decoded_token->sub;
+		// Remove unneeded ID token attributes.
+		foreach ( array( 'iss', 'aud', 'iat', 'exp', 'nonce', 'clientID' ) as $attr ) {
```
lbalmaceda: what's the reason to remove them? how will you know if a token works for a given audience or if the token is still valid (without making a network request)?

joshcanhelp: This is just saving userinfo, we don't use it as a token. I'm trying to get the information that we're saving to be more consistent between the two login flows. Also, this isn't saved as `id_token` so it's a bit confusing to use it as such.

lib/WP_Auth0_LoginManager.php
joshcanhelp: Again, standardizing between the two. Implicit has both and "make your code rely on that" is a little more involved than it might seem.

PR #479, opened 2018-06-06
Fix SLO callback URL 

joshcanhelp:
Reported here - https://wordpress.org/support/topic/sso-logout-issues/

SLO is using `home_url()`, which some sites don't have in the "Allowed Callback URLs" field for the Application. Changed to the main Auth0 callback URL which is always there. Callback is not used as a redirect so does not need to change for Implicit, just needs to be valid. 

Also adds SLO check to the admin + login pages and changes the redirect URL used after logout to point to an external page if there is one or the wp-login.php page.

### DIFF
--- a/templates/auth0-singlelogout-handler.php
+++ b/templates/auth0-singlelogout-handler.php
@@ -23,7 +23,10 @@ $logout_url = add_query_arg( 'SLO', 1, $logout_url );
       domain:'<?php echo sanitize_text_field( $this->a0_options->get( 'domain' ) ); ?>'
     });
 
-    var sloOptions = { 'responseType' : 'token id_token', 'redirectUri' : '<?php echo home_url() ?>' };
+    var sloOptions = {
+        'responseType' : 'token id_token',
+        'redirectUri' : '<?php echo $this->a0_options->get_wp_auth0_url( null ) ?>'
+    };
     webAuth.checkSession( sloOptions, function ( err ) {
             if ( err && err.error && 'login_required' === err.error ) {
                 window.location = '<?php echo $logout_url ?>';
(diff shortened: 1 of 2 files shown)


templates/auth0-singlelogout-handler.php
```diff
@@ -6,7 +6,7 @@
 
 $logout_url = wp_logout_url();
 $logout_url = html_entity_decode( $logout_url );
-$logout_url = add_query_arg( 'redirect_to', get_permalink(), $logout_url );
+$logout_url = add_query_arg( 'redirect_to', ( get_the_ID() ? get_permalink() : wp_login_url() ), $logout_url );
```
lbalmaceda: please explain the logic behind this line

templates/auth0-singlelogout-handler.php
```diff
@@ -23,7 +23,10 @@
       domain:'<?php echo sanitize_text_field( $this->a0_options->get( 'domain' ) ); ?>'
     });
 
-    var sloOptions = { 'responseType' : 'token id_token', 'redirectUri' : '<?php echo home_url() ?>' };
```
lbalmaceda: can this be considered a breaking change?

joshcanhelp: No. `checkSession` just needs a valid callback, it doesn't actually send the browser there.

templates/auth0-singlelogout-handler.php
joshcanhelp: `get_the_ID()` checks if we're on a page/post, which means `get_permalink()` will return something useful. If not, we send them to the login page where Lock will appear.

joshcanhelp:
@lbalmaceda - `home_url()` always returns something but what it returns is not always something that `/authorize` will accept (i.e. it's not always set in the "Allowed Callback URLs"). We have a URL we can use that's always there (saved when the App is created) so it's better to use that.

lbalmaceda:
Alright. I'm not so convinced though. Like I said, this looks like something we should "help fix" from the documentation side. A good doc on "how to achieve SLO" would mention this url needs to be set as a requirement, here and whitelisted in the tenant. That's what most of the SDKs do AFAIK. But again, I'm not going to block you for this. Your call :)

PR #486, opened 2018-06-14
Fix import-export tabs not working

### LINKED ISSUE #484: TypeError: jQuery(...).tab is not a function
`TypeError: jQuery(...).tab is not a function admin.php:689:5`

When in `/wp-admin` > `Auth0` > `Import-Export settings` clicking on the Export Settings tab causes this JS error and the page does not appear.

Running wp-auth0 v3.6.1

joshcanhelp:
Fixes #484

### DIFF
--- a/assets/js/admin.js
+++ b/assets/js/admin.js
@@ -58,6 +58,19 @@ jQuery(document).ready(function($) {
     });
 
     /*
+    Import and Export settings tabs
+     */
+    $('.js-a0-upload-toggle').click(function(){
+        $('#js-a0-upload-file').toggle();
+        $('#js-a0-paste-json').toggle();
+    });
+
+    $('.js-a0-import-export-tabs').click(function (e) {
+        e.preventDefault();
+        $(this).tab('show');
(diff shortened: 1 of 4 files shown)


assets/js/admin.js
```diff
@@ -57,6 +57,19 @@ jQuery(document).ready(function($) {
...
+    /*
+    Import and Export settings tabs
+     */
+    $('.js-a0-upload-toggle').click(function(){
+        $('#js-a0-upload-file').toggle();
+        $('#js-a0-paste-json').toggle();
+    });
+
+    $('.js-a0-import-export-tabs').click(function (e) {
+        e.preventDefault();
```
lbalmaceda: what's this line doing and why does this function returns an error while the other above doesn't?

joshcanhelp: The HTML is an `<a>` tag so we don't want it to act like a link. 

This function does not return an error ... it doesn't return anything. `e` is for `event`, so stop the default event.

PR #490, opened 2018-06-20
Show signup tab if action=register on wp-login.php page

### LINKED ISSUE #480: Optionally load client ID, secret and domain from environment
We have multiple environments that we keep in sync with database backups cloned between them. Obviously this confused the Auth0 plugin as it doesn't have the right credentials any more. 

Loading the credentials from the environment in a 12-factor app style would make this a lot easier. Perhaps if the values are blank in the database they could be loaded from the env? Another issue here is the user migration token, which could perhaps be generated in a more stable fashion so it doesn't need to be stored in the database?

### LINKED ISSUE #489: Question: Way to visit directly to Sign Up tab?
Is there a way to directly navigate to `/wp-login.php` and have the Sign Up tab be displaying instead of the login form?

Would be great if you could pass a query parameter like `/wp-login.php?initialscreen=signup` or `/wp-login.php?action=singup`

I've looked through the documentation and have found that in Lock for Web this is an option, but im not sure if its implemented in `wp-auth0` 
https://auth0.com/docs/libraries/lock/v11/configuration#initialscreen-string-

joshcanhelp:
Closes #489

### DIFF
--- a/lib/WP_Auth0_Lock10_Options.php
+++ b/lib/WP_Auth0_Lock10_Options.php
@@ -204,6 +204,10 @@ class WP_Auth0_Lock10_Options {
       $options_obj['disableSignupAction'] = true;
     }
 
+    if ( function_exists( 'login_header' ) && isset( $_GET['action'] ) && 'register' === $_GET['action'] ) {
+      $options_obj['initialScreen'] = 'signUp';
+    }
+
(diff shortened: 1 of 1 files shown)


talamihg:
That's great! Thanks so much.

PR #491, opened 2018-06-20
Fix missing signup fields

### LINKED ISSUE #487: Custom Fields
It doesn't seem that the Custom Signup Fields option is working. I've included some JSON, but its not changing Lock to include those fields.
```
[{
name: "first_name",
placeholder: "First Name"
}]
```

I am assuming that you want the square brackets around the entire array like shown here: https://auth0.com/docs/libraries/lock/v11/configuration#text-fields ?

Can confirm that the JSON block above is producing this on /wp-login.php 
```
<script type="text/javascript">document.addEventListener("DOMContentLoaded", function() { wpAuth0LockGlobal.settings.additionalSignUpFields=[{
name: "first_name",
placeholder: "First Name"
}]});</script>
```

joshcanhelp:
Fixes #487

### DIFF
--- a/assets/js/lock-init.js
+++ b/assets/js/lock-init.js
@@ -1,4 +1,4 @@
-/* globals jQuery, console, Cookies, wpAuth0LockGlobal, Auth0Lock, Auth0LockPasswordless */
+/* globals jQuery, console, Cookies, wpAuth0LockGlobal, wpAuth0LockGlobalFields, Auth0Lock, Auth0LockPasswordless */
 jQuery(document).ready(function ($) {
     var opts = wpAuth0LockGlobal;
     var loginForm = $( '#' + opts.loginFormId );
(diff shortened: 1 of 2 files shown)


templates/auth0-login-form.php
```diff
@@ -25,12 +25,13 @@
 $custom_signup_fields = (string) trim( $auth0_options->get( 'custom_signup_fields' ) );
 
 if ( $custom_js || $custom_signup_fields ) {
```
lbalmaceda: This line can be deleted. Both cases are accepted below.

assets/js/lock-init.js
```diff
@@ -25,6 +25,11 @@ jQuery(document).ready(function ($) {
         Cookies.set( opts.nonceCookieName, opts.settings.auth.params.nonce );
     }
 
+    // Look for additional fields to display
+    if ( wpAuth0LockGlobalFields ) {
+        opts.settings.additionalSignUpFields = wpAuth0LockGlobalFields;
```
lbalmaceda: Q: How does this setup work?

templates/auth0-login-form.php
joshcanhelp: I don't want the empty script tags though

assets/js/lock-init.js
joshcanhelp: The global `wpAuth0LockGlobalFields` variable is set earlier in execution (the other file you're reviewing) and this looks to see whether there is anything there and assigns if so. Previously, the block that set the fields JS was trying to assign to the `wpAuth0LockGlobal` and causing problems in some setups (the global not being available). JS globals are not the greatest but they're commonly used in WP to pass information from PHP or the DB into JS. These fields are stored as JS, not JSON, so I can't parse, store, then output like I would like to.

PR #495, opened 2018-06-28
Add a filter for nonce and state cookie names

### LINKED ISSUE #494: `auth0_state` cookie and Pantheon
We have a client who is on Pantheon and we are having issues with the latest version of the plugin and the "Invalid state" error. Looking at this troubleshooting guide I have a pretty good idea what the issue is (https://github.com/joshcanhelp/troubleshooting-invalid-state-wp/) based on step three of the plugin's new process.
The problem is that Pantheon Wordpress cookies have to have `STYXKEY_` at the beginning of their name or they are unreadable. As such, I believe that there is a change the `auth0_state` cookie is not being seen by our site. 
Is there a way for us to override the cookie name to `STYXKEY_auth0_state` perhaps or maybe a new setting field can be added where the cookie name is set?

joshcanhelp:
- Change the `protected` `get_storage_cookie_name()` to public + static
- Add a `auth0_state_cookie_name` and `auth0_none_cookie_name` filter to the two affected methods
- Change direct calling of the cookie name constants to the static method

**Sample usage:**

```php
/**
 * Prefix state and nonce cookie names. 
 * 
 * @param string $cookie_name - Cookie name to modify.
 * 
 * @return string
 */
function auth0_theme_hook_prefix_cookie_name( $cookie_name ) {
	return 'STYXKEY_' . $cookie_name;
}
 add_filter( 'auth0_state_cookie_name', 'auth0_theme_hook_prefix_cookie_name' );
 add_filter( 'auth0_nonce_cookie_name', 'auth0_theme_hook_prefix_cookie_name' );
```

Passes all functional tests. 

Closes #494


lib/WP_Auth0_Nonce_Handler.php
```diff
@@ -61,9 +61,10 @@ private function __construct() {
 	 * Start-up process to make sure we have something stored.
 	 */
 	protected function init() {
-		if ( defined( static::NONCE_COOKIE_NAME ) && isset( $_COOKIE[ static::NONCE_COOKIE_NAME ] ) ) {
+		// If a NONCE_COOKIE_NAME is not defined then we don't need to persist the nonce value.
+		if ( defined( static::NONCE_COOKIE_NAME ) && isset( $_COOKIE[ self::get_storage_cookie_name() ] ) ) {
 			// Have a cookie, don't want to generate a new one.
-			$this->unique = $_COOKIE[ static::NONCE_COOKIE_NAME ];
+			$this->unique = $_COOKIE[ self::get_storage_cookie_name()  ];
```
lbalmaceda: what about `self::get_cookie_prefix() + static::NONCE_COOKIE_NAME`? Same for the remaining calls

joshcanhelp: No need to exert that kind of control, name is irrelevant.

joshcanhelp:
@lbalmaceda - I'd rather the developer have full control over the name.

joshcanhelp:
@lbalmaceda - Added a sample usage to prefix all cookie names.

lbalmaceda:
Ah it's easier than I thought. OK!

galengidman:
Thank you, thank you, thank you.

PR #498, opened 2018-07-02
Add code quality tools and contrib instructions

joshcanhelp:


README.md
```diff
@@ -3,7 +3,7 @@
 Login by Auth0
 ====
 
-Single Sign On for Enterprises + Social Login + User/Passwords. For all your WordPress instances. Powered by Auth0.
+Single Sign On for Enterprises + Social Login + User/Passwords + Passwordless. For all your WordPress instances. Powered by Auth0.
```
cocojoe: Might be nicer as a bullet point list vs +

README.md
```diff
@@ -53,20 +56,22 @@ git clone https://github.com/joshcanhelp/auth0-wp-test.git ./wp-content/themes/a
 # Optional, clones the testing theme to assist with development
 ```
```
cocojoe: I feel this should be in the repo as an additional file, it seems strange for us to link to your own personal gist, appreciate there probably wasn't an accessible Auth0 one.

cocojoe:
will assume all your rules work 😄

README.md
lbalmaceda: I agree since this is getting bigger, but he's already using a bulleted list below this paragraph.

README.md
lbalmaceda: I don't mind as long as the content is kept up to date.

README.md
joshcanhelp: How about:

"SSO, Social Login, Username & Passwords,  Passwordless, and more."

README.md
joshcanhelp: I totally agree. I was a little apprehensive to do it on my own one. It a lot of code to review to put out there, TBH. Would be nice as part of the plugin, though, as you could symlink from the `themes` folder (or copy) and use it for testing. I can add in another PR?

joshcanhelp:
@cocojoe - They work but the plugin has a long way to go before passing all those sniffs. On PR at a time!

README.md
cocojoe: Is there more ? ;)

PR #502, opened 2018-07-18
Remove connection settings

joshcanhelp:


lib/admin/WP_Auth0_Admin.php
```diff
@@ -6,193 +6,6 @@ class WP_Auth0_Admin {
 
 	protected $router;
 
-	protected $providers = array(
```
joshcanhelp: We no longer control connections from the settings page so this is not needed.

lib/initial-setup/WP_Auth0_InitialSetup_Connections.php
```diff
@@ -3,239 +3,18 @@
 class WP_Auth0_InitialSetup_Connections {
 
 	protected $a0_options;
-	protected $providers = array(
```
joshcanhelp: We no longer control connections from the settings page so this is not needed. 

The `options` child array is still used in a method for this class - `WP_Auth0_InitialSetup_Connections::toggle_social()` - so those 3 were moved down. Method will be deprecated in another PR but need to retain functionality in the meantime.

lib/initial-setup/WP_Auth0_InitialSetup_Connections.php
```diff
@@ -3,239 +3,18 @@
...
-			'name'     => 'exact',
-			'icon'     => 'Exact',
-		),
-
-	);
 
 	public function __construct( WP_Auth0_Options $a0_options ) {
 		$this->a0_options = $a0_options;
-
-		add_action( 'wp_ajax_a0_initial_setup_set_connection', array( $this, 'update_connection' ) );
```
joshcanhelp: Unused AJAX hook

lib/initial-setup/WP_Auth0_InitialSetup_Connections.php
```diff
@@ -3,239 +3,18 @@
...
 	}
 
 	public function render( $step ) {
-		wp_enqueue_script( 'wpa0_async', WPA0_PLUGIN_LIB_URL . 'async.min.js' );
```
joshcanhelp: Connection pieces were removed, other vars were moved to the `templates/initial-setup/connections.php` template below

templates/settings.php
```diff
@@ -108,56 +108,6 @@
 </div>
 
 <script type="text/javascript">
-
```
joshcanhelp: This all updates connections based on settings, which is not there anymore.

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -138,10 +138,6 @@ public function consent_callback( $name ) {
 						$enabled_clients = array_diff( $connection->enabled_clients, array( $client_id ) );
 						WP_Auth0_Api_Client::update_connection( $domain, $app_token, $connection->id, array( 'enabled_clients' => array_values( $enabled_clients ) ) );
 					}
-				} elseif ( $connection->strategy !== 'auth0' ) {
```
joshcanhelp: Connection will only be DB

templates/settings.php
```diff
@@ -19,13 +19,18 @@
...
 		<ul class="nav nav-tabs" role="tablist">
-			<?php foreach ( array( 'basic', 'features', 'appearance', 'advanced', 'help' ) as $tab ) : ?>
-			  <li role="presentation"><a id="tab-<?php echo $tab; ?>" href="#<?php echo $tab; ?>" aria-controls="
-															<?php
-															echo $tab
-															?>
-				  " role="tab" data-toggle="tab" class="js-a0-settings-tabs"><?php echo ucfirst( $tab ); ?></a></li>
-			<?php endforeach; ?>
+			<?php
+			foreach ( array( 'basic', 'features', 'appearance', 'advanced', 'help' ) as $tab ) {
```
joshcanhelp: Refactored, whitespace scan was causing tab memory to fail

PR #507, opened 2018-07-24
Add translation ability

joshcanhelp:
- Add translation to strings throughout the plugin
- Add text-domain loading
- Add main POT file and English translation files (see below)

**Note for reviewer(s):** The PO and POT files (~95% of added lines in this PR) are all generated from translatable strings in the plugin. Much of this language came from the settings page, which was recently re-written and reviewed. Reviewing these is probably superfluous.


templates/wp-login-form.php
```diff
@@ -1,23 +0,0 @@
-<form method="post" class="wp-form-signin" action="">
```
joshcanhelp: Template not used

PR #509, opened 2018-07-31
Add constant settings support

### LINKED ISSUE #480: Optionally load client ID, secret and domain from environment

joshcanhelp:
Add support for defining settings in a constant. Closes #480.

This will allow for sensitive data like the client secret, API token, and migration token to be stored in constants. The options class now looks for constants of a specific name being defined, like so:

```
// Storing the client secret in a constant.
// Add to wp-config.php or anywhere before WP_Auth0_Options_Generic is instantiated.
// Constant name is "AUTH0_ENV_" + option key in caps.
define( 'AUTH0_ENV_CLIENT_SECRET', 'YOUR_CLIENT_SECRET_HERE' );
```

The default constant name should be `AUTH0_ENV_` followed by the option name to override in all caps (the prefix can be modified with the `wp_auth0_settings_constant_prefix` filter). Option keys listed [here](https://github.com/auth0/wp-auth0/blob/master/lib/WP_Auth0_Options.php) can all be overridden.

**Note:** the `migration_token` value is generated by the plugin when user migration is turned on. If there is already a value in the admin, make sure to set the constant to the same value. If that value needs to change, it also must be changed in the user migration script for the database connection being used. 

Once that is set, `WP_Auth0_Options_Generic::get_options()` will look for values before loading all options. `WP_Auth0_Options_Generic::get()` will pull from this adjusted array. 

The settings field will change display based on this new value and show the constant being used:

screenshot 2018-06-20 08 51 02

Saving the settings page after setting a constant value will validate the constant-set values and delete them from the options array being saved to the database (important to mention that in upcoming docs).


phpcs-test-ruleset.xml
```diff
@@ -7,7 +7,6 @@
```
joshcanhelp: Code standards ruleset

lib/WP_Auth0_Options_Generic.php
```diff
@@ -1,69 +1,221 @@
...
+	/**
+	 * Get the value of an overriding constant if one is set, return null if not.
+	 *
+	 * @param string $key - Option key to look for.
+	 *
+	 * @return string|null
+	 */
+	public function get_constant_val( $key ) {
+		$constant_name = $this->get_constant_name( $key );
+		return defined( $constant_name ) ? constant( $constant_name ) : null;
```
lbalmaceda: it sounds like this method is repeating the logic on the `has_constant_val` method. Do you think this can be improved? i.e. if tomorrow the logic to check whether a constant exists or not changes, you'll need to change 2 methods instead of 1.

lib/WP_Auth0_Options_Generic.php
```diff
@@ -1,69 +1,221 @@
...
+		}
+
 		$options[ $key ] = $value;
-		$this->_opt      = $options;
+		$this->_opts     = $options;
 
-		if ( $should_update ) {
-			update_option( $this->_options_name, $options );
+		// No database update so process completed successfully.
+		if ( ! $should_update ) {
```
lbalmaceda: is this update something you'll normally run after setting each key-value? Think of the case of setting a bunch of variables. I guess only the last call to this method would have this value as `true`. Right? Since I guess is not going to perform well doing so after each call. If so, would it make sense to remove this parameter from this method and just expose the `update_all` method for the user to call it directly on demand? You'll need to weight the most common use cases to make this decision.

lib/admin/WP_Auth0_Admin_Generic.php
```diff
@@ -213,6 +234,21 @@ protected function render_field_description( $text ) {
 		printf( '<div class="subelement"><span class="description">%s.</span></div>', $text );
 	}
 
+	/**
+	 * Check if the setting is provided by a constant and indicate.
+	 *
+	 * @param string $input_name - Input name for the field, used as option key.
+	 */
+	protected function render_const_notice( $input_name ) {
+		if ( $this->options->has_constant_val( $input_name ) ) {
```
lbalmaceda: this check is already done on some of the `render_const_notice` calls. Decide whether you'd like to do this here or outside, but don't check twice... it's an overkill. e.g. call


```php
protected function render_radio_buttons( array $buttons, $id, $input_name, $curr_value ) {
		if ( $field_is_const = $this->options->has_constant_val( $input_name ) ) {
			$this->render_const_notice( $input_name );
		}
```

templates/back-to-auth0.php
```diff
@@ -13,5 +13,5 @@
 	}
 </style>
 <div id="extra-options">
-	<a href="?"><?php printf( _e( '← Back to %s login', 'wp-auth0' ), $title ); ?></a>
+	<a href="?"><?php printf( __( '← Back to %s login', 'wp-auth0' ), $title ); ?></a>
```
lbalmaceda: what does this?

tests/testConstantSettings.php
```diff
@@ -0,0 +1,189 @@
...
+	 */
+	public function testThatConstructorStoresConstants() {
+		// Set a few constant overrides.
+		define( self::DEFAULT_CONSTANT_PREFIX . 'DOMAIN', rand() );
+		define( self::DEFAULT_CONSTANT_PREFIX . 'CLIENT_ID', rand() );
+		define( self::DEFAULT_CONSTANT_PREFIX . 'CLIENT_SECRET', rand() );
+
+		// Make sure we have the right number of overrides.
+		$opts = new WP_Auth0_Options();
+		$this->assertCount( 3, $opts->get_all_constant_keys() );
```
lbalmaceda: you can go further and check the actual key values

tests/testConstantSettings.php
```diff
@@ -0,0 +1,189 @@
...
+			$override_val  = self::FILTER_TEST_STRING . rand();
+			define( $constant_name, $override_val );
+
+			ob_start();
+			$admin->{$field['render_function']}( $field );
+			$field_html = ob_get_clean();
+
+			$input = $this->getDomListFromTagName( $field_html, 'input' );
+			$this->assertTrue( $input->item( 0 )->hasAttribute( 'disabled' ) );
+			$this->assertNotFalse( strpos( $field_html, self::CONSTANT_NOTICE_TEXT ) );
```
lbalmaceda: `assertNotFalse` vs `assertTrue` ?

lib/WP_Auth0_Options_Generic.php
joshcanhelp: > is this update something you'll normally run after setting each key-value?

It's run most of the time, yes. 

> Since I guess is not going to perform well doing so after each call. 

Not excellent but we don't `set` much except in the initial setup process and, in that case, there are a number of HTTP requests that will be the main slow-down. The main time settings are saved is when the page is saved and, in that case, it's a single update. 

> If so, would it make sense to remove this parameter from this method

BC there so I'll punt that to when we refactor this whole system.

tests/testConstantSettings.php
joshcanhelp: `strpos()` returns the position of the string as an integer (could be `0`) or `FALSE` if it's not found. I want to check that the string is there so I want to assert that it's not false. 

That said, there is an `assertContains` that works better here so I switched to that.

PR #510, opened 2018-07-31
Remove account cleanup tool

joshcanhelp:
Removing unused account cleanup tool


cocojoe:
In the description you should say why you are doing this.  Thanks

PR #528, opened 2018-08-29
Add skip strategies setting and tests

joshcanhelp:
**Summary:** Add a setting to the admin that skips email verification for specific strategies. This is being added based on Enterprise strategies that do not provide an `email_verified` flag in the userinfo object. 

**Admin Setting:**

screenshot 2018-08-29 12 52 34

**Specific Changes:**

- Add the `WP_Auth0_UsersRepo::get_meta()` method to abstract getting Auth0 user meta information; refactored all current uses of the core `get_user_meta()` function to use this new method.
- Refactor `WP_Auth0_LoginManager::login_user()` to use the new skip strategies field. 
- Add `WP_Auth0_Options::strategy_skips_verified_email()` method to check whether a specific strategy can be skipped. 
- Refactor `WP_Auth0_UsersRepo::create()` to use the new skip strategies field and simplify logic. 
- Add tests for all new functionality and `WP_Auth0_UsersRepo::create()`

### DIFF
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -530,9 +530,7 @@ $a0_plugin->init();
 if ( ! function_exists( 'get_auth0userinfo' ) ) {
 	function get_auth0userinfo( $user_id ) {
 
-		global $wpdb;
-
-		$profile = get_user_meta( $user_id, $wpdb->prefix . 'auth0_obj', true );
+		$profile = WP_Auth0_UsersRepo::get_meta( $user_id, 'auth0_obj' );
 
 		if ( $profile ) {
 			return WP_Auth0_Serializer::unserialize( $profile );
@@ -558,17 +556,15 @@ if ( ! function_exists( 'get_currentauth0userinfo' ) ) {
 if ( ! function_exists( 'get_currentauth0user' ) ) {
 	function get_currentauth0user() {
 
-		global $wpdb;
-
 		$current_user = wp_get_current_user();
 
-		$serialized_profile = get_user_meta( $current_user->ID, $wpdb->prefix . 'auth0_obj', true );
+		$serialized_profile = WP_Auth0_UsersRepo::get_meta( $current_user->ID, 'auth0_obj' );
 
 		$data = new stdClass;
 
 		$data->auth0_obj   = empty( $serialized_profile ) ? false : WP_Auth0_Serializer::unserialize( $serialized_profile );
-		$data->last_update = get_user_meta( $current_user->ID, $wpdb->prefix . 'last_update', true );
-		$data->auth0_id    = get_user_meta( $current_user->ID, $wpdb->prefix . 'auth0_id', true );
+		$data->last_update = WP_Auth0_UsersRepo::get_meta( $current_user->ID, 'last_update' );
+		$data->auth0_id    = WP_Auth0_UsersRepo::get_meta( $current_user->ID, 'auth0_id' );
 
 		return $data;
 	}
--- a/lib/WP_Auth0_DBManager.php
+++ b/lib/WP_Auth0_DBManager.php
@@ -431,7 +431,7 @@ class WP_Auth0_DBManager {
 		$repo = new WP_Auth0_UsersRepo( $this->a0_options );
 
 		foreach ( $userRows as $row ) {
-			$auth0_id = get_user_meta( $row->wp_id, $wpdb->prefix . 'auth0_id', true );
+			$auth0_id = WP_Auth0_UsersRepo::get_meta( $row->wp_id, 'auth0_id' );
 
 			if ( ! $auth0_id ) {
 				$repo->update_auth0_object( $row->wp_id, WP_Auth0_Serializer::unserialize( $row->auth0_obj ) );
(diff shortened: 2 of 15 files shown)


WP_Auth0.php
```diff
@@ -526,9 +526,7 @@ private function autoloader( $class ) {
 if ( ! function_exists( 'get_auth0userinfo' ) ) {
```
joshcanhelp: Switching to new `WP_Auth0_UsersRepo::get_meta` helper function

lib/WP_Auth0_DBManager.php
```diff
@@ -431,7 +431,7 @@ protected function migrate_users_data() {
 		$repo = new WP_Auth0_UsersRepo( $this->a0_options );
```
joshcanhelp: Switching to new `WP_Auth0_UsersRepo::get_meta` helper function

lib/WP_Auth0_Export_Users.php
```diff
@@ -69,7 +69,8 @@ public function a0_export_users( $user_ids = null ) {
 		global $wpdb;
```
joshcanhelp: Switching to new `WP_Auth0_UsersRepo::get_meta` helper function

lib/WP_Auth0_UsersRepo.php
```diff
@@ -68,90 +68,83 @@ public function tokenHasRequiredScopes( $jwt ) {
...
+	 * @param null|string $role - TODO: Deprecate, not used
+	 * @param bool        $skip_email_verified - TODO: Deprecate, not used
+	 *
+	 * @return int|null|WP_Error
+	 *
+	 * @throws WP_Auth0_CouldNotCreateUserException
+	 * @throws WP_Auth0_EmailNotVerifiedException
+	 * @throws WP_Auth0_RegistrationNotEnabledException
+	 */
+	public function create( $userinfo, $token, $access_token = null, $role = null, $skip_email_verified = false ) {
```
joshcanhelp: This is the meat of this PR. This method was confusing to read and debug so just patching in the skip strategy logic would have probably taken as much time as a line-by-line re-write.

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -45,6 +45,12 @@ public function init() {
 				'id'       => 'wpa0_verified_email',
 				'function' => 'render_verified_email',
 			),
+			array(
```
joshcanhelp: Declare the new settings field.

lib/WP_Auth0_UsersRepo.php
```diff
@@ -68,90 +68,83 @@ public function tokenHasRequiredScopes( $jwt ) {
...
 			$user_id = WP_Auth0_Users::create_user( $userinfo, $role );
 
 			// Check if user was created
 			if ( is_wp_error( $user_id ) ) {
 				throw new WP_Auth0_CouldNotCreateUserException( $user_id->get_error_message() );
-			} elseif ( $user_id == -2 ) {
-				$msg = __( 'Could not create user. The registration process were rejected. Please verify that your account is whitelisted for this system. Please contact your site’s administrator.', 'wp-auth0' );
-
-				throw new WP_Auth0_CouldNotCreateUserException( $msg );
+			} elseif ( -2 === $user_id ) {
```
cocojoe: Where does -2 come from? Could this be a constant that better defines what this means?

cocojoe:
@joshcanhelp by *Default* it will skip Twitter? and Enterprise Strategies, where does that list come from?

joshcanhelp:
> by Default it will skip Twitter? and Enterprise Strategies, where does that list come from?

No, no default skipping, field will be blank by default.

lib/WP_Auth0_UsersRepo.php
joshcanhelp: It's what `WP_Auth0_Users::create_user()` returns for that condition. I didn't want to go much further with this PR but I totally agree that this magic value is not ideal 👍

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -780,7 +815,7 @@ public function link_accounts_validation( $old_options, $input ) {
 	 * @return array
 	 */
 	public function loginredirection_validation( $old_options, $input ) {
-		$new_redirect_url = strtolower( $input['default_login_redirection'] );
+		$new_redirect_url = esc_url_raw( strtolower( $input['default_login_redirection'] ) );
```
cocojoe: why does this one get an `esc_url_raw` but the $old_redirect_url doesn't?

cocojoe:
Should probably have an example of connection name entries in the `Skip Strategies` or could be placeholder?

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: We don't need to sanitize the old one to check it as it's already saved. If the new email is different after sanitization, then we update with that one (assuming it passes all the other checks).

PR #551, opened 2018-10-02
Rename the Auto Login setting to ULP; move to features tab

joshcanhelp:
### Changes

There has been some general confusion around how to use the ULP with this plugin. This PR moves the setting to be more visible and changes the name to more clearly explain what the setting does. I left a reference from the old setting to the new one to help with UX. 

### References

[Universal Login vs Embedded](https://auth0.com/docs/guides/login/universal-vs-embedded)

### Testing

No functionality changes so no testing needed.

### DIFF
--- a/lib/admin/WP_Auth0_Admin_Features.php
+++ b/lib/admin/WP_Auth0_Admin_Features.php
@@ -220,7 +220,7 @@ class WP_Auth0_Admin_Features extends WP_Auth0_Admin_Generic {
 		$this->render_switch( $args['label_for'], $args['opt_name'], 'wpa0_auto_login_method' );
 		$this->render_field_description(
 			__( 'Use the Universal Login Page (ULP) for authentication. ', 'wp-auth0' ) .
-			__( 'When turned on, <code>wp-login.php</code> will be redirected to the hosted login page. ', 'wp-auth0' ) .
+			__( 'When turned on, <code>wp-login.php</code> will redirect to the hosted login page. ', 'wp-auth0' ) .
 			__( 'When turned off, <code>wp-login.php</code> will show an embedded login form. ', 'wp-auth0' ) .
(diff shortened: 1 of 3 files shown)


lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -195,6 +207,46 @@ public function render_passwordless_enabled( $args = array() ) {
...
+	 *
+	 * @param array $args - callback args passed in from add_settings_field().
+	 *
+	 * @see WP_Auth0_Admin_Generic::init_option_section()
+	 * @see add_settings_field()
+	 */
+	public function render_auto_login( $args = array() ) {
+		$this->render_switch( $args['label_for'], $args['opt_name'], 'wpa0_auto_login_method' );
+		$this->render_field_description(
+			__( 'Use the Universal Login Page (ULP) for authentication. ', 'wp-auth0' ) .
```
luisrudge: this is the only place you use `(ULP)` (in this PR at least). not a deal breaker for me, your call.

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -195,6 +207,46 @@ public function render_passwordless_enabled( $args = array() ) {
...
+			__( 'When turned on, <code>wp-login.php</code> will be redirected to the hosted login page. ', 'wp-auth0' ) .
```
luisrudge: will redirect instead of will be redirected?

lib/admin/WP_Auth0_Admin_Features.php
joshcanhelp: @luisrudge - Used on line 225 below as well.

luisrudge: damn tunnel vision. I was looking in other code blocks 😬

PR #557, opened 2018-10-05
Deprecate unused login methods and props

joshcanhelp:
- Deprecated unused `$admin_role` and `$ignore_unverified_email` parameters in the `WP_Auth0_LoginManager` construct method
- Deprecated unused `end_session()` method in the `WP_Auth0_LoginManager` class
- Deprecate `$role` parameter in the `create_user` method in the `WP_Auth0_Users` class
- Deprecate `$access_token`, `$role`, and `$skip_email_verified` parameters in the `create` method of the `WP_Auth0_UsersRepo` class

**No functional changes 👍**

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -24,7 +24,7 @@ class WP_Auth0_LoginManager {
 	/**
 	 * Should the new user have an administrator role?
 	 *
-	 * TODO: Deprecate, not used
+	 * @deprecated - 3.8.0
 	 *
 	 * @var bool|null
 	 */
@@ -33,7 +33,7 @@ class WP_Auth0_LoginManager {
 	/**
 	 * Ignore verified email requirement in Settings > Advanced.
 	 *
-	 * TODO: Deprecate, not used
+	 * @deprecated - 3.8.0
 	 *
 	 * @var bool
 	 */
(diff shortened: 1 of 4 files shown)


lib/WP_Auth0_LoginManager.php
```diff
@@ -774,6 +772,28 @@ protected function die_on_login( $msg = '', $code = 0, $login_link = true ) {
 		);
 	}
 
+	/*
+	 *
+	 * DEPRECATED
```
lbalmaceda: why this flying on its own away from a valid method?

lib/WP_Auth0_Settings_Section.php
```diff
@@ -65,8 +65,5 @@ public function init_menu() {
 		}
 	}
 
-	// TODO: deprecate
-	public function redirect_to_help() {
-
-	}
+	public function redirect_to_help() {}
```
lbalmaceda: is it worthy to add the `deprecated` docblock here as well?

lib/WP_Auth0_Users.php
```diff
@@ -1,5 +1,14 @@
 <?php
 class WP_Auth0_Users {
+
+	/**
+	 * Create a WordPress user with Auth0 data.
+	 *
+	 * @param object       $userinfo - User profile data from Auth0.
+	 * @param null|boolean $role - Set the sole as administrator - @deprecated - 3.8.0.
```
lbalmaceda: sole is role?

lib/WP_Auth0_LoginManager.php
joshcanhelp: It's not, just a section indicator, everything below is deprecated

lib/WP_Auth0_Settings_Section.php
joshcanhelp: I'm not entirely sure what to do with it yet. Serves no purpose but is used in a hook above. Will address when this class gets a once-over 👍

PR #558, opened 2018-10-05
Deprecate a0_render_message method

joshcanhelp:
Deprecation docblocks and error triggering for the unused `a0_render_message` method in the `WP_Auth0` class.

**No functional changes 👍**

### DIFF
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -488,6 +488,9 @@ class WP_Auth0 {
 	 * @codeCoverageIgnore - Deprecated
 	 */
 	public function a0_render_message() {
+		// phpcs:ignore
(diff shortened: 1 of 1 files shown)


WP_Auth0.php
```diff
@@ -494,6 +482,22 @@ private function autoloader( $class ) {
...
 	 */
 
+	/**
+	 * @deprecated - 3.8.0, not used and no replacement provided.
+	 *
+	 * @codeCoverageIgnore - Deprecated
+	 */
+	public function a0_render_message() {
+		$message = null;
+
```
lbalmaceda: shouldn't this print the warning like you do in other methods?

```php
trigger_error( sprintf( __( 'Method %s is deprecated.', 'wp-auth0' ), __METHOD__ ), E_USER_DEPRECATED );
```

PR #559, opened 2018-10-05
Deprecate WP_Auth0_Email_Verification::ajax_resend_email

joshcanhelp:
Deprecate the `WP_Auth0_Email_Verification::ajax_resend_email()` method.

**No functional changes** 👍


lib/WP_Auth0_Email_Verification.php
```diff
@@ -109,6 +94,31 @@ public function resend_verification_email() {
...
 	}
+
+	/*
+	 *
+	 * DEPRECATED
+	 *
+	 */
+
+	/**
+	 * AJAX handler to request that the verification email be resent.
```
lbalmaceda: resent -> re-sent ?

PR #565, opened 2018-10-08
Fix tests to run in same process

joshcanhelp:
### Changes

Changed the test runner to run in the same process, speeding tests up by a lot. 

### References

[PHPUnit annotations](https://phpunit.de/manual/6.5/en/appendixes.annotations.html)

### Testing

Changes to test suite only. 

### Checklist

### DIFF
--- a/lib/WP_Auth0_EditProfile.php
+++ b/lib/WP_Auth0_EditProfile.php
@@ -59,7 +59,6 @@ class WP_Auth0_EditProfile {
 	 * @codeCoverageIgnore
 	 */
 	public function admin_enqueue_scripts() {
-		global $user_id;
 		global $pagenow;
 
 		if ( ! in_array( $pagenow, array( 'profile.php', 'user-edit.php' ) ) ) {
@@ -73,14 +72,14 @@ class WP_Auth0_EditProfile {
 			WPA0_VERSION
 		);
 
-		$profile  = get_auth0userinfo( $user_id );
+		$profile  = get_auth0userinfo( $GLOBALS['user_id'] );
 		$strategy = isset( $profile->sub ) ? WP_Auth0_Users::get_strategy( $profile->sub ) : '';
 
 		wp_localize_script(
 			'wpa0_user_profile',
 			'wpa0UserProfile',
 			array(
-				'userId'         => intval( $user_id ),
+				'userId'         => intval( $GLOBALS['user_id'] ),
 				'userStrategy'   => sanitize_text_field( $strategy ),
 				'deleteIdNonce'  => wp_create_nonce( 'delete_auth0_identity' ),
 				'deleteMfaNonce' => wp_create_nonce( 'delete_auth0_mfa' ),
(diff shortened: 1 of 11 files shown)


lib/WP_Auth0_EditProfile.php
```diff
@@ -73,14 +72,14 @@ public function admin_enqueue_scripts() {
 			WPA0_VERSION
 		);
 
-		$profile  = get_auth0userinfo( $user_id );
+		$profile  = get_auth0userinfo( $GLOBALS['user_id'] );
```
joshcanhelp: Different way to access globals in PHP, works better for testing.

lib/profile/WP_Auth0_Profile_Delete_Data.php
```diff
@@ -39,13 +39,12 @@ public function init() {
 	 * IMPORTANT: Internal callback use only, do not call this function directly!
 	 */
 	public function show_delete_identity() {
-		global $user_id;
 
-		if ( ! current_user_can( 'edit_users', $user_id ) ) {
+		if ( ! isset( $GLOBALS['user_id'] ) || ! current_user_can( 'edit_users', $GLOBALS['user_id'] ) ) {
```
joshcanhelp: Different way to access globals in PHP, works better for testing.

lib/profile/WP_Auth0_Profile_Delete_Mfa.php
```diff
@@ -51,16 +51,16 @@ public function init() {
 	 * IMPORTANT: Internal callback use only, do not call this function directly!
 	 */
 	public function show_delete_mfa() {
-		global $user_id;
-		if ( ! current_user_can( 'edit_users', $user_id ) ) {
+
+		if ( ! isset( $GLOBALS['user_id'] ) || ! current_user_can( 'edit_users', $GLOBALS['user_id'] ) ) {
```
joshcanhelp: Different way to access globals in PHP, works better for testing.

phpunit.xml.dist
```diff
@@ -1,12 +1,11 @@
 <?xml version="1.0"?>
 <phpunit
 	bootstrap="tests/bootstrap.php"
-	backupGlobals="false"
+	backupGlobals="true"
```
joshcanhelp: Keep the global vars the same between tests

tests/testConstantSettings.php
```diff
@@ -39,6 +39,8 @@ class TestConstantSettings extends TestCase {
 
 	/**
 	 * Test that setting a constant will store the constant key.
+	 *
+	 * @runInSeparateProcess
```
joshcanhelp: Separate process required here because constants are set (cannot be unset)

joshcanhelp:
@cocojoe - They do in practice but while testing, it was causing issues, possibly with how PHPUnit backs them up between tests.

cocojoe:
I wondered if it had to do with that, the one off use in a method is probably clearer than using `global xxx`. 
In PHP docs both methods are legit.

PR #576, opened 2018-10-29
Update JWT library

joshcanhelp:
### Changes

- Update Firebase JWT library from v2.0.0 to v5.0.0  and removed extra library files (tests, meta, etc)
- Remove fallback to `/userinfo`, replace with ID token 
- Move ID token cleanup to new `WP_Auth0_LoginManager::clean_id_token()`
- Add `InvalidArgumentException` catching in `WP_Auth0_LoginManager::init_auth0()`
- Remove exception handling in `WP_Auth0_LoginManager::implicit_login()` (handled in calling method)
- Add a higher leeway 
- Update autoloader for new path

### References

- [JWT library changes](https://github.com/firebase/php-jwt/compare/v2.0.0...master) ([commit here](https://github.com/auth0/wp-auth0/pull/576/commits/04cb28cfd90bf6d2816552aace12c46682237980))

### Testing

No testing is in place for the login flow as it's highly coupled.


lib/WP_Auth0_LoginManager.php
```diff
@@ -283,25 +287,13 @@ public function redirect_login() {
...
-
-			if ( 200 !== $userinfo_resp_code || empty( $userinfo_resp_body ) ) {
-
-				WP_Auth0_ErrorManager::insert_auth0_error( __METHOD__ . ' L:' . __LINE__, $userinfo_resp );
-				throw new WP_Auth0_LoginFlowValidationException(
-					__( 'Error getting user information', 'wp-auth0' ),
-					$userinfo_resp_code
-				);
-			}
+		// Management API call failed, fallback to ID token.
```
cocojoe: What would be the expectation for a management API call to fail?

cocojoe:
Code is mostly the JWT Lib changes, thanks for adding you changed leeway as it's hard to tell if you changed anything vs changes that came from the update. That's a big jump in Majors from 2.0 to 5.0.

lib/WP_Auth0_LoginManager.php
joshcanhelp: If Client Credential Grant is turned off

PR #582, opened 2018-11-07
Fix Connection update over-writing Connection settings.

joshcanhelp:
### Changes

This fixes a bug that can cause unintended changes to database connection settings when saving a new password policy. Specific changes:

- Add object property checking to `WP_Auth0_Api_Client::update_connection()` to avoid unnecessary payload errors from the server.
- Remove duplicate connection cleanup from `WP_Auth0_Api_Operations::update_wordpress_connection()`
- Return old password policy setting if an API call fails during validation
- Pass entire connection object to the update API call to avoid unintended settings changes in password policy validation and new connection creation
- Add a large number of tests to confirm behavior
- Docblocks to help investigation and refactoring in the future

### References

- [WP forum report](https://wordpress.org/support/topic/db-migration-automaticly-disable-it-self-after-plugin-save-in-wp-admin/)

### Testing

#### Scenario 1: Updates to the password policy in wp-admin

1. Start with a working WP-Auth0 install with a DB connection active
2. In the Auth0 dashboard, make sure the connection has a non-default setting (username required turned ON or custom DB migration)
3. In wp-admin, change the password policy and save

**Expected behavior:**

Password policy is changed, other settings are not affected.

#### Scenario 2: Connection creation during the setup wizard

1. Start with a WP install without WP-Auth0 installed
2. Activate the plugin and walk through the setup wizard (any option)

**Expected behavior:**

Install completes without errors in WordPress; New client created associated to new DB connection in Auth0.

### Checklist

### DIFF
--- a/lib/WP_Auth0_Api_Client.php
+++ b/lib/WP_Auth0_Api_Client.php
@@ -611,6 +611,17 @@ class WP_Auth0_Api_Client {
 		return json_decode( $response['body'] );
 	}
 
+	/**
+	 * Update a Connection via the Management API.
+	 * Note: $payload must be a complete settings object, not just the property to change.
+	 *
+	 * @param string   $domain - Auth0 Domain.
+	 * @param string   $app_token - Valid Auth0 Management API token.
+	 * @param string   $id - DB Connection ID.
+	 * @param stdClass $payload - DB Connection settings, will override existing.
+	 *
+	 * @return bool|object
+	 */
 	public static function update_connection( $domain, $app_token, $id, $payload ) {
 		$endpoint = "https://$domain/api/v2/connections/$id";
 
@@ -618,6 +629,14 @@ class WP_Auth0_Api_Client {
 
 		$headers['Authorization'] = "Bearer $app_token";
 		$headers['content-type']  = 'application/json';
+
+		unset( $payload->name );
+		unset( $payload->strategy );
+		unset( $payload->id );
+
+		if ( ! empty( $payload->enabled_clients ) ) {
+			$payload->enabled_clients = array_values( $payload->enabled_clients );
+		}
 
 		$response = wp_remote_post(
 			$endpoint,
--- a/lib/WP_Auth0_Api_Operations.php
+++ b/lib/WP_Auth0_Api_Operations.php
@@ -30,10 +30,6 @@ class WP_Auth0_Api_Operations {
 		$connection->options->customScripts->login    = $login_script;
 		$connection->options->customScripts->get_user = $get_user_script;
 
-		unset( $connection->name );
-		unset( $connection->strategy );
-		unset( $connection->id );
-
 		WP_Auth0_Api_Client::update_connection( $domain, $app_token, $connection_id, $connection );
 
 	}
@@ -99,34 +95,10 @@ class WP_Auth0_Api_Operations {
 		$response = WP_Auth0_Api_Client::create_connection( $domain, $app_token, $body );
 
 		if ( $response === false ) {
-
 			return false;
-
 		}
 
-		$connections   = WP_Auth0_Api_Client::search_connection( $domain, $app_token, 'auth0' );
-		$db_connection = null;
-
-		$migration_connection_id = $response->id;
-
-		foreach ( $connections as $connection ) {
-			if ( $migration_connection_id != $connection->id && in_array( $client_id, $connection->enabled_clients ) ) {
-				$db_connection = $connection;
-
-				$enabled_clients = array_diff( $db_connection->enabled_clients, array( $client_id ) );
-
-				WP_Auth0_Api_Client::update_connection(
-					$domain,
-					$app_token,
-					$db_connection->id,
-					array(
-						'enabled_clients' => array_values( $enabled_clients ),
-					)
-				);
-			}
-		}
-
-		return $migration_connection_id;
+		return $response->id;
 	}
 
 	// $input['geo_rule'] = ( isset( $input['geo_rule'] ) ? $input['geo_rule'] : 0 );
--- a/lib/admin/WP_Auth0_Admin_Advanced.php
+++ b/lib/admin/WP_Auth0_Admin_Advanced.php
@@ -741,10 +741,6 @@ class WP_Auth0_Admin_Advanced extends WP_Auth0_Admin_Generic {
 					$connection->options->enabledDatabaseCustomization = false;
 					$connection->options->import_mode                  = false;
 
-					unset( $connection->name );
-					unset( $connection->strategy );
-					unset( $connection->id );
-
 					$response = WP_Auth0_Api_Client::update_connection( $input['domain'], $input['auth0_app_token'], $old_options['db_connection_id'], $connection );
 				} else {
 					$response = false;
--- a/lib/admin/WP_Auth0_Admin_Features.php
+++ b/lib/admin/WP_Auth0_Admin_Features.php
@@ -427,8 +427,10 @@ class WP_Auth0_Admin_Features extends WP_Auth0_Admin_Generic {
 
 			foreach ( $connections as $connection ) {
 				if ( in_array( $input['client_id'], $connection->enabled_clients ) ) {
-					$patch       = array( 'options' => array( 'passwordPolicy' => $input['password_policy'] ) );
-					$update_resp = WP_Auth0_Api_Client::update_connection( $domain, $app_token, $connection->id, $patch );
+					$u_connection                          = clone $connection;
+					$u_connection->options->passwordPolicy = $input['password_policy'];
+
+					$update_resp = WP_Auth0_Api_Client::update_connection( $domain, $app_token, $u_connection->id, $u_connection );
 
 					if ( false === $update_resp ) {
 						$this->add_validation_error(
--- a/lib/admin/WP_Auth0_Admin_Features.php
+++ b/lib/admin/WP_Auth0_Admin_Features.php
@@ -423,6 +423,8 @@ class WP_Auth0_Admin_Features extends WP_Auth0_Admin_Generic {
 					__( 'No database connections found for this application. ', 'wp-auth0' ) .
 					$this->get_dashboard_link( 'connections/database', __( 'See all database connections', 'wp-auth0' ) )
 				);
+				$input['password_policy'] = $old_options['password_policy'];
+				return $input;
 			}
 
 			foreach ( $connections as $connection ) {
@@ -438,6 +440,7 @@ class WP_Auth0_Admin_Features extends WP_Auth0_Admin_Generic {
 							__( 'Please manually review and update the policy. ', 'wp-auth0' ) .
 							$this->get_dashboard_link( 'connections/database', __( 'See all database connections', 'wp-auth0' ) )
 						);
+						$input['password_policy'] = $old_options['password_policy'];
 					}
 				}
 			}
--- a/lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
+++ b/lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
@@ -149,19 +149,24 @@ class WP_Auth0_InitialSetup_Consent {
 		if ( $should_create_and_update_connection && ! empty( $connections ) && is_array( $connections ) ) {
 			foreach ( $connections as $connection ) {
 
-				if ( in_array( $client_id, $connection->enabled_clients ) ) {
-
-					if ( $db_connection_name === $connection->name ) {
-						$connection_exists = $connection->id;
-						if ( isset( $connection->options ) && isset( $connection->options->passwordPolicy ) ) {
-							$connection_pwd_policy = $connection->options->passwordPolicy;
-						}
-					} else {
-						$u_connection                  = clone $connection;
-						$u_connection->enabled_clients = array_diff( $u_connection->enabled_clients, array( $client_id ) );
-						WP_Auth0_Api_Client::update_connection( $domain, $app_token, $u_connection->id, $u_connection );
-					}
+				// We only want to check/update connections that are active for this Application.
+				if ( ! in_array( $client_id, $connection->enabled_clients ) ) {
+					continue;
 				}
+
+				// Connection with the same name, use it below.
+				if ( $db_connection_name === $connection->name ) {
+					$connection_exists = $connection->id;
+					if ( isset( $connection->options ) && isset( $connection->options->passwordPolicy ) ) {
+						$connection_pwd_policy = $connection->options->passwordPolicy;
+					}
+					continue;
+				}
+
+				// Different Connection, update to remove the Application created above.
+				$u_connection                  = clone $connection;
+				$u_connection->enabled_clients = array_diff( $u_connection->enabled_clients, array( $client_id ) );
+				WP_Auth0_Api_Client::update_connection( $domain, $app_token, $u_connection->id, $u_connection );
 			}
 		}
 
--- a/lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
+++ b/lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
@@ -16,6 +16,16 @@ class WP_Auth0_InitialSetup_Consent {
 	public function render( $step ) {
 	}
 
+	/**
+	 * Used by both Setup Wizard installation flows.
+	 * Called in WP_Auth0_InitialSetup_ConnectionProfile::callback() when an API token is used during install.
+	 * Called in self::callback() when returning from consent URL install.
+	 *
+	 * @param string $domain - Auth0 domain for the Application.
+	 * @param string $access_token - Management API access token.
+	 * @param string $type - Installation type, "social" (AKA standard) or "enterprise".
+	 * @param bool   $hasInternetConnection - True if the installing site be reached by Auth0, false if not.
+	 */
 	public function callback_with_token( $domain, $access_token, $type, $hasInternetConnection = true ) {
 
 		$this->a0_options->set( 'auth0_app_token', $access_token );
@@ -92,6 +102,12 @@ class WP_Auth0_InitialSetup_Consent {
 		return $obj->access_token;
 	}
 
+	/**
+	 * Used by both Setup Wizard installation flows.
+	 * Called by self::callback_with_token() to create a Client, Connection, and Client Grant.
+	 *
+	 * @param string $name - Client name to use.
+	 */
 	public function consent_callback( $name ) {
 
 		$domain    = $this->a0_options->get( 'domain' );
@@ -128,19 +144,22 @@ class WP_Auth0_InitialSetup_Consent {
 		$connection_exists     = false;
 		$connection_pwd_policy = null;
 
-		$connections = WP_Auth0_Api_Client::search_connection( $domain, $app_token );
+		$connections = WP_Auth0_Api_Client::search_connection( $domain, $app_token, 'auth0' );
 
-		foreach ( $connections as $connection ) {
+		if ( $should_create_and_update_connection && ! empty( $connections ) && is_array( $connections ) ) {
+			foreach ( $connections as $connection ) {
 
-			if ( in_array( $client_id, $connection->enabled_clients ) ) {
-				if ( $connection->strategy === 'auth0' && $should_create_and_update_connection ) {
+				if ( in_array( $client_id, $connection->enabled_clients ) ) {
 
 					if ( $db_connection_name === $connection->name ) {
-						$connection_exists     = $connection->id;
-						$connection_pwd_policy = ( isset( $connection->options ) && isset( $connection->options->passwordPolicy ) ) ? $connection->options->passwordPolicy : null;
+						$connection_exists = $connection->id;
+						if ( isset( $connection->options ) && isset( $connection->options->passwordPolicy ) ) {
+							$connection_pwd_policy = $connection->options->passwordPolicy;
+						}
 					} else {
-						$enabled_clients = array_diff( $connection->enabled_clients, array( $client_id ) );
-						WP_Auth0_Api_Client::update_connection( $domain, $app_token, $connection->id, array( 'enabled_clients' => array_values( $enabled_clients ) ) );
+						$u_connection                  = clone $connection;
+						$u_connection->enabled_clients = array_diff( $u_connection->enabled_clients, array( $client_id ) );
+						WP_Auth0_Api_Client::update_connection( $domain, $app_token, $u_connection->id, $u_connection );
 					}
 				}
 			}
@@ -191,6 +210,6 @@ class WP_Auth0_InitialSetup_Consent {
 		}
 
 		wp_redirect( admin_url( 'admin.php?page=wpa0-setup&step=2&profile=' . $this->state ) );
-		exit();
+		exit;
 	}
 }
--- a/tests/testAdminFeatures.php
+++ b/tests/testAdminFeatures.php
@@ -1,0 +1,215 @@
+<?php
+/**
+ * Contains Class TestAdminFeatures.
+ *
+ * @package WP-Auth0
+ *
+ * @since 3.8.1
+ */
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * Class TestAdminFeatures.
+ */
+class TestAdminFeatures extends TestCase {
+
+	use HttpHelpers;
+
+	use RedirectHelpers;
+
+	use SetUpTestDb;
+
+	/**
+	 * Instance of WP_Auth0_Options.
+	 *
+	 * @var WP_Auth0_Options
+	 */
+	public static $opts;
+
+	/**
+	 * WP_Auth0_ErrorLog instance.
+	 *
+	 * @var WP_Auth0_ErrorLog
+	 */
+	protected static $error_log;
+
+	/**
+	 * Runs once before all tests.
+	 */
+	public static function setUpBeforeClass() {
+		parent::setUpBeforeClass();
+		self::$opts      = WP_Auth0_Options::Instance();
+		self::$error_log = new WP_Auth0_ErrorLog();
+	}
+
+	/**
+	 * Test that the validation does not run if the current value matches the old value.
+	 */
+	public function testThatInputIsNotValidatedIfItIsNotChanged() {
+		$admin_features = new WP_Auth0_Admin_Features( self::$opts );
+
+		$old_input = [ 'password_policy' => 'fair' ];
+		$new_input = [ 'password_policy' => 'fair' ];
+		$result    = $admin_features->security_validation( $old_input, $new_input );
+
+		$this->assertEquals( $old_input, $result );
+
+		global $wp_settings_errors;
+		$this->assertEmpty( $wp_settings_errors );
+	}
+
+	/**
+	 * Test the connection search request structure.
+	 */
+	public function testThatConnectionsAreSearchedDuringValidation() {
+		$this->startHttpHalting();
+
+		$test_domain    = 'test-wp.auth0.com';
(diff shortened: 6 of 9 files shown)


lib/WP_Auth0_Api_Operations.php
```diff
@@ -30,10 +30,6 @@ public function update_wordpress_connection( $app_token, $connection_id, $passwo
 		$connection->options->customScripts->login    = $login_script;
 		$connection->options->customScripts->get_user = $get_user_script;
 
-		unset( $connection->name );
```
joshcanhelp: Moved to `WP_Auth0_Api_Client::update_connection()`

lib/WP_Auth0_Api_Operations.php
```diff
@@ -107,26 +103,21 @@ public function create_wordpress_connection( $app_token, $migration_enabled, $pa
 		$connections   = WP_Auth0_Api_Client::search_connection( $domain, $app_token, 'auth0' );
 		$db_connection = null;
 
-		$migration_connection_id = $response->id;
+		$created_connection_id = $response->id;
```
joshcanhelp: Confusing var name

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -741,10 +741,6 @@ public function migration_ws_validation( $old_options, $input ) {
 					$connection->options->enabledDatabaseCustomization = false;
 					$connection->options->import_mode                  = false;
 
-					unset( $connection->name );
```
joshcanhelp: Moved to `WP_Auth0_Api_Client::update_connection()`

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -427,8 +427,8 @@ public function security_validation( $old_options, $input ) {
 
 			foreach ( $connections as $connection ) {
 				if ( in_array( $input['client_id'], $connection->enabled_clients ) ) {
-					$patch       = array( 'options' => array( 'passwordPolicy' => $input['password_policy'] ) );
-					$update_resp = WP_Auth0_Api_Client::update_connection( $domain, $app_token, $connection->id, $patch );
+					$connection['options']['passwordPolicy'] = $input['password_policy'];
```
joshcanhelp: Here's the bug ... `$patch` did not include existing Connection settings

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -128,19 +144,21 @@ public function consent_callback( $name ) {
 		$connection_exists     = false;
 		$connection_pwd_policy = null;
 
-		$connections = WP_Auth0_Api_Client::search_connection( $domain, $app_token );
+		$connections = WP_Auth0_Api_Client::search_connection( $domain, $app_token, 'auth0' );
```
joshcanhelp: Added strategy filter here to reduce the number of Connections returned.

lib/WP_Auth0_Api_Client.php
```diff
@@ -619,6 +630,14 @@ public static function update_connection( $domain, $app_token, $id, $payload ) {
 		$headers['Authorization'] = "Bearer $app_token";
 		$headers['content-type']  = 'application/json';
 
+		unset( $payload->name );
```
joshcanhelp: These must not be set or the API endpoint returns a payload error.

lib/WP_Auth0_Api_Client.php
```diff
@@ -619,6 +630,14 @@ public static function update_connection( $domain, $app_token, $id, $payload ) {
...
+		unset( $payload->strategy );
+		unset( $payload->id );
+
+		if ( ! empty( $payload->enabled_clients ) ) {
```
joshcanhelp: If the array is not consecutive then `json_encode` converts this to an object rather than an array.

lib/WP_Auth0_Api_Operations.php
```diff
@@ -99,34 +95,10 @@ public function create_wordpress_connection( $app_token, $migration_enabled, $pa
...
 			return false;
-
-		}
-
-		$connections   = WP_Auth0_Api_Client::search_connection( $domain, $app_token, 'auth0' );
-		$db_connection = null;
-
-		$migration_connection_id = $response->id;
-
-		foreach ( $connections as $connection ) {
```
joshcanhelp: This method is only called in one place and the calling method already does this check.

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -128,19 +144,21 @@ public function consent_callback( $name ) {
...
 
-		foreach ( $connections as $connection ) {
+		if ( $should_create_and_update_connection && ! empty( $connections ) && is_array( $connections ) ) {
```
joshcanhelp: Logic here is still not great but did not want to completely refactor this in a patch release. Main thing here is checking whether we have an array and avoiding the loop if we're not using the values later.

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -128,19 +144,21 @@ public function consent_callback( $name ) {
...
 
-			if ( in_array( $client_id, $connection->enabled_clients ) ) {
-				if ( $connection->strategy === 'auth0' && $should_create_and_update_connection ) {
+				if ( in_array( $client_id, $connection->enabled_clients ) ) {
 
 					if ( $db_connection_name === $connection->name ) {
-						$connection_exists     = $connection->id;
-						$connection_pwd_policy = ( isset( $connection->options ) && isset( $connection->options->passwordPolicy ) ) ? $connection->options->passwordPolicy : null;
+						$connection_exists = $connection->id;
+						if ( isset( $connection->options ) && isset( $connection->options->passwordPolicy ) ) {
```
joshcanhelp: Ternary was too long and confusing to read.

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -128,19 +144,21 @@ public function consent_callback( $name ) {
...
+							$connection_pwd_policy = $connection->options->passwordPolicy;
+						}
 					} else {
-						$enabled_clients = array_diff( $connection->enabled_clients, array( $client_id ) );
-						WP_Auth0_Api_Client::update_connection( $domain, $app_token, $connection->id, array( 'enabled_clients' => array_values( $enabled_clients ) ) );
+						$connection->enabled_clients = array_diff( $connection->enabled_clients, array( $client_id ) );
```
joshcanhelp: Avoiding the same settings override as reported in the original bug.

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -128,19 +144,22 @@ public function consent_callback( $name ) {
...
+			foreach ( $connections as $connection ) {
 
-			if ( in_array( $client_id, $connection->enabled_clients ) ) {
-				if ( $connection->strategy === 'auth0' && $should_create_and_update_connection ) {
+				if ( in_array( $client_id, $connection->enabled_clients ) ) {
 
 					if ( $db_connection_name === $connection->name ) {
-						$connection_exists     = $connection->id;
-						$connection_pwd_policy = ( isset( $connection->options ) && isset( $connection->options->passwordPolicy ) ) ? $connection->options->passwordPolicy : null;
+						$connection_exists = $connection->id;
```
lbalmaceda: Can this be renamed? I get why you'd call it like that, but down in the code you store "connection exists" as the "connection id". I can't think now of a better name 😛

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -128,19 +144,22 @@ public function consent_callback( $name ) {
...
+						if ( isset( $connection->options ) && isset( $connection->options->passwordPolicy ) ) {
+							$connection_pwd_policy = $connection->options->passwordPolicy;
+						}
 					} else {
-						$enabled_clients = array_diff( $connection->enabled_clients, array( $client_id ) );
-						WP_Auth0_Api_Client::update_connection( $domain, $app_token, $connection->id, array( 'enabled_clients' => array_values( $enabled_clients ) ) );
+						$u_connection                  = clone $connection;
```
lbalmaceda: Add a comment above this line: "//disable connection for this application/client id"

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -191,6 +210,6 @@ public function consent_callback( $name ) {
 		}
 
 		wp_redirect( admin_url( 'admin.php?page=wpa0-setup&step=2&profile=' . $this->state ) );
-		exit();
+		exit;
```
lbalmaceda: was this change intended?

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -423,19 +423,24 @@ public function security_validation( $old_options, $input ) {
 					__( 'No database connections found for this application. ', 'wp-auth0' ) .
 					$this->get_dashboard_link( 'connections/database', __( 'See all database connections', 'wp-auth0' ) )
 				);
+				$input['password_policy'] = $old_options['password_policy'];
+				return $input;
```
joshcanhelp: Exit here to skip anything else that might process (`foreach` will throw an error if it's given anything other than an array)

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -423,19 +423,24 @@ public function security_validation( $old_options, $input ) {
 					__( 'No database connections found for this application. ', 'wp-auth0' ) .
 					$this->get_dashboard_link( 'connections/database', __( 'See all database connections', 'wp-auth0' ) )
 				);
+				$input['password_policy'] = $old_options['password_policy'];
```
joshcanhelp: Nothing changed at Auth0 so we should not change the setting in WordPress either.

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -423,19 +423,24 @@ public function security_validation( $old_options, $input ) {
...
+
+					$update_resp = WP_Auth0_Api_Client::update_connection( $domain, $app_token, $u_connection->id, $u_connection );
 
 					if ( false === $update_resp ) {
 						$this->add_validation_error(
 							__( 'There was a problem updating the password policy. ', 'wp-auth0' ) .
 							__( 'Please manually review and update the policy. ', 'wp-auth0' ) .
 							$this->get_dashboard_link( 'connections/database', __( 'See all database connections', 'wp-auth0' ) )
 						);
+						$input['password_policy'] = $old_options['password_policy'];
```
joshcanhelp: Nothing changed at Auth0 so we should not change the setting in WordPress either.

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -423,19 +423,24 @@ public function security_validation( $old_options, $input ) {
 					__( 'No database connections found for this application. ', 'wp-auth0' ) .
```
lbalmaceda: Can't leave a comment for the line 416 
```
 		if ( $old_options['password_policy'] !== $input['password_policy'] ) {
```
but seems like you could check the opposite value and return early instead of nesting everything here.

lib/admin/WP_Auth0_Admin_Features.php
lbalmaceda: I don't see where in the code you're saving the "new value" *before* doing the request, but this line seems to add nothing. If anything, the `update_connection` method should NOT modify options unless the request was successful (as part of the method logic). Why would it be able to modify the options if it hasn't checked the request results first?

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -128,19 +144,22 @@ public function consent_callback( $name ) {
...
+						$connection_exists = $connection->id;
+						if ( isset( $connection->options ) && isset( $connection->options->passwordPolicy ) ) {
+							$connection_pwd_policy = $connection->options->passwordPolicy;
+						}
 					} else {
-						$enabled_clients = array_diff( $connection->enabled_clients, array( $client_id ) );
-						WP_Auth0_Api_Client::update_connection( $domain, $app_token, $connection->id, array( 'enabled_clients' => array_values( $enabled_clients ) ) );
+						$u_connection                  = clone $connection;
+						$u_connection->enabled_clients = array_diff( $u_connection->enabled_clients, array( $client_id ) );
+						WP_Auth0_Api_Client::update_connection( $domain, $app_token, $u_connection->id, $u_connection );
```
lbalmaceda: what if any of this requests fail? How are you going to capture those errors and what do you plan to do next?

tests/testAdminFeatures.php
```diff
@@ -0,0 +1,215 @@
...
+		$new_input = [
+			'password_policy' => 'good',
+			'domain'          => $test_domain,
+			'auth0_app_token' => $test_token,
+		];
+
+		$caught_request = [];
+		try {
+			$admin_features->security_validation( $old_input, $new_input );
+		} catch ( Exception $e ) {
```
lbalmaceda: why would the above call throw an exception?

tests/testAdminFeatures.php
```diff
@@ -0,0 +1,215 @@
...
+		$new_input = [
+			'password_policy' => 'good',
+			'domain'          => 'test-wp.auth0.com',
+			'auth0_app_token' => implode( '.', [ uniqid(), uniqid(), uniqid() ] ),
+			'client_id'       => 'TEST_CLIENT_ID',
+		];
+
+		$this->http_request_type = [ 'success_get_connections', 'halt' ];
+
+		$caught_request = [ 'Nothing caught' ];
```
lbalmaceda: sounds odd that below you're treating this as a dictionary but here you store things as an array.

tests/testAdminFeatures.php
```diff
@@ -0,0 +1,215 @@
...
+		try {
+			$admin_features->security_validation( $old_input, $new_input );
+		} catch ( Exception $e ) {
+			$caught_request = unserialize( $e->getMessage() );
+		}
+
+		$this->assertEquals( 'https://test-wp.auth0.com/api/v2/connections/TEST_CONN_ID', $caught_request['url'] );
```
lbalmaceda: where is the method getting the `TEST_CONN_ID` from?

tests/testAdminFeatures.php
```diff
@@ -0,0 +1,215 @@
...
+		$this->assertEquals( $new_input['password_policy'], $result['password_policy'] );
+
+		global $wp_settings_errors;
+		$this->assertEmpty( $wp_settings_errors );
+	}
+
+	/**
+	 * Runs after each test method.
+	 */
+	public function tearDown() {
```
lbalmaceda: better if all the "pre/post tests" methods are together, usually on the top of the class.

tests/testApiOperations.php
```diff
@@ -0,0 +1,173 @@
...
+	public static function setUpBeforeClass() {
+		parent::setUpBeforeClass();
+		self::$opts      = WP_Auth0_Options::Instance();
+		self::$error_log = new WP_Auth0_ErrorLog();
+	}
+
+	/**
+	 * Test that a basic create connection command requests properly.
+	 */
+	public function testThatCreateConnectionRequestsCorrectly() {
```
lbalmaceda: Odd naming. Consider the suffix you've used above "WithoutErrors". Applies to more tests on this file

tests/testApiOperations.php
```diff
@@ -0,0 +1,173 @@
...
+	 * Setup for entire test class.
+	 */
+	public static function setUpBeforeClass() {
+		parent::setUpBeforeClass();
+		self::$opts      = WP_Auth0_Options::Instance();
+		self::$error_log = new WP_Auth0_ErrorLog();
+	}
+
+	/**
+	 * Test that a basic create connection command requests properly.
```
lbalmaceda: normally the test name is enought, no need to add docblocks on tests. Though now that you added them no need to delete them, just for future reference 👍

tests/testApiOperations.php
```diff
@@ -0,0 +1,173 @@
...
+	 */
+	public function testThatCreateConnectionRequestsCorrectly() {
+		$this->startHttpHalting();
+
+		$api_ops    = new WP_Auth0_Api_Operations( self::$opts );
+		$test_token = implode( '.', [ uniqid(), uniqid(), uniqid() ] );
+
+		self::$opts->set( 'domain', 'test-wp.auth0.com' );
```
lbalmaceda: why are you changing the `opts` after using them to instantiate `api_ops` ?

tests/testApiOperations.php
```diff
@@ -0,0 +1,173 @@
...
+		self::$opts->set( 'client_id', 'TEST_CLIENT_ID' );
+
+		$caught_http = [];
+		try {
+			$api_ops->create_wordpress_connection( $test_token, false, 'good' );
+		} catch ( Exception $e ) {
+			$caught_http = unserialize( $e->getMessage() );
```
lbalmaceda: minor, but you called this `caught_request` in other tests.

tests/testApiOperations.php
```diff
@@ -0,0 +1,173 @@
...
+		parent::setUpBeforeClass();
+		self::$opts      = WP_Auth0_Options::Instance();
+		self::$error_log = new WP_Auth0_ErrorLog();
+	}
+
+	/**
+	 * Test that a basic create connection command requests properly.
+	 */
+	public function testThatCreateConnectionRequestsCorrectly() {
+		$this->startHttpHalting();
```
lbalmaceda: this sounds like a candidate for a "pre each test" method, even if not used in all the tests cases.

tests/testApiOperations.php
```diff
@@ -0,0 +1,173 @@
...
+			$caught_http = unserialize( $e->getMessage() );
+		}
+
+		$this->assertEquals( 'https://test-wp2.auth0.com/api/v2/connections', $caught_http['url'] );
+		$this->assertEquals( 'Bearer ' . $test_token, $caught_http['headers']['Authorization'] );
+		$this->assertEquals( 'DB-Test-Blog', $caught_http['body']['name'] );
+		$this->assertEquals( 'auth0', $caught_http['body']['strategy'] );
+		$this->assertContains( 'TEST_CLIENT_ID_2', $caught_http['body']['enabled_clients'] );
+
+		$this->assertEquals( true, $caught_http['body']['options']['requires_username'] );
```
lbalmaceda: consider, if possible, using `assertTrue` instead

tests/testApiOperations.php
```diff
@@ -0,0 +1,173 @@
...
+			$caught_http['body']['options']['customScripts']['login']
+		);
+
+		$this->assertContains(
+			'request.post("http://example.org/index.php?a0_action=migration-ws-get-user"',
+			$caught_http['body']['options']['customScripts']['get_user']
+		);
+
+		$this->assertContains(
+			'form:{username:email, access_token:"TEST_MIGRATION_TOKEN',
```
lbalmaceda: is this intentionally not having a `}` at the end of the string in test?

tests/testInitialSetupConsent.php
```diff
@@ -0,0 +1,291 @@
...
+	public function testThatInvalidStateRedirectsProperly() {
+		$this->startRedirectHalting();
+
+		$setup_consent = new WP_Auth0_InitialSetup_Consent( self::$opts );
+		$test_domain   = 'test-wp.auth0.com';
+		$test_token    = implode( '.', [ uniqid(), uniqid(), uniqid() ] );
+
+		$caught_redirect = [];
+		try {
+			$setup_consent->callback_with_token( $test_domain, $test_token, 'invalid_state' );
```
lbalmaceda: what's the previous context? how can you ensure that `invalid_state` is an invalid state? normally, you'd manually set that state before this call.

tests/testInitialSetupConsent.php
```diff
@@ -0,0 +1,291 @@
...
+	 */
+	public function testThatClientCreationIsAttempted() {
+		$this->startHttpHalting();
+
+		$setup_consent = new WP_Auth0_InitialSetup_Consent( self::$opts );
+		$test_token    = implode( '.', [ uniqid(), uniqid(), uniqid() ] );
+
+		$caught_http = [];
+		try {
+			$setup_consent->callback_with_token( 'test-wp.auth0.com', $test_token, 'social' );
```
lbalmaceda: how is this test logic different from the "test invalid state redirects"? Sounds like whatever caused the above test to fail should do the same here. What's changing? and if nothing is, why would an invalid_state scenario still make the create client request?

tests/testInitialSetupConsent.php
```diff
@@ -0,0 +1,291 @@
...
+		$this->http_request_type = [
+			// Successful client creation.
+			'success_create_client',
+			// Get en existing connection enabled for this client.
+			'success_get_connections',
+			// Connection updated successfully.
+			'success_update_connection',
+			// Connection created successfully.
+			'success_create_connection',
+			// Client grant failed.
```
lbalmaceda: why would this happen? Seems like this should belong in a different test case than "create new connection succeeds"

tests/traits/httpHelpers.php
```diff
@@ -62,7 +62,7 @@ public function stopHttpHalting() {
 	 * Use this at the top of tests that should test behavior for different HTTP responses.
 	 */
 	public function startHttpMocking() {
-		add_filter( 'pre_http_request', [ $this, 'httpMock' ], 1 );
+		add_filter( 'pre_http_request', [ $this, 'httpMock' ], 1, 3 );
```
lbalmaceda: what changed here?

lib/WP_Auth0_Api_Client.php
joshcanhelp: If `$payload->enabled_clients` is anything other than an array with consecutive keys (like, if one of the keys in the middle is removed) then `json_encode()` turns it into an `object` which is rejected by the server.

lib/admin/WP_Auth0_Admin_Features.php
joshcanhelp: I could but minimal changes for the patch ... I don't want to go crazy refactoring.

lib/admin/WP_Auth0_Admin_Features.php
joshcanhelp: This method returns `$input`, which contains the new values. This reverts this specific value to the value it was before the admin change. API call failed so nothing should update.

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -16,6 +16,16 @@ public function __construct( WP_Auth0_Options $a0_options ) {
...
 
+	/**
+	 * Used by both Setup Wizard installation flows.
+	 * Called in WP_Auth0_InitialSetup_ConnectionProfile::callback() when an API token is used during install.
+	 * Called in self::callback() when returning from consent URL install.
+	 *
+	 * @param string $domain - Auth0 domain for the Application.
+	 * @param string $access_token - Management API access token.
+	 * @param string $type - Installation type, "social" (AKA standard) or "enterprise".
+	 * @param bool   $hasInternetConnection - True if the installing site be reached by Auth0, false if not.
```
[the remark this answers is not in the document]
joshcanhelp: Just documenting, not refactoring

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
joshcanhelp: Public static method so no, not currently. `WP_Auth0_Api_Client` is getting refactored one method at a time so it will go away soon.

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
joshcanhelp: I understand but looking for the minimum number of changes here. This is just one of many things that should be addressed in this method.

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
joshcanhelp: Not necessary for this PR

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
joshcanhelp: Lots of better ways to do this, yes :)

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -128,19 +144,22 @@ public function consent_callback( $name ) {
...
+				if ( in_array( $client_id, $connection->enabled_clients ) ) {
 
 					if ( $db_connection_name === $connection->name ) {
-						$connection_exists     = $connection->id;
-						$connection_pwd_policy = ( isset( $connection->options ) && isset( $connection->options->passwordPolicy ) ) ? $connection->options->passwordPolicy : null;
+						$connection_exists = $connection->id;
+						if ( isset( $connection->options ) && isset( $connection->options->passwordPolicy ) ) {
+							$connection_pwd_policy = $connection->options->passwordPolicy;
+						}
 					} else {
```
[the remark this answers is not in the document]
joshcanhelp: I'll keep it as-is for now. Agree that this would be better

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
joshcanhelp: OK to continue it they fail. That method logs it internally.

tests/testAdminFeatures.php
joshcanhelp: I'm forcing it to with `$this->startHttpHalting()` above

tests/testAdminFeatures.php
joshcanhelp: Just a way to know if an exception was thrown or not

tests/testAdminFeatures.php
joshcanhelp: Mocked API call

tests/testApiOperations.php
joshcanhelp: Code sniffer beeps at me without it

tests/testApiOperations.php
joshcanhelp: Can't halt and mock at the same time though ... some improvements are possible here, for sure.

tests/testApiOperations.php
joshcanhelp: Yes, next character is a newline

tests/testInitialSetupConsent.php
joshcanhelp: There are only two values that `callback_with_token` accept here ... not actually state but that's the name used in the method.

tests/testInitialSetupConsent.php
joshcanhelp: `"social"` is one of two valid "state" values here so this proceeds to the API call

tests/testInitialSetupConsent.php
joshcanhelp: Maybe the token doesn't have the right scopes or HTTP error or any number of things.

tests/traits/httpHelpers.php
joshcanhelp: Technically nothing but that last arg is the number of params that the hooked function receives.

tests/testInitialSetupConsent.php
lbalmaceda: In that case, that could be moved to a different test other than the "happy path" of creating the connection when missing

PR #584, opened 2018-11-09
Change logged-in user redirect to login_init hook

joshcanhelp:
### Changes

PR #583 did not address all warnings that came from this change. Moving this to an existing method hooked to `login_init` and adding tests for this and all other functionality in the `login_auto` method. 

### References

#583

### Testing

### DIFF
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -372,18 +372,6 @@ class WP_Auth0 {
 			return;
 		}
 
-		// If the user has a WP session, determine where they should end up and redirect.
-		if ( is_user_logged_in() ) {
-			$login_redirect = empty( $_REQUEST['redirect_to'] ) ?
-				$this->a0_options->get( 'default_login_redirection' ) :
-				filter_var( $_REQUEST['redirect_to'], FILTER_SANITIZE_URL );
-
-			// Add a cache buster to avoid an infinite redirect loop on pages that check for auth.
-			$login_redirect = add_query_arg( time(), '', $login_redirect );
-			wp_safe_redirect( $login_redirect );
-			exit;
-		}
-
 		wp_enqueue_style( 'auth0', WPA0_PLUGIN_CSS_URL . 'login.css', false, WPA0_VERSION );
 	}
 
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -103,6 +103,27 @@ class WP_Auth0_LoginManager {
 	 */
 	public function login_auto() {
 
+		// Do not redirect anywhere if this is a logout action.
+		if ( isset( $_GET['action'] ) && 'logout' === $_GET['action'] ) {
+			return false;
+		}
+
+		// Do not redirect login page override.
+		if ( isset( $_GET['wle'] ) ) {
+			return false;
+		}
+
+		// Do not redirect non-GET requests.
+		if ( strtolower( $_SERVER['REQUEST_METHOD'] ) !== 'get' ) {
+			return false;
+		}
+
+		// Do not redirect Auth0 login processing.
+		// TODO: This should be removed as this page is no longer used for callbacks.
+		if ( null !== $this->query_vars( 'auth0' ) ) {
+			return false;
+		}
+
 		// If the user has a WP session, determine where they should end up and redirect.
 		if ( is_user_logged_in() ) {
 			$login_redirect = empty( $_REQUEST['redirect_to'] ) ?
@@ -115,21 +136,8 @@ class WP_Auth0_LoginManager {
 			exit;
 		}
 
-		if (
-			// Nothing to do.
-			( ! $this->a0_options->get( 'auto_login', false ) )
-			// Auth0 is not ready to process logins.
-			|| ! WP_Auth0::ready()
-			// Do not redirect POST requests.
-			|| strtolower( $_SERVER['REQUEST_METHOD'] ) !== 'get'
-			// Do not redirect login page override.
-			|| isset( $_GET['wle'] )
-			// Do not redirect log out action.
-			|| ( isset( $_GET['action'] ) && 'logout' === $_GET['action'] )
-			// Do not redirect Auth0 login processing.
-			// TODO: This should be removed as this page is no longer used for callbacks.
-			|| null !== $this->query_vars( 'auth0' )
-		) {
+		// Do not use the ULP if the setting is off or if the plugin is not configured.
+		if ( ! $this->a0_options->get( 'auto_login', false ) || ! WP_Auth0::ready() ) {
 			return false;
 		}
 
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -96,9 +96,25 @@ class WP_Auth0_LoginManager {
 	}
 
 	/**
-	 * Redirect to a specific connection designated in Settings > Advanced
+	 * Redirect logged-in users from wp-login.php.
+	 * Redirect to Universal Login Page under certain conditions and if the option is turned on.
+	 *
+	 * @return bool
 	 */
 	public function login_auto() {
+
+		// If the user has a WP session, determine where they should end up and redirect.
+		if ( is_user_logged_in() ) {
+			$login_redirect = empty( $_REQUEST['redirect_to'] ) ?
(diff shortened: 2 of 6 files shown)


tests/testRenderForm.php
```diff
@@ -87,63 +90,4 @@ public function testThatFormRendersWhenAuth0IsReady() {
...
-		$value = $on ? uniqid() : null;
-		self::$opts->set( 'domain', $value );
-		self::$opts->set( 'client_id', $value );
-		self::$opts->set( 'client_secret', $value );
-	}
-
-	/**
-	 * Run after each test.
-	 */
-	public function tearDown() {
```
joshcanhelp: Moved up ☝️

tests/testRenderForm.php
```diff
@@ -87,63 +90,4 @@ public function testThatFormRendersWhenAuth0IsReady() {
...
-		}
-		$this->assertTrue( $caught_exception );
-	}
-
-	/**
-	 * Set the Auth0 plugin settings.
-	 *
-	 * @param boolean $on - True to turn Auth0 on, false to turn off.
-	 */
-	public static function auth0Ready( $on = true ) {
```
joshcanhelp: Moved to `tests/traits/optionsHelpers.php` 👇

tests/testRenderForm.php
```diff
@@ -87,63 +90,4 @@ public function testThatFormRendersWhenAuth0IsReady() {
 
 		$this->assertContains( 'auth0-login-form', self::$wp_auth0->render_form( self::$html ) );
 	}
-
-	/**
-	 * Test that a specific region and domain return the correct number of IP addresses.
-	 */
-	public function testThatLoggedInUserIsRedirected() {
```
joshcanhelp: Moved to proper test suite ☝️

tests/testLoginManager.php
```diff
@@ -133,4 +173,131 @@ function ( $auth_url, $params ) {
 		);
 		$this->assertEquals( 'https://test-custom.auth0.com/authorize?test=this', $auth_url );
 	}
+
+	/**
+	 * Test that logged-in users are redirected from the wp-login.php page.
+	 */
+	public function testThatLoggedInUserIsRedirectedFromWpLogin() {
```
joshcanhelp: Moved from `tests/testRenderForm.php` 👇

tests/testLoginManager.php
```diff
@@ -133,4 +173,131 @@ function ( $auth_url, $params ) {
...
+
+		// Redirect will have a dynamic cache breaker so cannot assertEquals here.
+		$this->assertNotFalse( strpos( $caught_redirect['location'], 'http://example.org/custom?' ) );
+		$this->assertEquals( 302, $caught_redirect['status'] );
+	}
+
+	/**
+	 * Test that the ULP redirect happens.
+	 */
+	public function testUlpRedirect() {
```
joshcanhelp: Set everything up to make sure the ULP redirect happens under normal conditions.

lib/WP_Auth0_LoginManager.php
```diff
@@ -111,11 +127,10 @@ public function login_auto() {
 			// Do not redirect log out action.
 			|| ( isset( $_GET['action'] ) && 'logout' === $_GET['action'] )
 			// Do not redirect Auth0 login processing.
+			// TODO: This should be removed as this page is no longer used for callbacks.
 			|| null !== $this->query_vars( 'auth0' )
-			// Do not redirect if already authenticated.
```
joshcanhelp: Dealing with this condition in the block added above in this method ☝️

WP_Auth0.php
```diff
@@ -372,18 +372,6 @@ public function render_auth0_login_css() {
 			return;
 		}
 
-		// If the user has a WP session, determine where they should end up and redirect.
```
joshcanhelp: Meat of the PR here. This still was not early enough to precede any HTML output, leading to a PHP warning.

lib/WP_Auth0_LoginManager.php
```diff
@@ -111,11 +127,10 @@ public function login_auto() {
...
-			|| is_user_logged_in()
 		) {
-			return;
+			return false;
```
joshcanhelp: This is used in a hook call so return value is never used ... expect in tests now!

tests/testLoginManager.php
```diff
@@ -14,7 +14,50 @@
...
+	 * @var WP_Auth0_UsersRepo
+	 */
+	protected static $users_repo;
+
+	/**
+	 * Setup for entire test class.
+	 */
+	public static function setUpBeforeClass() {
+		parent::setUpBeforeClass();
+		self::$opts       = WP_Auth0_Options::Instance();
```
lbalmaceda: this is done once, but its contents are changed multiple times in different tests. I think it should be moved into a method that runs "before each test" so you can ensure the tests are fully independent of each other.

e.g. some tests set the `client_id` property, some others don't. **Independently of** if those tests are running methods that use or not that `client_id` value from the props, if not part of the context then make sure that the props object is always with the values you need for test and nothing else. By doing what I mentioned in the paragraph above you can ensure that those props get reseted on each test run. And each test should prepare their own test context/environment prior to call the "method in test".

tests/testLoginManager.php
```diff
@@ -14,7 +14,50 @@
...
+
+	/**
+	 * Run after each test.
+	 */
+	public function tearDown() {
+		parent::tearDown();
+		self::auth0Ready( false );
+		$this->stopRedirectHalting();
+		self::$error_log->clear();
+		self::$opts->set( 'custom_domain', '' );
```
lbalmaceda: Seems easier to throw this instance and create a new one than be changing manually each prop, as you could be forgetting some of them when adding new tests. Note this are tests and performance hit is not relevant here. Instantiate whatever you need the times you need 👍

tests/testLoginManager.php
```diff
@@ -133,4 +173,133 @@ function ( $auth_url, $params ) {
 		);
 		$this->assertEquals( 'https://test-custom.auth0.com/authorize?test=this', $auth_url );
 	}
+
+	/**
+	 * Test that logged-in users are redirected from the wp-login.php page.
+	 */
+	public function testThatLoggedInUserIsRedirectedFromWpLogin() {
+		$this->startRedirectHalting();
```
lbalmaceda: This is what I didn't fully understand from yesterday's response. Couldn't you call this line at the very end of the "run before each test" method? It's cleaner than having to copy-paste it on every test.

tests/testLoginManager.php
```diff
@@ -133,4 +173,133 @@ function ( $auth_url, $params ) {
...
+		}
+		$this->assertEquals( 302, $caught_redirect['status'] );
+
+		// Test that request method will stop redirect.
+		$req_method_backup         = $_SERVER['REQUEST_METHOD'];
+		$_SERVER['REQUEST_METHOD'] = 'POST';
+		$this->assertFalse( $login_manager->login_auto() );
+		$_SERVER['REQUEST_METHOD'] = $req_method_backup;
+
+		// Test that WP login override will skip the redirect.
```
lbalmaceda: All these 3 checks below should be moved into 3 different tests as the condition for the test changes.

tests/traits/optionsHelpers.php
```diff
@@ -0,0 +1,33 @@
...
+	public static $opts;
+
+	/**
+	 * Set the Auth0 plugin settings.
+	 *
+	 * @param boolean $on - True to turn Auth0 on, false to turn off.
+	 */
+	public static function auth0Ready( $on = true ) {
+		$value = $on ? uniqid() : null;
+		self::$opts->set( 'domain', $value );
```
lbalmaceda: sounds odd, even for a test, that a domain's test value is not a url shaped string but a random string.

joshcanhelp: The value is not used or checked anywhere so it's irrelevant. In this way, I can set this to something to make the plugin "active" and if I ever try to check for a specific value, it will fail (correctly).

tests/testLoginManager.php
joshcanhelp: Cleaner but not needed for every test. Some of these cannot be combined because they throw exceptions at different places.

tests/testLoginManager.php
joshcanhelp: Also ... `WP_Auth0_Options::Instance()` returns an existing object with options stored in memory.

PR #585, opened 2018-11-13
Add WooCommerce password change action.

joshcanhelp:
### Changes

Add Auth0 password change functionality to WooCommerce edit account + tests. 

### References

[Reported on the WP support forums](https://wordpress.org/support/topic/password-change-issue). 

### Testing

This PR adds unit tests. It can also be tested manually by:

1. Start with Auth0 and WooCommerce installed and configured
2. Go to the edit account page for WooCommerce (front-end of the site)
3. Attempt to change the password to something invalid; expect a visible warning, no changes made, and an error in the error log
4. Attempt to change the password to something valid; expect the change to complete with no errors

* [x] This change has been tested on the latest version of the platform

### Checklist

### DIFF
--- a/lib/profile/WP_Auth0_Profile_Change_Password.php
+++ b/lib/profile/WP_Auth0_Profile_Change_Password.php
@@ -1,4 +1,11 @@
 <?php
+/**
+ * Contains class WP_Auth0_Profile_Change_Password.
+ *
+ * @package WP-Auth0
+ *
+ * @since 3.8.0
+ */
 
 /**
  * Class WP_Auth0_Profile_Change_Password.
@@ -29,6 +36,7 @@ class WP_Auth0_Profile_Change_Password {
 	public function init() {
 		add_action( 'user_profile_update_errors', array( $this, 'validate_new_password' ), 10, 2 );
 		add_action( 'validate_password_reset', array( $this, 'validate_new_password' ), 10, 2 );
+		add_action( 'woocommerce_save_account_details_errors', array( $this, 'validate_new_password' ), 10, 2 );
 	}
 
 	/**
@@ -44,14 +52,17 @@ class WP_Auth0_Profile_Change_Password {
 	public function validate_new_password( $errors, $user ) {
 
 		// Exit if we're not changing the password.
-		if ( empty( $_POST['pass1'] ) ) {
+		// The pass1 key is for core WP, password_1 is WooCommerce.
+		if ( empty( $_POST['pass1'] ) && empty( $_POST['password_1'] ) ) {
 			return false;
 		}
-		$new_password = $_POST['pass1'];
+
+		$field_name   = ! empty( $_POST['pass1'] ) ? 'pass1' : 'password_1';
+		$new_password = $_POST[ $field_name ];
 
 		if ( isset( $_POST['user_id'] ) ) {
 			$wp_user_id = absint( $_POST['user_id'] );
-		} elseif ( is_object( $user ) && $user instanceof WP_User ) {
+		} elseif ( is_object( $user ) && ! empty( $user->ID ) ) {
 			$wp_user_id = absint( $user->ID );
 		} else {
 			return false;
@@ -83,7 +94,7 @@ class WP_Auth0_Profile_Change_Password {
 
 		// Add an error message to appear at the top of the page.
 		$error_msg = is_string( $result ) ? $result : __( 'Password could not be updated.', 'wp-auth0' );
-		$errors->add( 'auth0_password', $error_msg, array( 'form-field' => 'pass1' ) );
+		$errors->add( 'auth0_password', $error_msg, array( 'form-field' => $field_name ) );
 		return false;
 	}
 }
--- a/lib/profile/WP_Auth0_Profile_Change_Password.php
+++ b/lib/profile/WP_Auth0_Profile_Change_Password.php
@@ -34,8 +34,14 @@ class WP_Auth0_Profile_Change_Password {
 	 * @codeCoverageIgnore - Tested in TestProfileChangePassword::testInitHooks()
 	 */
 	public function init() {
+
+		// Used during profile update in wp-admin.
 		add_action( 'user_profile_update_errors', array( $this, 'validate_new_password' ), 10, 2 );
+
+		// Used during password reset on wp-login.php
 		add_action( 'validate_password_reset', array( $this, 'validate_new_password' ), 10, 2 );
+
+		// Used during WooCommerce edit account save.
 		add_action( 'woocommerce_save_account_details_errors', array( $this, 'validate_new_password' ), 10, 2 );
 	}
 
@@ -44,8 +50,8 @@ class WP_Auth0_Profile_Change_Password {
 	 * Hooked to: user_profile_update_errors, validate_password_reset
 	 * IMPORTANT: Internal callback use only, do not call this function directly!
 	 *
-	 * @param WP_Error        $errors - WP_Error object to use if validation fails.
-	 * @param boolean|WP_User $user - Boolean update or WP_User instance, depending on action.
+	 * @param WP_Error         $errors - WP_Error object to use if validation fails.
+	 * @param boolean|stdClass $user - Boolean update or WP_User instance, depending on action.
 	 *
 	 * @return boolean
 	 */
--- a/lib/profile/WP_Auth0_Profile_Change_Password.php
+++ b/lib/profile/WP_Auth0_Profile_Change_Password.php
@@ -61,8 +61,10 @@ class WP_Auth0_Profile_Change_Password {
 		$new_password = $_POST[ $field_name ];
 
 		if ( isset( $_POST['user_id'] ) ) {
+			// Input field from user edit or profile update.
 			$wp_user_id = absint( $_POST['user_id'] );
 		} elseif ( is_object( $user ) && ! empty( $user->ID ) ) {
+			// User object passed in from an action.
 			$wp_user_id = absint( $user->ID );
 		} else {
 			return false;
--- a/tests/testApiChangePassword.php
+++ b/tests/testApiChangePassword.php
@@ -123,7 +123,7 @@ class TestApiChangePassword extends TestCase {
 		$this->assertCount( 2, $log );
 		$this->assertEquals( 'caught_api_error', $log[0]['code'] );
 
-		// 4. Make sure that a weak password error returns the correct message.
+		// 3. Make sure that a weak password error returns the correct message.
 		$this->http_request_type = 'failed_weak_password';
 		$this->assertEquals(
 			'Password is too weak, please choose a different one.',
@@ -133,7 +133,7 @@ class TestApiChangePassword extends TestCase {
 		$this->assertCount( 3, $log );
 		$this->assertEquals( '400', $log[0]['code'] );
 
-		// 5. Make sure it succeeds.
+		// 4. Make sure it succeeds.
 		$this->http_request_type = 'success_empty_body';
 		$this->assertTrue( $change_password->call( uniqid(), uniqid() ) );
 		$this->assertCount( 3, self::$error_log->get() );
--- a/tests/testApiChangePassword.php
+++ b/tests/testApiChangePassword.php
@@ -58,7 +58,7 @@ class TestApiChangePassword extends TestCase {
 	}
 
 	/**
-	 * Test the request sent by the Client Credentials call.
+	 * Test the request sent by the change password call.
 	 */
 	public function testRequest() {
 		$this->startHttpHalting();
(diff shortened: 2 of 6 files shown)


lib/profile/WP_Auth0_Profile_Change_Password.php
```diff
@@ -44,14 +52,17 @@ public function init() {
...
 
 		// Exit if we're not changing the password.
-		if ( empty( $_POST['pass1'] ) ) {
+		// The pass1 key is for core WP, password_1 is WooCommerce.
+		if ( empty( $_POST['pass1'] ) && empty( $_POST['password_1'] ) ) {
 			return false;
 		}
-		$new_password = $_POST['pass1'];
+
+		$field_name   = ! empty( $_POST['pass1'] ) ? 'pass1' : 'password_1';
```
lbalmaceda: - please rename this to something more specific like `$psw_field_name` 
- what if `$_POST['password_1']` is empty?

lib/profile/WP_Auth0_Profile_Change_Password.php
```diff
@@ -44,14 +52,17 @@ public function init() {
...
+		$new_password = $_POST[ $field_name ];
 
 		if ( isset( $_POST['user_id'] ) ) {
 			$wp_user_id = absint( $_POST['user_id'] );
-		} elseif ( is_object( $user ) && $user instanceof WP_User ) {
+		} elseif ( is_object( $user ) && ! empty( $user->ID ) ) {
```
lbalmaceda: will this second condition have the same effect as the one that was here previously?

tests/testApiChangePassword.php
```diff
@@ -127,15 +127,15 @@ public function testCall() {
...
 			'Password is too weak, please choose a different one.',
-			$delete_mfa->call( uniqid(), uniqid() )
+			$change_password->call( uniqid(), uniqid() )
 		);
 		$log = self::$error_log->get();
 		$this->assertCount( 3, $log );
 		$this->assertEquals( '400', $log[0]['code'] );
 
-		// 4. Make sure it succeeds.
+		// 5. Make sure it succeeds.
```
lbalmaceda: this is 4. The one above is 3.

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
-		// Create a stub for the WP_Auth0_Api_Change_Password class.
-		$mock_api_test_password = $this
-			->getMockBuilder( WP_Auth0_Api_Change_Password::class )
-			->setMethods( [ 'call' ] )
-			->setConstructorArgs( [ self::$options, self::$api_client_creds ] )
-			->getMock();
-		$mock_api_test_password->method( 'call' )->willReturn( true, false );
-
-		$change_password = new WP_Auth0_Profile_Change_Password( $mock_api_test_password );
+		$_POST['pass1']   = uniqid();
```
lbalmaceda: add comment here explaining what `pass1` means and how is that different to `password_1` used below

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
+		$_POST['user_id'] = $user_id;
+		$this->storeAuth0Data( $user_id );
+		$this->assertTrue( $change_password->validate_new_password( $errors, false ) );
+		$this->assertEmpty( $errors->get_error_messages() );
 
-		// Call should fail because of a missing password.
+		// Test core WP password field.
```
lbalmaceda: sounds like part of another test case

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
+		unset( $_POST['pass1'] );
 		$this->assertFalse( $change_password->validate_new_password( $errors, false ) );
```
lbalmaceda: why would this fail? when separated, explain the use case on the test title or a comment

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
 
-		$_POST['pass1'] = uniqid();
+		$_POST['password_1'] = uniqid();
+		$this->assertTrue( $change_password->validate_new_password( $errors, false ) );
```
lbalmaceda: no check for errors here?

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
 
-		// Call should fail with a password because of a missing user ID.
+		// Test WooCommerce password field.
```
lbalmaceda: sounds like part of another test case

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
 
-		// Call should fail with a user object or user_id in $_POST because of no Auth0 data stored.
-		$this->assertFalse( $change_password->validate_new_password( $errors, $user_obj ) );
+		$_POST['pass1']   = uniqid();
 		$_POST['user_id'] = $user_id;
+		$this->storeAuth0Data( $user_id );
+		$this->assertTrue( $change_password->validate_new_password( $errors, false ) );
+		$this->assertEmpty( $errors->get_error_messages() );
+
+		// Test core WP profile update screen field.
```
lbalmaceda: sounds like part of another test case

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
+		unset( $_POST['user_id'] );
 		$this->assertFalse( $change_password->validate_new_password( $errors, false ) );
 
-		$this->storeAuth0Data( $user_id, 'not-auth0' );
+		// Test user object.
```
lbalmaceda: sounds like part of another test case

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
+		$errors   = new WP_Error();
+
+		$mock_api_test_password = $this->getStub( true );
+		$change_password        = new WP_Auth0_Profile_Change_Password( $mock_api_test_password );
 
-		// Call should fail with a user object or user_id in $_POST because of no Auth0 data stored.
-		$this->assertFalse( $change_password->validate_new_password( $errors, $user_obj ) );
+		$_POST['pass1']   = uniqid();
 		$_POST['user_id'] = $user_id;
+		$this->storeAuth0Data( $user_id );
```
lbalmaceda: > testThatMissingUserDataSkipsUpdate

this is not empty, or what do you mean by "missing user data"?

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
-		// Call should fail with a password because of a missing user ID.
+		// Test WooCommerce password field.
+		unset( $_POST['password_1'] );
 		$this->assertFalse( $change_password->validate_new_password( $errors, false ) );
+	}
+
+	/**
+	 * Test that empty user data will skip the password update.
+	 */
+	public function testThatMissingUserDataSkipsUpdate() {
```
lbalmaceda: what's way do you have to know if the password was or not updated?

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
+		$mock_api_test_password = $this->getStub( true );
+		$change_password        = new WP_Auth0_Profile_Change_Password( $mock_api_test_password );
 
+		$_POST['pass1']   = uniqid();
+		$_POST['user_id'] = $user_id;
 		$this->storeAuth0Data( $user_id );
+		$this->assertTrue( $change_password->validate_new_password( $errors, false ) );
+		$this->assertEmpty( $errors->get_error_messages() );
+
+		// Test that an unlinked user will not be updated.
```
lbalmaceda: sounds like part of another test case

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
+		self::$users_repo->delete_auth0_object( $user_id );
+		$this->assertFalse( $change_password->validate_new_password( $errors, false ) );
+
+		// Test that a linked, non-DB user will not be updated.
```
lbalmaceda: sounds like part of another test case

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
+	public function testThatApiFailureSetsErrorsUnsetsPassword() {
+		$user_id = $this->createUser()->ID;
+		$errors  = new WP_Error();
+
+		// API call mocked to succeed.
+		$mock_api_test_password = $this->getStub( true );
+		$change_password        = new WP_Auth0_Profile_Change_Password( $mock_api_test_password );
+
+		// Confirm that data is set to succeed.
+		$_POST['pass1']   = uniqid();
```
lbalmaceda: Please explain what these properties mean in each case either by using constants or adding comments along the test case. I can't follow the use case without those details

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
+		$_POST['pass2']   = $_POST['pass1'];
+		$_POST['user_id'] = $user_id;
+		$this->storeAuth0Data( $user_id );
 		$this->assertTrue( $change_password->validate_new_password( $errors, false ) );
+		$this->assertEmpty( $errors->get_error_messages() );
+
+		// API call mocked to fail.
```
lbalmaceda: Above you added `Confirm that data is set to succeed.` but here you're testing the fail scenario. Why, if the test says it will test only the failure case? Seems these belongs to another test case

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,135 @@ public function testInitHooks() {
...
+		$mock_api_test_password = $this->getStub( false );
+		$change_password        = new WP_Auth0_Profile_Change_Password( $mock_api_test_password );
 
-		// Call should fail on the second call with a mocked API.
 		$this->assertFalse( $change_password->validate_new_password( $errors, false ) );
 		$this->assertEquals( 'Password could not be updated.', $errors->errors['auth0_password'][0] );
 		$this->assertEquals( 'pass1', $errors->error_data['auth0_password']['form-field'] );
 		$this->assertFalse( isset( $_POST['pass1'] ) );
```
lbalmaceda: why would this fields get unset after a failure?

lib/profile/WP_Auth0_Profile_Change_Password.php
joshcanhelp: > Please rename this to something more specific like $psw_field_name

We're in a validate password method, that's a bit overkill. 

> what if $_POST['password_1'] is empty?

See lines above, if they're both empty, the method ends early.

lib/profile/WP_Auth0_Profile_Change_Password.php
joshcanhelp: No, totally different but the hook we're attaching to does not use `WP_User` and, really, all we need is the user ID.

tests/testProfileChangePassword.php
joshcanhelp: This is explained a couple lines down when the actual test happens. This is just setup and is well documented in the validate method itself.

tests/testProfileChangePassword.php
joshcanhelp: First part is just setup to show that it's succeeding. Then we remove pieces and show that it's failing. Each failure case kicks back `false` regardless of how it fails so this is the best way to make sure that what we're actually doing is causing the failure.

tests/testProfileChangePassword.php
joshcanhelp: Not technically a "failure" in that something went wrong. Just escaping early without calling the API.

tests/testProfileChangePassword.php
joshcanhelp: Actually, an error is only set if the API fails so if it returns `true`, it never got to the point where an error was set. Removing that error check from the other tests to avoid confusion.

tests/testProfileChangePassword.php
joshcanhelp: Either missing form field with a user ID or missing when passed in by WP.

tests/testProfileChangePassword.php
joshcanhelp: Can you clarify?

tests/testProfileChangePassword.php
joshcanhelp: Documented in the method itself but it halts the password change from happening when coming from core.

tests/testProfileChangePassword.php
joshcanhelp: Same setup as the others, want to show that it's succeeding, then show the specific case that's failing.

tests/testProfileChangePassword.php
joshcanhelp: This is all described in the method, I don't see any reason to duplicate those comments in the test cases:

https://github.com/auth0/wp-auth0/blob/master/lib/profile/WP_Auth0_Profile_Change_Password.php#L44

joshcanhelp:
@lbalmaceda - Responded and changes made. Thank you!

tests/testProfileChangePassword.php
joshcanhelp: Password confirmation

tests/testProfileChangePassword.php
joshcanhelp: Not clear what you're asking but ... `validate_new_password` returns `true` when the password was changed, which is contingent on the API call working.

tests/testProfileChangePassword.php
joshcanhelp: Password confirmation

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,181 @@ public function testInitHooks() {
...
+
+		// Core WP form fields sent for password update.
+		$_POST['pass1'] = $password;
+		$_POST['pass2'] = $password;
+
+		// Store userinfo for a DB strategy user.
+		$this->storeAuth0Data( $user->ID, 'auth0' );
+
+		$this->assertTrue( $change_password->validate_new_password( $errors, $user ) );
+		$this->assertEquals( $password, $_POST['pass1'] );
```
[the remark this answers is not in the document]
joshcanhelp: > can this change in anyway when validate_new_password is called?

Yes, alters the global var

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,181 @@ public function testInitHooks() {
...
+		$errors  = new WP_Error();
+
+		// Provide everything except a user record.
+		$change_password = $this->getStub( true );
+		$_POST['pass1']  = uniqid();
+		$this->storeAuth0Data( $user_id );
 
-		// Call should fail with a password because of a missing user ID.
 		$this->assertFalse( $change_password->validate_new_password( $errors, false ) );
+	}
```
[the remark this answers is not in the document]
joshcanhelp: Good to make sure in some cases. I added it where I thought it was needed.

tests/testProfileChangePassword.php
```diff
@@ -81,58 +81,181 @@ public function testInitHooks() {
...
+	 */
+	public function testThatNonAuth0UserSkipsUpdate() {
+		$user_id = $this->createUser()->ID;
+		$errors  = new WP_Error();
+
+		// Provide everything except Auth0 userinfo.
+		$change_password  = $this->getStub( true );
+		$_POST['pass1']   = uniqid();
 		$_POST['user_id'] = $user_id;
+
```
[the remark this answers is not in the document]
joshcanhelp: No

PR #595, opened 2018-12-01
Fix migration login route output and add tests

joshcanhelp:
### Changes

- Add `WP_Auth0_Options_Generic::reset()` to reset all options back to defaults
- Add dependency injection to `WP_Auth0_Routes` for the `WP_Auth0_Ip_Check` class
- Add error output for migration login route to help with troubleshooting

### Testing

* [x] This change has been tested on the latest version of the platform/language

### Checklist

* [x] All active GitHub CI checks have passed - `composer pre-commit`

### DIFF
--- a/lib/WP_Auth0_Routes.php
+++ b/lib/WP_Auth0_Routes.php
@@ -2,10 +2,25 @@
 
 class WP_Auth0_Routes {
 
+	/**
+	 * @var WP_Auth0_Options
+	 */
 	protected $a0_options;
 
-	public function __construct( WP_Auth0_Options $a0_options ) {
+	/**
+	 * @var WP_Auth0_Ip_Check
+	 */
+	protected $ip_check;
+
+	/**
+	 * WP_Auth0_Routes constructor.
+	 *
+	 * @param WP_Auth0_Options       $a0_options - WP_Auth0_Options instance.
+	 * @param WP_Auth0_Ip_Check|null $ip_check - WP_Auth0_Ip_Check instance.
+	 */
+	public function __construct( WP_Auth0_Options $a0_options, WP_Auth0_Ip_Check $ip_check = null ) {
 		$this->a0_options = $a0_options;
+		$this->ip_check   = $ip_check instanceof WP_Auth0_Ip_Check ? $ip_check : new WP_Auth0_Ip_Check( $a0_options );
 	}
 
 	public function init() {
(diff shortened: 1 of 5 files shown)


lib/WP_Auth0_Routes.php
```diff
@@ -22,11 +37,6 @@ public function setup_rewrites( $force_ws = false ) {
 
 		add_rewrite_rule( '^auth0', 'index.php?auth0=1', 'top' );
 		add_rewrite_rule( '^\.well-known/oauth2-client-configuration', 'index.php?a0_action=oauth2-config', 'top' );
-
-		// if ( $force_ws || $this->a0_options->get( 'migration_ws' ) ) {
```
cocojoe: This is no longer relevant?

lib/WP_Auth0_Routes.php
```diff
@@ -44,20 +54,32 @@ public function custom_requests( $wp ) {
...
+		$output = null;
 		if ( ! empty( $page ) ) {
 			switch ( $page ) {
 				case 'oauth2-config':
 					$this->oauth2_config();
 					exit;
 				case 'migration-ws-login':
-					$this->migration_ws_login();
-					exit;
+					$output = $this->migration_ws_login();
```
cocojoe: What's going on here that you buffer it and then either echo it or return it? Only appears to be one case that needs to do this? Is the intention to have more since you broke out the  if (output) logic later on vs handling in the case?

lib/WP_Auth0_Routes.php
joshcanhelp: @cocojoe - It's been commented out for a while, no need to keep in in there.

lib/WP_Auth0_Routes.php
joshcanhelp: All these routes handle execution and, in some cases echo out a string, some just return. I'm combining the behavior into one flow so JSON is echoed and the process exits at the same time. In case the methods need to return instead of echoing, I added that `query_var` (mostly used for testing at this point but could be helpful in a custom API route).

cocojoe: I still think the code could be tidier for this single use case. However, makes no real odds in terms of execution.

joshcanhelp: All of these routes will do the same, just wanted to keep PRs a little more streamlined 👍

PR #604, opened 2018-12-17
Remove unnecessary callback; add notice if plugin is already setup

joshcanhelp:
### Changes

- Add a notice on the Setup Wizard and remove the link from the menu if the plugin has already been configured
- Remove `wp-login.php` from callbacks (not used anymore)

### DIFF
--- a/lib/WP_Auth0_Settings_Section.php
+++ b/lib/WP_Auth0_Settings_Section.php
@@ -53,7 +53,7 @@ class WP_Auth0_Settings_Section {
 
 			add_submenu_page( $main_menu, __( 'Help', 'wp-auth0' ), __( 'Help', 'wp-auth0' ), 'manage_options', 'wpa0-help', array( $this, 'redirect_to_help' ) );
 
-			add_submenu_page( $main_menu, __( 'Auth0 for WordPress - Setup Wizard', 'wp-auth0' ), __( 'Setup Wizard', 'wp-auth0' ), 'manage_options', 'wpa0-setup', array( $this->initial_setup, 'render_setup_page' ) );
+			add_submenu_page( null, __( 'Auth0 for WordPress - Setup Wizard', 'wp-auth0' ), __( 'Setup Wizard', 'wp-auth0' ), 'manage_options', 'wpa0-setup', array( $this->initial_setup, 'render_setup_page' ) );
 		}
 
 		add_submenu_page( $main_menu, __( 'Export Users Data', 'wp-auth0' ), __( 'Export Users Data', 'wp-auth0' ), 'manage_options', 'wpa0-users-export', array( $this->users_exporter, 'render_export_users' ) );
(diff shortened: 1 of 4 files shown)


lib/WP_Auth0_Settings_Section.php
```diff
@@ -53,7 +53,7 @@ public function init_menu() {
 
 			add_submenu_page( $main_menu, __( 'Help', 'wp-auth0' ), __( 'Help', 'wp-auth0' ), 'manage_options', 'wpa0-help', array( $this, 'redirect_to_help' ) );
 
-			add_submenu_page( $main_menu, __( 'Auth0 for WordPress - Setup Wizard', 'wp-auth0' ), __( 'Setup Wizard', 'wp-auth0' ), 'manage_options', 'wpa0-setup', array( $this->initial_setup, 'render_setup_page' ) );
+			add_submenu_page( null, __( 'Auth0 for WordPress - Setup Wizard', 'wp-auth0' ), __( 'Setup Wizard', 'wp-auth0' ), 'manage_options', 'wpa0-setup', array( $this->initial_setup, 'render_setup_page' ) );
```
joshcanhelp: Removes the menu item when the plugin is already setup.

lib/initial-setup/WP_Auth0_InitialSetup_ConnectionProfile.php
```diff
@@ -11,7 +11,6 @@ public function __construct( WP_Auth0_Options $a0_options ) {
 	}
 
 	public function render( $step ) {
-		$scopes = WP_Auth0_Api_Client::GetConsentScopestoShow();
```
joshcanhelp: Moved to `templates/initial-setup/connection_profile.php`

lib/WP_Auth0_Settings_Section.php
cocojoe: But it's always `null` how can it be toggled ?

joshcanhelp: This is in an `if` block that checks if the plugin is configured or not. If it's already configured, we don't want to show the wizard.

cocojoe: ok expanded the view, so if it's not ready I can see it adds the `Setup Wizard`. 
So it must be already added via WP_Auth0_Admin.php? Hence the requirement to removal?

cocojoe:
@joshcanhelp you should modify your codecov to allow a little leeway. Take a look at https://github.com/auth0/Auth0.swift/blob/master/codecov.yml

joshcanhelp:
@cocojoe - Well, I am adding a fair bit of untested code. This isn't really the type that would benefit from a test but still, I see CodeCov's point :)

cocojoe:
@joshcanhelp  Yeah that's why you need a little tolerance in the patch/project % values, sometimes it's just not possible and I don't want to end up having to force merge everytime 😄

joshcanhelp:
@cocojoe - I can rebase and merge here without the check, just need an approval.

PR #607, opened 2018-12-20
Remove Social Amplificator functionality

joshcanhelp:
### Changes

Removes the Social Amplificator functionality and adds notes to formally deprecate classes. 

This does not introduce any breaking changes, just removes the widgets and options from the admin screens.

### DIFF
--- a/lib/WP_Auth0_Api_Client.php
+++ b/lib/WP_Auth0_Api_Client.php
@@ -305,7 +305,6 @@ class WP_Auth0_Api_Client {
 
 			'callbacks'           => array(
 				$options->get_wp_auth0_url(),
-				$options->get_wp_auth0_url(),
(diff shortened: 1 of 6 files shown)


lib/WP_Auth0_Api_Client.php
```diff
@@ -305,6 +305,7 @@ public static function create_client( $domain, $app_token, $name ) {
 
 			'callbacks'           => array(
 				$options->get_wp_auth0_url(),
+				$options->get_wp_auth0_url(),
```
lbalmaceda: why does this array contains the same url twice?

joshcanhelp: Odd ... that should not be a part of this PR. I'll remove it.

PR #617, opened 2019-01-08
Fix Migration Token Generation; Add JSON Content-Type header

joshcanhelp:
### Changes

- Add a check for existing token so it does not get regenerated when the endpoints are turned off, then on again (current behavior is to regenerate without warning, which could unintentionally cause login issues). 
- Add `Content-Type: application/json` header to endpoints that output JSON

**Note:** We added an internal task to add a control to regenerate this token in wp-admin. 

### Testing

- Adjusted existing tests to follow how this setting is handled
- Unable to add unit tests for header output (script `exit`s, making testing impossible)

### Checklist

### DIFF
--- a/lib/WP_Auth0_Routes.php
+++ b/lib/WP_Auth0_Routes.php
@@ -77,6 +77,7 @@ class WP_Auth0_Routes {
 			return false;
 		}
 
+		$json_header = true;
 		switch ( $page ) {
 			case 'oauth2-config':
 				$output = wp_json_encode( $this->oauth2_config() );
@@ -88,7 +89,8 @@ class WP_Auth0_Routes {
 				$output = wp_json_encode( $this->migration_ws_get_user() );
 				break;
 			case 'coo-fallback':
-				$output = $this->coo_fallback();
+				$json_header = false;
+				$output      = $this->coo_fallback();
 				break;
 			default:
 				return false;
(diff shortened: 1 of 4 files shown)


lib/WP_Auth0_Routes.php
```diff
@@ -98,10 +100,27 @@ public function custom_requests( $wp, $return = false ) {
...
 	}
 
+	/**
+	 * Use with the wp_headers filter to add a Content-Type header for JSON output.
+	 *
+	 * @param array $headers - Existing headers to modify.
+	 *
+	 * @return mixed
+	 */
+	public function add_json_header( array $headers ) {
```
joshcanhelp: Needs to be `public` to work with `add_filter()` call

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -612,6 +612,7 @@ public function migration_ws_validation( array $old_options, array $input ) {
 			return $input;
 		}
 
+		$input['migration_token']    = $this->options->get( 'migration_token' );
```
joshcanhelp: This setting is not typical in that it does not accept an input. The token is generated and then remains in the database (or is received from a [constant](https://auth0.com/docs/cms/wordpress/configuration#php-constant-setting-storage)).

tests/testOptionLoginRedirect.php
```diff
@@ -1,6 +1,6 @@
 <?php
```
joshcanhelp: Renaming test suite

lib/WP_Auth0_Routes.php
```diff
@@ -98,10 +100,27 @@ public function custom_requests( $wp, $return = false ) {
 			return $output;
 		}
 
+		if ( $json_header ) {
```
lbalmaceda: is this OK here or should it go before the `if ($return)` above ?? Maybe the `send_headers()` function might need to be called as well and you missed that (I don't really know)

lib/admin/WP_Auth0_Admin_Advanced.php
lbalmaceda: what if the value is missing/null? When generated I guess this is not the case. But when taking it from a constant value, are you checking this is not null in the "constants reader" logic?

joshcanhelp: If the value is empty/null, then a token gets generated. 

The constant loading populates `$this->options->get()` so that call catches both.

lib/WP_Auth0_Routes.php
joshcanhelp: No need for a header if we're just spitting back the data rather than outputting.

PR #623, opened 2019-01-25
Autoload traits and suppress log output in tests

joshcanhelp:
### Changes

Test suite changes only:

- Autoload traits and classes for test suite
- Pipe `error_log` output to `/dev/null` instead of `STDOUT`
- Minor composer script changes

### Testing

No changes to test coverage.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -5,6 +5,7 @@
  * @package WP-Auth0
  */
 
+ini_set( 'error_log', '/dev/null' );
 echo 'PHP version: ' . phpversion() . "\n";
 
 $_tests_dir = getenv( 'WP_TESTS_DIR' );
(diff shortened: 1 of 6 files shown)


tests/bootstrap.php
```diff
@@ -5,6 +5,7 @@
  * @package WP-Auth0
  */
 
+ini_set( 'error_log', '/dev/null' );
```
damieng: Does this mean users and CI can no longer report the errors?  Or do they still go somewhere useful as well?

tests/testConstantSettings.php
```diff
@@ -13,7 +13,7 @@
  */
 class TestConstantSettings extends WP_Auth0_Test_Case {
 
-	use domDocumentHelpers;
+	use DomDocumentHelpers;
```
damieng: Did an external dependency change the casing on this?

joshcanhelp: No, was having trouble with the auto-loader picking up the file.

tests/bootstrap.php
joshcanhelp: `error_log` is not tested currently ... or really very useful. It's legacy functionality that still remains in a handful of API calls. We have our own error log in the plugin that reports issues. 

In short: it was not used for testing and would just output needlessly to the console.

PR #624, opened 2019-02-04
Replace Rules-based settings with prompt to the dashboard [SDK-474]

joshcanhelp:
### Changes

There are currently several settings that attempt to create/activate/deactivate Rules in the Auth0 dashboard:

- Features > Multifactor Authentication Multifactor-Guardian-Do-Not-Rename
- Features > Fullcontact integration Enrich-profile-with-FullContact-Do-Not-Rename
- Features > Store Geolocation Store-Geo-Location-Do-Not-Rename
- Features > Store Zipcode Income Enrich-profile-with-Zipcode-Income-Do-Not-Rename
- Advanced > Link Users with Same Email Account-Linking-Do-Not-Rename

These require a valid API token or a client credentials grant (not implemented for these settings currently) and often fail because the Auth0 state is different than the WP admin one. It also makes changes for the whole tenant rather than just the WP site, which could cause issues.

This PR removes this settings for sites where the above features is currently turned off. If the feature is marked as on, there is a prompt added to manage it in the dashboard.

### DIFF
--- a/lib/admin/WP_Auth0_Admin_Advanced.php
+++ b/lib/admin/WP_Auth0_Admin_Advanced.php
@@ -80,13 +80,20 @@ class WP_Auth0_Admin_Advanced extends WP_Auth0_Admin_Generic {
 				'id'       => 'wpa0_cdn_url',
 				'function' => 'render_cdn_url',
 			),
-			array(
+		);
+
+		// TODO: Remove this once feature has been removed
+		if ( $this->options->get( 'link_auth0_users' ) ) {
+			$options[] = array(
 				'name'     => __( 'Link Users with Same Email', 'wp-auth0' ),
 				'opt'      => 'link_auth0_users',
 				'id'       => 'wpa0_link_auth0_users',
 				'function' => 'render_link_auth0_users',
-			),
-			array(
+			);
+		}
+
+		$options = $options + array(
+			( count( $options ) ) => array(
 				'name'     => __( 'Auto Provisioning', 'wp-auth0' ),
 				'opt'      => 'auto_provisioning',
 				'id'       => 'wpa0_auto_provisioning',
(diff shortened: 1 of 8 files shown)


lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -80,13 +80,20 @@ public function init() {
 				'id'       => 'wpa0_cdn_url',
 				'function' => 'render_cdn_url',
 			),
-			array(
+		);
+
+		// TODO: Remove this once feature has been removed
+		if ( $this->options->get( 'link_auth0_users' ) ) {
```
joshcanhelp: Do not show the setting if it's not currently turned on.

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -80,13 +80,20 @@ public function init() {
...
 				'opt'      => 'link_auth0_users',
 				'id'       => 'wpa0_link_auth0_users',
 				'function' => 'render_link_auth0_users',
-			),
-			array(
+			);
+		}
+
+		$options = $options + array(
+			( count( $options ) ) => array(
```
joshcanhelp: Needed to make sure the setting stays in the same place, if we're showing it.

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -294,16 +301,20 @@ public function render_cdn_url( $args = array() ) {
...
 	 * @see add_settings_field()
+	 *
+	 * @codeCoverageIgnore - To be deprecated
 	 */
 	public function render_link_auth0_users( $args = array() ) {
-		$this->render_switch( $args['label_for'], $args['opt_name'] );
 		$this->render_field_description(
-			__( 'Link accounts with the same verified e-mail address. ', 'wp-auth0' ) .
-			__( 'See the "Require Verified Email" setting above for more information on email verification', 'wp-auth0' )
+			__( 'This feature may currently be active. ', 'wp-auth0' ) .
```
joshcanhelp: "may currently be active" because this setting does not (and did not before) change if the rule was changed in the dashboard .

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -256,47 +264,56 @@ public function render_auto_login_method( $args = array() ) {
...
 	 */
 	public function render_mfa( $args = array() ) {
-		$this->render_switch( $args['label_for'], $args['opt_name'] );
 		$this->render_field_description(
-			__( 'Mark this if you want to enable multifactor authentication with Auth0 Guardian. ', 'wp-auth0' ) .
-			sprintf(
-				__( 'You can enable other MFA providers in the %s. ', 'wp-auth0' ),
-				$this->get_dashboard_link( 'multifactor' )
-			) . __( 'For more information, see our ', 'wp-auth0' ) .
+			__( 'MFA is a method of verifying identity by requiring more than 1 piece of identifying information. ', 'wp-auth0' ) .
```
joshcanhelp: We'll keep the MFA prompt here, regardless of whether the feature is on or not.

lbalmaceda: I found this particular line confusing. For me it is clearer to say "MFA is a method to verify a user's identity by checking a second factor in addition to the password". The phrases below are ok!

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -80,13 +80,20 @@ public function init() {
 				'id'       => 'wpa0_cdn_url',
 				'function' => 'render_cdn_url',
 			),
-			array(
+		);
+
+		// TODO: Remove this once feature has been removed
+		if ( $this->options->get( 'link_auth0_users' ) ) {
+			$options[] = array(
```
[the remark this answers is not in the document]
joshcanhelp: This just pushes a new item to the end of an array. Same effect as `array_push()` ([ref](https://secure.php.net/manual/en/function.array-push.php))

PR #625, opened 2019-02-16
Do not update Auth0 Application when SSO is turned on

joshcanhelp:
### Changes

- Remove Auth0 Application update from SSO validation middleware
- Add SSO on to Application created process during setup wizard
- Set Features tab defaults to `int` to match validation
- Remove other deprecated middleware

### Testing

### DIFF
--- a/lib/WP_Auth0_Api_Client.php
+++ b/lib/WP_Auth0_Api_Client.php
@@ -303,6 +303,7 @@ class WP_Auth0_Api_Client {
 			'name'                => $name,
 			'app_type'            => 'regular_web',
 
+			// Callback URLs for Auth Code and Hybrid/Implicit
 			'callbacks'           => array(
 				$options->get_wp_auth0_url(),
 			),
@@ -320,10 +321,14 @@ class WP_Auth0_Api_Client {
 				wp_login_url(),
 			),
 
+			// Advanced > Grant Types
 			'grant_types'         => self::get_client_grant_types(),
 			'jwt_configuration'   => array(
 				'alg' => self::DEFAULT_CLIENT_ALG,
 			),
+
+			// "Use Auth0 to do Single Sign On"
+			'sso'                 => true,
 		);
 
 		$response = wp_remote_post(
--- a/lib/WP_Auth0_Api_Client.php
+++ b/lib/WP_Auth0_Api_Client.php
@@ -323,6 +323,8 @@ class WP_Auth0_Api_Client {
 
 			// Advanced > Grant Types
 			'grant_types'         => self::get_client_grant_types(),
+
+			// Advanced > OAuth > JsonWebToken Signature Algorithm
 			'jwt_configuration'   => array(
 				'alg' => self::DEFAULT_CLIENT_ALG,
 			),
--- a/lib/WP_Auth0_Options.php
+++ b/lib/WP_Auth0_Options.php
@@ -207,9 +207,9 @@ class WP_Auth0_Options extends WP_Auth0_Options_Generic {
 
 			// Features
 			'password_policy'           => 'fair',
-			'sso'                       => false,
-			'singlelogout'              => false,
-			'override_wp_avatars'       => true,
+			'sso'                       => 0,
+			'singlelogout'              => 0,
+			'override_wp_avatars'       => 1,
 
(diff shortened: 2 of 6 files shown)


lib/WP_Auth0_Api_Client.php
```diff
@@ -320,10 +321,16 @@ public static function create_client( $domain, $app_token, $name ) {
...
+			// Advanced > Grant Types
 			'grant_types'         => self::get_client_grant_types(),
+
+			// Advanced > OAuth > JsonWebToken Signature Algorithm
 			'jwt_configuration'   => array(
 				'alg' => self::DEFAULT_CLIENT_ALG,
 			),
+
+			// "Use Auth0 to do Single Sign On"
+			'sso'                 => true,
```
joshcanhelp: This option is default on for new applications so just making sure it's on for admins who want to use SSO in the plugin.

lib/WP_Auth0_Options.php
```diff
@@ -207,9 +207,9 @@ protected function defaults() {
 
 			// Features
 			'password_policy'           => 'fair',
-			'sso'                       => false,
-			'singlelogout'              => false,
-			'override_wp_avatars'       => true,
+			'sso'                       => 0,
```
joshcanhelp: The settings are validated and checked for `int` 1 or 0. This adjusts the default value (used when the plugin is first installed) to use the same.

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -12,7 +12,6 @@ class WP_Auth0_Admin_Advanced extends WP_Auth0_Admin_Generic {
 	protected $actions_middlewares = array(
 		'basic_validation',
 		'migration_ws_validation',
-		'link_accounts_validation',
```
joshcanhelp: Validation middleware left over from previous commit (#624)

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -11,12 +11,7 @@ class WP_Auth0_Admin_Features extends WP_Auth0_Admin_Generic {
 
 	protected $actions_middlewares = array(
 		'basic_validation',
-		'georule_validation',
```
joshcanhelp: Deprecated SSO validation middleware and others left over from previous commit (#624)

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -378,21 +373,28 @@ public function render_override_wp_avatars( $args = array() ) {
 	}
 
 	public function basic_validation( $old_options, $input ) {
-		// SLO will be turned off in WP_Auth0_Admin_Features::sso_validation() if SSO is not on.
-		$input['singlelogout']        = ( isset( $input['singlelogout'] ) ? $input['singlelogout'] : 0 );
-		$input['override_wp_avatars'] = ( isset( $input['override_wp_avatars'] ) ? $input['override_wp_avatars'] : 0 );
+		$input['sso'] = ! empty( $input['sso'] ) ? 1 : 0;
```
joshcanhelp: Adding in validation from removed `sso_validation` middleware.

lib/admin/WP_Auth0_Admin_Features.php
```diff
@@ -378,21 +373,28 @@ public function render_override_wp_avatars( $args = array() ) {
...
+
+		// Turn SLO off if SSO is off.
+		$input['singlelogout']        = ! empty( $input['singlelogout'] ) && $input['sso'] ? 1 : 0;
```
joshcanhelp: Using `empty` here because that array key may not be set. Not using `empty` for `$input['sso']` because that is always normalized above.

lib/WP_Auth0_Options.php
damieng: Maybe add a comment to say 0=false, 1=true or is it well known in the PHP world?

lib/admin/WP_Auth0_Admin_Features.php
damieng: Maybe flip the inversion and results to simplify?   `empty( $input['sso'] ) ? 0 : 1;`

PR #632, opened 2019-03-14
Better Management API token handling

### LINKED ISSUE #631: OpenSSL error in WP_Auth0_Api_Client_Credentials::call
### Description

The error happens, if a user tries to change his password.
Since there is no valid api_token, Wordpress calls `WP_Auth0_Api_Client_Credentials::call()`.
The JWT::decode() fails to verify the signature:
```
[14-Mar-2019 10:26:40 UTC] PHP Warning:  openssl_verify(): supplied key param cannot be coerced into a public key in /srv/www/app/wp-content/plugins/auth0/lib/php-jwt/JWT.php on line 223
```
The DomainException is thrown:
```
DomainException: OpenSSL error: error:0906D06C:PEM routines:PEM_read_bio:no start line
```
and written to the Auth0 Log.

The line 80 in `php-jwt/JWT.php`:
```
if ( false === ( $sig = static::urlsafeB64Decode( $cryptob64 ) ) ) {
```
seems to result in a broken `$sig`.

I checked the token returned from the Auth0 Management API with https://jwt.io/ and the signature is verified successfully.

What could be the problem?

### Environment

Please provide the following:

* WP-Auth0 version: 3.9.0
* WordPress version: 4.9.9
* PHP version: 7.0

The Client Secret is not base64 encoded.
The JsonWebToken Signature Algorithm of the client is set to HS256 in the Auth0 management console and also in the Wordpress Auth0 settings.

joshcanhelp:
### Changes

- Remove Management API access token validation 
- Store new API tokens and scope in transients

### References

Closes #631

### Testing

- [Failing tests commit](https://github.com/auth0/wp-auth0/pull/632/commits/56f9c236a26eca97eb003253a191da03790750d5)
- [More failing tests commit](https://github.com/auth0/wp-auth0/pull/632/commits/c1b56f675cc9da00bf9827fc45d289509ab017bd)
- [Functionality changes commit](https://github.com/auth0/wp-auth0/pull/632/commits/61533780e7be345fe18381cfe617b25d8f90d551)
- [Passing tests commit](https://github.com/auth0/wp-auth0/pull/632/commits/c3ba87abcfc54b1d91afa49af6ee2a25c7affb67)

* [x] This change has been tested on the latest version of WP

### Checklist

### DIFF
--- a/lib/api/WP_Auth0_Api_Abstract.php
+++ b/lib/api/WP_Auth0_Api_Abstract.php
@@ -54,20 +54,6 @@ abstract class WP_Auth0_Api_Abstract {
 	protected $api_client_creds;
 
 	/**
-	 * API token from plugin settings or Client Credentials call.
-	 *
-	 * @var string
-	 */
-	protected $api_token;
-
-	/**
-	 * Decoded API token from plugin settings.
-	 *
-	 * @var object
-	 */
-	protected $api_token_decoded;
-
-	/**
 	 * API path.
 	 *
 	 * @var string
@@ -180,47 +166,35 @@ abstract class WP_Auth0_Api_Abstract {
 	 */
 	protected function set_bearer( $scope ) {
 
-		$this->api_token = wp_cache_get( self::CACHE_KEY, WPA0_CACHE_GROUP );
-		if ( ! $this->api_token ) {
-			$this->api_token = $this->options->get( 'auth0_app_token' ); // REMOVE
-		}
+		$api_token = WP_Auth0_Api_Client_Credentials::get_stored_token();
 
-		if ( $this->api_token ) {
-			try {
-				$this->api_token_decoded = $this->decode_jwt( $this->api_token );
-			} catch ( Exception $e ) {
-				// If we can't decode the token, try a client credentials grant below.
-				$this->api_token = null;
-			}
-		}
-
-		// Could not decode the stored API token or none was found so try to get one via API.
-		if ( ! $this->api_token_decoded && $this->api_client_creds instanceof WP_Auth0_Api_Client_Credentials ) {
-			$this->api_token         = $this->api_client_creds->call();
-			$this->api_token_decoded = $this->api_client_creds->get_token_decoded();
+		// No stored API token so need to get a new one.
+		if ( ! $api_token && $this->api_client_creds instanceof WP_Auth0_Api_Client_Credentials ) {
+			$api_token = $this->api_client_creds->call();
 		}
 
 		// No token to use, error recorded in previous steps.
-		if ( ! $this->api_token_decoded ) {
-			wp_cache_delete( self::CACHE_KEY, WPA0_CACHE_GROUP );
+		if ( ! $api_token ) {
 			return false;
+		}
+
+		if ( WP_Auth0_Api_Client_Credentials::check_stored_scope( $scope ) ) {
+			// Scope exists, add to the header and cache.
+			$this->add_header( 'Authorization', 'Bearer ' . $api_token );
+			return true;
 		}
 
 		// API token is missing the required scope.
-		if ( ! $this->api_token_has_scope( $scope ) ) {
-			WP_Auth0_ErrorManager::insert_auth0_error(
-				__METHOD__,
-				// translators: The $scope var here is a machine term and should not be translated.
-				sprintf( __( 'API token does not include the scope %s.', 'wp-auth0' ), $scope )
-			);
-			wp_cache_delete( self::CACHE_KEY, WPA0_CACHE_GROUP );
-			return false;
-		}
+		WP_Auth0_ErrorManager::insert_auth0_error(
+			__METHOD__,
+			// translators: The $scope var here is a machine term and should not be translated.
+			sprintf( __( 'API token does not include the scope %s.', 'wp-auth0' ), $scope )
+		);
 
-		// Scope exists, add to the header and cache.
-		$this->add_header( 'Authorization', 'Bearer ' . $this->api_token );
-		wp_cache_set( self::CACHE_KEY, $this->api_token, WPA0_CACHE_GROUP );
-		return true;
+		// Delete the stored token so we can try again.
+		WP_Auth0_Api_Client_Credentials::delete_store();
+		return false;
+
 	}
 
 	/**
@@ -383,7 +357,9 @@ abstract class WP_Auth0_Api_Abstract {
 	}
 
 	/**
-	 * Decode an Auth0 Management API token.
+	 * Decode an RS256 Auth0 Management API token.
+	 *
+	 * TODO: Deprecate
 	 *
 	 * @param string $token - API JWT to decode.
 	 *
@@ -395,12 +371,13 @@ abstract class WP_Auth0_Api_Abstract {
 	 * @throws BeforeValidException         Provided JWT used before it's eligible as defined by 'nbf'.
 	 * @throws BeforeValidException         Provided JWT used before it's been created as defined by 'iat'.
 	 * @throws ExpiredException             Provided JWT has since expired, as defined by the 'exp' claim.
+	 *
+	 * @codeCoverageIgnore - To be deprecated.
 	 */
 	protected function decode_jwt( $token ) {
 		return JWT::decode(
 			$token,
-			$this->options->get_client_secret_as_key(),
-			// Management API tokens are always RS256.
+			WP_Auth0_Api_Client::JWKfetch( $this->domain ),
 			array( 'RS256' )
 		);
 	}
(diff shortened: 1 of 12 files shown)


lib/api/WP_Auth0_Api_Abstract.php
```diff
@@ -180,47 +182,37 @@ protected function set_path( $path ) {
 	 */
 	protected function set_bearer( $scope ) {
 
-		$this->api_token = wp_cache_get( self::CACHE_KEY, WPA0_CACHE_GROUP );
-		if ( ! $this->api_token ) {
-			$this->api_token = $this->options->get( 'auth0_app_token' ); // REMOVE
-		}
+		$api_token = WP_Auth0_Api_Client_Credentials::get_stored_token();
```
lbalmaceda: What's the reason to have this method static and the `call()` one part of the instance instead?

lib/api/WP_Auth0_Api_Abstract.php
```diff
@@ -180,47 +182,37 @@ protected function set_path( $path ) {
...
-			$this->api_token_decoded = $this->api_client_creds->get_token_decoded();
+		// No stored API token so need to get a new one.
+		if ( ! $api_token && $this->api_client_creds instanceof WP_Auth0_Api_Client_Credentials ) {
+			$api_token = $this->api_client_creds->call();
 		}
 
 		// No token to use, error recorded in previous steps.
-		if ( ! $this->api_token_decoded ) {
-			wp_cache_delete( self::CACHE_KEY, WPA0_CACHE_GROUP );
+		if ( ! $api_token ) {
```
lbalmaceda: shouldn't this case be covered by a exception thrown in `call()` if the request fails somehow?

lib/api/WP_Auth0_Api_Abstract.php
```diff
@@ -180,47 +182,37 @@ protected function set_path( $path ) {
...
 			return false;
 		}
 
+		if ( WP_Auth0_Api_Client_Credentials::check_stored_scope( $scope ) ) {
```
lbalmaceda: What is this line doing?

lib/api/WP_Auth0_Api_Client_Credentials.php
```diff
@@ -77,17 +89,50 @@ protected function handle_response( $method ) {
...
+	 * Check the stored API token scope for a specific scope.
+	 *
+	 * @param string $scope - Scope to check for.
+	 *
+	 * @return bool
+	 */
+	public static function check_stored_scope( $scope ) {
+		$stored_scope = get_transient( self::SCOPE_TRANSIENT_KEY );
+		$scopes       = explode( ' ', $stored_scope );
+		return ! empty( $scopes ) && in_array( $scope, $scopes );
```
lbalmaceda: this `! empty( $scopes )` is redundant. You're already looking for a match with `in_array`. 
Also I don't see you _exploding_ the `$scope` parameter. What if the value has multiple scopes separated with spaces? Adding a tests scenario for this will help you fix it

tests/testApiChangeEmail.php
```diff
@@ -129,6 +138,65 @@ public function testThatSuccessfulApiCallReturnsTrue() {
...
+	 * Test that the bearer setting succeeds if there is a token stored with the correct scope.
+	 */
+	public function testThatSetBearerSucceedsWithStoredToken() {
+		$this->startHttpMocking();
+
+		set_transient( 'auth0_api_token', uniqid() );
+		set_transient( 'auth0_api_token_scope', 'update:users' );
+
+		$this->http_request_type = 'success_empty_body';
+		$api                     = new WP_Auth0_Api_Change_Email( self::$opts, self::$api_client_creds );
```
lbalmaceda: "Test that the bearer setting succeeds". Where is this setting being called here? Maybe you are not missing the call but the name of the test method should be different and not disclose "black box" internals.

tests/testProfileChangePassword.php
```diff
@@ -20,7 +20,7 @@ class TestProfileChangePassword extends WP_Auth0_Test_Case {
 
 	use UsersHelper;
 
-	use httpHelpers;
+	use HttpHelpers;
```
lbalmaceda: You're changing the case of the class name but the file name is still the same

lib/api/WP_Auth0_Api_Abstract.php
joshcanhelp: `call()` is part of an abstract so it needs to be defined. This can accept parameters required for the API call. In this case, no parameters are needed but other child classes use that. 

The static methods don't require any parameters and act on global state (database).

lib/api/WP_Auth0_Api_Abstract.php
joshcanhelp: No exceptions thrown in `call()` or else they would have to be caught. It just returns bool whether it worked or not.

lib/api/WP_Auth0_Api_Abstract.php
joshcanhelp: Checks that the scope we have stored (from the CC grant call) matches the scope we need.

lib/api/WP_Auth0_Api_Client_Credentials.php
joshcanhelp: > this `! empty( $scopes )` is redundant

You're right, `explode()` always returns an array even if the "string" is `null`/`false`. I'll change that. 

> Also I don't see you exploding the $scope parameter. What if the value has multiple scopes separated with spaces? Adding a tests scenario for this will help you fix it

There is a test for multiple scopes. The `$scope` parameter only excepts one (we only ever need one for the API call [at least the ones we have currently])

tests/testApiChangeEmail.php
joshcanhelp: This is basically a generic test against `set_bearer` instead of reflecting/exposing. This class requires the bearer token to be set so I'm making sure that it is and checking that it was stored. 

> Where is this setting being called here?

I'm not clear on your question.

tests/testProfileChangePassword.php
joshcanhelp: File name is fine, will get picked up by the autoloader. Testing individual files failed when the case of the trait didn't match the case of the `use` statement.

tests/testApiChangeEmail.php
lbalmaceda: Not a question but rather an obversation. That the name of the test suggest you're going to call `set_bearer` but in the test you don't. It's true that you might be calling it internally when a new instance of `WP_Auth0_Api_Change_Email` is created, but the test should not know what methods are called internally. The test should only know that a new instance of `WP_Auth0_Api_Change_Email ` is being created. That's why I mentioned to only test the "black box" of the method and not its internals.

lib/api/WP_Auth0_Api_Abstract.php
```diff
@@ -180,47 +182,37 @@ protected function set_path( $path ) {
...
-			}
-		}
-
-		// Could not decode the stored API token or none was found so try to get one via API.
-		if ( ! $this->api_token_decoded && $this->api_client_creds instanceof WP_Auth0_Api_Client_Credentials ) {
-			$this->api_token         = $this->api_client_creds->call();
-			$this->api_token_decoded = $this->api_client_creds->get_token_decoded();
+		// No stored API token so need to get a new one.
+		if ( ! $api_token && $this->api_client_creds instanceof WP_Auth0_Api_Client_Credentials ) {
+			$api_token = $this->api_client_creds->call();
```
lbalmaceda: is this call aware of the required `$scope`?

lib/api/WP_Auth0_Api_Abstract.php
```diff
@@ -180,47 +182,37 @@ protected function set_path( $path ) {
...
-		}
+			)
+		);
 
-		// Scope exists, add to the header and cache.
-		$this->add_header( 'Authorization', 'Bearer ' . $this->api_token );
-		wp_cache_set( self::CACHE_KEY, $this->api_token, WPA0_CACHE_GROUP );
-		return true;
+		// Delete the stored token so we can try again.
+		WP_Auth0_Api_Client_Credentials::delete_store();
```
lbalmaceda: why would you need to delete the token? What if the user wants to use the stored token for a different API endpoint call?

lib/api/WP_Auth0_Api_Client_Credentials.php
```diff
@@ -77,17 +89,50 @@ protected function handle_response( $method ) {
...
+	 */
+	public static function delete_store() {
+		delete_transient( self::TOKEN_TRANSIENT_KEY );
+		delete_transient( self::SCOPE_TRANSIENT_KEY );
+	}
+
+	/**
+	 * Check the stored API token scope for a specific scope.
+	 *
+	 * @param string $scope - Scope to check for.
```
lbalmaceda: maybe clarify here that the scope param is for a single scope vs a space separated string of scopes. This was confusing to me on the initial review because when we build /authorize URLs we have a "scope" parameter that accepts as many scopes as wanted, separated by a space.

lib/api/WP_Auth0_Api_Abstract.php
joshcanhelp: No, it doesn't need to be. You get the scopes back that the Application is authorized for.

lib/api/WP_Auth0_Api_Abstract.php
joshcanhelp: Then it tries again. We don't want the token stored without required scopes or else we can't try again.

lib/api/WP_Auth0_Api_Client_Credentials.php
joshcanhelp: This is only for the client credentials grant (hence the class name)

PR #637, opened 2019-03-22
Remove automatic client grant creation

joshcanhelp:
### Changes

In version 3.5.0 and 3.5.1, a Client Credentials grant was automatically created between the WP site and the Management API to retrieve user data. This could only happen if a valid API access token was already saved in the setting page. This grant is no longer required for the plugin to function properly and uncommon for sites that have been live for longer than the token expiration (24 hours be default). 

The number of sites that could use this automatic update was small to begin with and the number of sites that are running < 3.5.x are down to 8% ([plugin stats](https://wordpress.org/plugins/auth0/advanced/)).

This PR removes the automatic check and attempted fix in preparation for removing the admin-stored API token. Sites that update from < 3.5 will still receive user data via the ID token and updates that did not function before the update (user email and password updates, email verification resending) will continue to not function but log a clear error message. 

This PR also deprecates the methods that would display an admin message if the process failed. 

### Testing

### Checklist

### DIFF
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -441,11 +441,6 @@ class WP_Auth0 {
 		delete_option( 'widget_wp_auth0_widget' );
 		delete_option( 'widget_wp_auth0_social_amplification_widget' );
 
-		delete_option( 'wp_auth0_client_grant_failed' );
-		delete_option( 'wp_auth0_client_grant_success' );
-		delete_option( 'wp_auth0_grant_types_failed' );
-		delete_option( 'wp_auth0_grant_types_success' );
-
 		delete_transient( WPA0_JWKS_CACHE_TRANSIENT_NAME );
 	}
 
--- a/lib/WP_Auth0_DBManager.php
+++ b/lib/WP_Auth0_DBManager.php
@@ -16,9 +16,6 @@ class WP_Auth0_DBManager {
 		}
 
 		add_action( 'plugins_loaded', array( $this, 'check_update' ) );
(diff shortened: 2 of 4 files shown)


WP_Auth0.php
```diff
@@ -441,11 +441,6 @@ public static function uninstall() {
 		delete_option( 'widget_wp_auth0_widget' );
 		delete_option( 'widget_wp_auth0_social_amplification_widget' );
 
-		delete_option( 'wp_auth0_client_grant_failed' );
```
joshcanhelp: Options are removed upon plugin update so no need to remove when the plugin is uninstalled.

lib/WP_Auth0_DBManager.php
```diff
@@ -102,128 +98,13 @@ public function install_db( $version_to_install = null, $app_token = '' ) {
 			}
 		}
 
-		// App token needed for following updates
```
joshcanhelp: API token is no longer used during the migration process.

lib/admin/WP_Auth0_Admin_Basic.php
```diff
@@ -335,19 +335,6 @@ public function basic_validation( $old_options, $input ) {
 			? $input['auth0_app_token'] // TO BE DEPRECATED
 			: $old_options['auth0_app_token'] ); // TO BE DEPRECATED
 
-		// If we have an app token, get and store the audience
```
joshcanhelp: Removing the automatic update attempt when a new API token is saved.

PR #639, opened 2019-03-25
Do not store the API token on install

joshcanhelp:
### Changes

- Change the Setup Wizard to no longer store the API token used in preparation for removing the App Token setting. 
- Add TODO comments for deprecations before release.

### References

Continues the work from #637 and #632 to remove the App Token field (use a Client Credentials grant for Management API work and store in a expiring transient).

### Testing

This can be tested manually by installing the plugin from scratch on a new site. Expect no changes to behavior.

### Checklist

### DIFF
--- a/lib/WP_Auth0_Api_Client.php
+++ b/lib/WP_Auth0_Api_Client.php
@@ -175,6 +175,7 @@ class WP_Auth0_Api_Client {
 	}
 
 	/**
+	 * TODO: Deprecate
 	 *
 	 * @param string $domain - tenant domain
 	 * @param string $access_token - access token with at least `openid` scope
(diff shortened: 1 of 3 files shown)


lib/WP_Auth0_Api_Client.php
```diff
@@ -175,6 +175,7 @@ public static function get_client_token() {
 	}
```
joshcanhelp: Comments only in this file.

damieng:
Do we need to clean this token up on existing installs or did it never ship?

joshcanhelp:
@damieng - When the field is removed, I'll delete the stored one as well. For 99%+ installs, the token is already expired.

PR #651, opened 2019-03-29
Remove auto-logout; deprecation notice for SSO setting

joshcanhelp:
### Changes

- Remove auto-logout check when "Single Logout" is turned on. This was causing many, many API calls for dubious utility and prone to confusing behavior.
- Add Auth0 logout redirect when "Single Logout" is turned on.
- Add a deprecation notice to "Single Sign-On" referencing the ULP setting. This setting still works as it did before (`checkSession` call on `wp-login.php`) but will be removed in the next major.

### References

[Docs PR commit](https://github.com/auth0/docs/pull/7384/commits/264c8994e3a64902204a07bb364749bd98d64bdf)


WP_Auth0.php
```diff
@@ -626,17 +626,20 @@ function get_auth0_curatedBlogName() {
...
 /**
- * @param array $classes
+ * Add new classes to the body element on all front-end and login pages.
+ *
+ * @param array $classes - Array of existing classes.
+ *
  * @return array
  */
-function wp_auth0_hook_add_body_class( array $classes ) {
+function wp_auth0_filter_body_class( array $classes ) {
```
joshcanhelp: Public global function but only added a couple of days ago and not released.

PR #652, opened 2019-03-30
Deprecate WP_Auth0_UsersRepo methods

joshcanhelp:
### Changes

This PR deprecates the following methods:

- `WP_Auth0_UsersRepo::init()` - still used but will be removed in the next major.
- `WP_Auth0_UsersRepo::getUser()` - no longer used
- `WP_Auth0_UsersRepo::tokenHasRequiredScopes()` - no longer used

### DIFF
--- a/lib/WP_Auth0_Configure_JWTAUTH.php
+++ b/lib/WP_Auth0_Configure_JWTAUTH.php
@@ -3,14 +3,14 @@
 /**
  * Class WP_Auth0_Configure_JWTAUTH
  *
- * @deprecated 3.10.0, plugin is deprecated and removed from the WP plugin repo.
+ * @deprecated - 3.10.0, plugin is deprecated and removed from the WP plugin repo.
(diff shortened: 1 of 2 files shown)


lib/WP_Auth0_Configure_JWTAUTH.php
```diff
@@ -3,14 +3,14 @@
 /**
  * Class WP_Auth0_Configure_JWTAUTH
  *
- * @deprecated 3.10.0, plugin is deprecated and removed from the WP plugin repo.
+ * @deprecated - 3.10.0, plugin is deprecated and removed from the WP plugin repo.
```
cocojoe: Is the `-` a formatting requirement ? I'll approve anyway, just curious.

joshcanhelp: No, just inconsistent with the rest

PR #661, opened 2019-04-01
Deprecations for init functions calling add_action or add_filter

joshcanhelp:
### Changes

Deprecates `init()` methods that call `add_action` or `add_filter`. These will be moved to global functions in the next major release to allow un-hooking when needed.

### DIFF
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -78,16 +78,17 @@ class WP_Auth0 {
 	public function __construct( $options = null ) {
 		spl_autoload_register( array( $this, 'autoloader' ) );
 		$this->a0_options = $options instanceof WP_Auth0_Options ? $options : WP_Auth0_Options::Instance();
+		$this->basename   = plugin_basename( __FILE__ );
 	}
 
 	/**
-	 * Initialize the plugin and its modules setting all the hooks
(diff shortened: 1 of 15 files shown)


WP_Auth0.php
```diff
@@ -78,16 +78,17 @@ class WP_Auth0 {
...
+	 *
+	 * @deprecated - 3.10.0, will move add_action calls out of this class in the next major.
+	 *
+	 * @codeCoverageIgnore - Deprecated.
 	 */
 	public function init() {
-		$this->basename = plugin_basename( __FILE__ );
-
-		$ip_checker = new WP_Auth0_Ip_Check();
-		$ip_checker->init();
```
joshcanhelp: Setting was already removed so this call was doing noting.

WP_Auth0.php
```diff
@@ -119,10 +120,6 @@ public function init() {
 
 		add_filter( 'plugin_action_links_' . $this->basename, array( $this, 'wp_add_plugin_settings_link' ) );
 
-		if ( isset( $_GET['message'] ) ) {
```
joshcanhelp: Very old functionality, not used.

PR #665, opened 2019-04-08
Add wp_auth0_get_option function to get option values

joshcanhelp:
### Changes

Add `wp_auth0_get_option()` function for getting option values. This will make it easier for plugins/themes that extend the plugin. 

### Testing

### DIFF
--- a/WP_Auth0.php
+++ b/WP_Auth0.php
@@ -29,6 +29,8 @@ define( 'WPA0_JWKS_CACHE_TRANSIENT_NAME', 'WP_Auth0_JWKS_cache' );
 
 define( 'WPA0_LANG', 'wp-auth0' ); // deprecated; do not use for translations
 
+require_once 'functions.php';
+
 /*
  * Localization
  */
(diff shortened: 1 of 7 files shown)


WP_Auth0.php
```diff
@@ -571,53 +573,6 @@ public static function get_plugin_dir_url() {
 $a0_plugin = new WP_Auth0();
 $a0_plugin->init();
 
-if ( ! function_exists( 'get_auth0userinfo' ) ) {
```
joshcanhelp: Moved as-is to `pluggable.php` below

pluggable.php
```diff
@@ -0,0 +1,71 @@
...
+	 * Return a stored option value.
+	 *
+	 * @since 3.10.0
+	 *
+	 * @param string $key - Settings key to get.
+	 * @param mixed  $default - Default value to return if not found.
+	 *
+	 * @return mixed
+	 */
+	function wp_auth0_get_option( $key, $default = null ) {
```
joshcanhelp: Only new function is here.

PR #670, opened 2019-04-11
Settings page adjustments

joshcanhelp:
### Changes

- Reorder Embedded options
- Minor re-wording on 2 other options

### DIFF
--- a/lib/admin/WP_Auth0_Admin_Basic.php
+++ b/lib/admin/WP_Auth0_Admin_Basic.php
@@ -73,7 +73,7 @@ class WP_Auth0_Admin_Basic extends WP_Auth0_Admin_Generic {
 				'function' => 'render_client_signing_algorithm',
 			),
 			array(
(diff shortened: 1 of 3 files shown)


lib/admin/WP_Auth0_Admin_Basic.php
```diff
@@ -73,7 +73,7 @@ public function init() {
 				'function' => 'render_client_signing_algorithm',
 			),
 			array(
-				'name'     => __( 'Cache Time (in minutes)', 'wp-auth0' ),
+				'name'     => __( 'JWKS Cache Time (in minutes)', 'wp-auth0' ),
```
damieng: Should this be a lower-case S?

damieng: And also are they going to know what JWK is?

joshcanhelp: Not officially:
https://auth0.com/docs/jwks

> And also are they going to know what JWK is?

If they don't, they shouldn't be screwing around with it 😃

PR #671, opened 2019-04-16
Fix options overwriting with defaults and removing constant values on update

joshcanhelp:
### Changes

This PR fixes 2 issues that surface during database migrations. 

- Fix updating options removes constant-set values from memory
- Fix getting options replacing empty values with defaults
- Update database migrations to batch-save new options
- Add `remove()` method to remove option keys from memory and database 

### Testing

**See commit log for commits with failing tests.**

* [x] This change has been tested on WP 5.2-beta

### Checklist

### DIFF
--- a/lib/WP_Auth0_DBManager.php
+++ b/lib/WP_Auth0_DBManager.php
@@ -149,22 +149,22 @@ class WP_Auth0_DBManager {
 			}
 
 			// Nullify and delete all removed options.
-			$options->remove( 'auth0js-cdn', false );
-			$options->remove( 'passwordless_cdn_url', false );
-			$options->remove( 'cdn_url_legacy', false );
+			$options->remove( 'auth0js-cdn' );
+			$options->remove( 'passwordless_cdn_url' );
+			$options->remove( 'cdn_url_legacy' );
 
-			$options->remove( 'social_twitter_key', false );
-			$options->remove( 'social_twitter_secret', false );
-			$options->remove( 'social_facebook_key', false );
-			$options->remove( 'social_facebook_secret', false );
-			$options->remove( 'connections', false );
+			$options->remove( 'social_twitter_key' );
+			$options->remove( 'social_twitter_secret' );
+			$options->remove( 'social_facebook_key' );
+			$options->remove( 'social_facebook_secret' );
+			$options->remove( 'connections' );
 
-			$options->remove( 'chart_idp_type', false );
-			$options->remove( 'chart_gender_type', false );
-			$options->remove( 'chart_age_type', false );
-			$options->remove( 'chart_age_from', false );
-			$options->remove( 'chart_age_to', false );
-			$options->remove( 'chart_age_step', false );
+			$options->remove( 'chart_idp_type' );
+			$options->remove( 'chart_gender_type' );
+			$options->remove( 'chart_age_type' );
+			$options->remove( 'chart_age_from' );
+			$options->remove( 'chart_age_to' );
+			$options->remove( 'chart_age_step' );
 
 			// Migrate WLE setting
 			$new_wle_value = $options->get( 'wordpress_login_enabled' ) ? 'link' : 'isset';
--- a/lib/WP_Auth0_DBManager.php
+++ b/lib/WP_Auth0_DBManager.php
@@ -47,9 +47,9 @@ class WP_Auth0_DBManager {
 			if ( ! empty( $dict ) ) {
 
 				if ( json_decode( $dict ) === null ) {
-					$options->set( 'language', $dict );
+					$options->set( 'language', $dict, false );
 				} else {
-					$options->set( 'language_dictionary', $dict );
+					$options->set( 'language_dictionary', $dict, false );
 				}
 			}
 		}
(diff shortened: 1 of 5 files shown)


lib/WP_Auth0_DBManager.php
```diff
@@ -118,28 +118,10 @@ public function install_db( $version_to_install = null ) {
...
 			}
 
-			// Force passwordless_method to delete
-			$update_options = $options->get_options();
-			unset( $update_options['passwordless_method'] );
-			update_option( $options->get_options_name(), $update_options );
-		}
-
-		// Plugin version < 3.7.0
-		if ( ( $this->current_db_version < 19 && 0 !== $this->current_db_version ) || 19 === $version_to_install ) {
```
joshcanhelp: This setting has been removed, no need to migrate

tests/testConstantSettings.php
```diff
@@ -88,19 +88,22 @@ public function testThatConstantValuesAreUsed() {
...
 		// Try to set an option with the constant set.
 		$new_value  = str_shuffle( $constant_val );
-		$set_result = $opts->set( $opt_name, $new_value, false );
+		$set_result = $opts->set( 'domain', $new_value );
 		$this->assertFalse( $set_result );
-		$this->assertEquals( $constant_val, $opts->get( $opt_name ) );
+		$this->assertEquals( $constant_val, $opts->get( 'domain' ) );
+
+		// Test that the constant value remains if a new option is set.
+		$opts->set( 'client_id', $new_value );
```
joshcanhelp: This the is test for the issue found while testing updates.

tests/testWPAuth0DbMigrations.php
```diff
@@ -13,41 +13,6 @@
  */
 class TestWPAuth0DbMigrations extends WP_Auth0_Test_Case {
 
-	/**
-	 * Test a DB upgrade from v18 to v19.
-	 */
-	public function testV19Update() {
```
joshcanhelp: This migration is no longer being run, setting has been removed.

lib/WP_Auth0_Options_Generic.php
```diff
@@ -174,16 +174,40 @@ public function set( $key, $value, $should_update = true ) {
...
 	/**
 	 * Save the options array as it exists in memory.
 	 *
 	 * @return bool
 	 */
 	public function update_all() {
+		$options = $this->get_options();
+
 		foreach ( $this->get_all_constant_keys() as $key ) {
-			unset( $this->_opts[ $key ] );
```
joshcanhelp: 🐞 was here ... we're removing from memory instead of just from the database.

lib/WP_Auth0_Options_Generic.php
```diff
@@ -110,14 +110,9 @@ public function get_options_name() {
...
 			$options = get_option( $this->_options_name, array() );
+			// Brand new install, no saved options so get all defaults.
 			if ( empty( $options ) || ! is_array( $options ) ) {
-
-				// Brand new install, no saved options so get all defaults.
 				$options = $this->defaults();
-			} else {
-
-				// Make sure we have settings for everything we need.
-				$options = array_merge( $this->defaults(), $options );
```
joshcanhelp: 🐞 was here ... we're overwriting missing settings, which will replace a false-y value (key missing) with a potentially truth-y one (if default is truth-y)

lib/WP_Auth0_Options_Generic.php
```diff
@@ -174,16 +169,40 @@ public function set( $key, $value, $should_update = true ) {
...
+
+		// Cannot remove a setting that is being overridden by a constant.
+		if ( $this->has_constant_val( $key ) ) {
+			return;
+		}
+
+		unset( $options[ $key ] );
+		$this->_opts = $options;
+
+		if ( $should_update ) {
```
lbalmaceda: I don't see any line in this diff passing should_update=true. Since the param is being introduced on this same PR I wonder **why** you need it and if if makes sense to remove it and always (or never) update the db.

tests/testConstantSettings.php
lbalmaceda: I don't understand what's being tested here. The steps I can follow are:
- `$constant_val` is a random string.
- You define a constant named "DOMAIN" with that value.
- You verify the value matches what you get via `$opts->get("domain")`. (How can I know that `$opts->get( 'domain' )` is going to take the value from the defined constant? )

- you then try to overwrite the "domain" option and fail. 
- You assert the value matches what was set initially. Still the value from the constant 👍 .
- But then you try to set client_id to some value, and still right below check again that "domain" option is still the constant. So:

1) Why checking twice that the `$opts->get( 'domain' )` is equal to the constant defined?
2) Why setting `client_id` option if not being asserted below?

tests/testWPAuth0Options.php
```diff
@@ -48,6 +48,22 @@ public function testThatDefaultsWork() {
 		}
 	}
 
+	/**
+	 * Test that a missing option (considered an empty value) is not updated to a default "on" value.
+	 */
+	public function testThatGetOptionsDoesNotUseDefaultsWhenNotEmpty() {
+
+		$db_options = get_option( self::OPTIONS_NAME );
+		unset( $db_options['auto_login'] );
```
lbalmaceda: any reason why you chose these 2 options? would it make sense to first assert they have a value, then unset it. Then call update. Then assert still empty?

tests/testConstantSettings.php
joshcanhelp: Yes, a little confusing 😄 which is what made it take so long to figure out where the 🐞 was coming from.

The constant settings are not stored in the database so, before saving, they are removed from the array that's saving it. The 🐞 here was that these were being removed from memory as well (what `get()` accesses) instead of only from the ones saved to the database. So changing a different option and then saving would change the correct option and save the correct array but remove all constant-stored settings for the rest of the request. On the next page load, all would be fine (unless the save happens again).

tests/testWPAuth0Options.php
joshcanhelp: These options have a default of `1` but, when saved, they were being removed from the options array instead of settings to a false-y value like `0`. This isn't optimal but works fine as the `get()` would send back something false-y. But if the defaults change (as they did in this version) and an update is run, then the missing options would be updated to defaults, which could be different.

lib/WP_Auth0_Options_Generic.php
```diff
@@ -174,16 +169,35 @@ public function set( $key, $value, $should_update = true ) {
 		return $this->update_all();
 	}
 
+	/**
+	 * Remove a setting from the options array in memory.
+	 *
+	 * @param string $key - Option key name to remove.
+	 */
+	public function remove( $key ) {
+		$options = $this->get_options();
```
lbalmaceda: you shouldn't need to call this method before asserting the `$key` is a constant. Let's move this line below that check and save that call for when it's actually required.

PR #675, opened 2019-04-22
Fix Auth0 logout redirect

joshcanhelp:
### Changes

Fix logout redirect if the plugin is not ready (no domain name or client ID)

### Testing

### Checklist

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -560,12 +560,12 @@ class WP_Auth0_LoginManager {
 	 * @link https://codex.wordpress.org/Plugin_API/Action_Reference/wp_logout
 	 */
 	public function logout() {
-		$is_sso        = (bool) $this->a0_options->get( 'sso' );
-		$is_slo        = (bool) $this->a0_options->get( 'singlelogout' );
-		$is_auto_login = (bool) $this->a0_options->get( 'auto_login' );
+		if ( ! WP_Auth0::ready() ) {
+			return;
+		}
 
 		// If SSO/SLO is in use, redirect to Auth0 to logout there as well.
-		if ( $is_sso || $is_slo ) {
+		if ( $this->a0_options->get( 'sso' ) || $this->a0_options->get( 'singlelogout' ) ) {
 			$return_to    = apply_filters( 'auth0_slo_return_to', home_url() );
 			$redirect_url = $this->auth0_logout_url( $return_to );
 			$redirect_url = apply_filters( 'auth0_logout_url', $redirect_url );
@@ -574,7 +574,7 @@ class WP_Auth0_LoginManager {
 		}
 
 		// If auto-login is in use, cannot redirect back to login page.
-		if ( $is_auto_login ) {
+		if ( (bool) $this->a0_options->get( 'auto_login' ) ) {
 			wp_redirect( home_url() );
 			exit;
 		}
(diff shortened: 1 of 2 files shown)


lib/WP_Auth0_LoginManager.php
```diff
@@ -560,12 +560,12 @@ private function do_login( $user, $userinfo, $is_new, $id_token, $access_token,
 	 * @link https://codex.wordpress.org/Plugin_API/Action_Reference/wp_logout
 	 */
 	public function logout() {
-		$is_sso        = (bool) $this->a0_options->get( 'sso' );
-		$is_slo        = (bool) $this->a0_options->get( 'singlelogout' );
-		$is_auto_login = (bool) $this->a0_options->get( 'auto_login' );
+		if ( ! WP_Auth0::ready() ) {
+			return;
```
lbalmaceda: so if you call this when the SDK is not ready (I suppose) this call would return without an error? How can they know they need to wait and retry?

lib/WP_Auth0_LoginManager.php
```diff
@@ -560,12 +560,12 @@ private function do_login( $user, $userinfo, $is_new, $id_token, $access_token,
...
+		}
 
 		// If SSO/SLO is in use, redirect to Auth0 to logout there as well.
-		if ( $is_sso || $is_slo ) {
+		if ( $this->a0_options->get( 'sso' ) || $this->a0_options->get( 'singlelogout' ) ) {
```
lbalmaceda: the auto-login check is casting to `bool`, why these aren't?

tests/testLoginManagerLogout.php
```diff
@@ -0,0 +1,122 @@
...
+
+		$logout_url_slo = parse_url( $redirect_data_slo['location'] );
+		$this->assertNotFalse( $logout_url_slo );
+		$this->assertEquals( 'https', $logout_url_slo['scheme'] );
+		$this->assertEquals( 'test.auth0.com', $logout_url_slo['host'] );
+		$this->assertEquals( '/v2/logout', $logout_url_slo['path'] );
+		$this->assertContains( 'client_id=__test_client_id__', $logout_url_slo['query'] );
+		$this->assertContains( 'returnTo=' . rawurlencode( home_url() ), $logout_url_slo['query'] );
+
+		// Test that SSO has the same behavior.
```
lbalmaceda: please move to a different test, even if most of the conditions are shared 👍

tests/testLoginManagerLogout.php
```diff
@@ -0,0 +1,122 @@
...
+		}
+
+		$this->assertEquals( 302, $redirect_data_sso['status'] );
+		$this->assertEquals( $redirect_data_slo['location'], $redirect_data_sso['location'] );
+	}
+
+	/**
+	 * Test that a redirect to the homepage happens if ULP is turned on.
+	 */
+	public function testThatRedirectHappensIfUlp() {
```
lbalmaceda: "if ULP is turned on" but I only see you change a "auto-login" parameter. Is the naming correctly?

lib/WP_Auth0_LoginManager.php
joshcanhelp: The `ready` here is checking if the plugin is configured. If not, then we don't need to do anything during this process.

lib/WP_Auth0_LoginManager.php
joshcanhelp: None of them need to, will fix auto-login

tests/testLoginManagerLogout.php
joshcanhelp: I'm checking against the result of the previous method call, though.

tests/testLoginManagerLogout.php
joshcanhelp: Yes. "Auto login" === ULP

PR #676, opened 2019-04-22
Use existing migration token during setup

joshcanhelp:
### Changes

This PR fixes an issue when a migration token exists (set in a constant) but does not get saved to the custom DB scripts. 

### Testing

### Checklist

### DIFF
--- a/lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
+++ b/lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
@@ -170,9 +170,12 @@ class WP_Auth0_InitialSetup_Consent {
 		if ( $should_create_and_update_connection ) {
 
 			if ( $connection_exists === false ) {
-				$migration_token = JWT::urlsafeB64Encode( openssl_random_pseudo_bytes( 64 ) );
-				$operations      = new WP_Auth0_Api_Operations( $this->a0_options );
-				$response        = $operations->create_wordpress_connection(
+				$migration_token = $this->a0_options->get( 'migration_token' );
+				if ( empty( $migration_token ) ) {
+					$migration_token = JWT::urlsafeB64Encode( openssl_random_pseudo_bytes( 64 ) );
+				}
+				$operations = new WP_Auth0_Api_Operations( $this->a0_options );
+				$response   = $operations->create_wordpress_connection(
 					$this->access_token,
 					$this->hasInternetConnection,
 					'fair',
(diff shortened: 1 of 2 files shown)


lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -170,9 +170,12 @@ public function consent_callback( $name ) {
 		if ( $should_create_and_update_connection ) {
 
 			if ( $connection_exists === false ) {
-				$migration_token = JWT::urlsafeB64Encode( openssl_random_pseudo_bytes( 64 ) );
-				$operations      = new WP_Auth0_Api_Operations( $this->a0_options );
-				$response        = $operations->create_wordpress_connection(
+				$migration_token = $this->a0_options->get( 'migration_token' );
+				if ( empty( $migration_token ) ) {
+					$migration_token = JWT::urlsafeB64Encode( openssl_random_pseudo_bytes( 64 ) );
```
lbalmaceda: what if this "is empty logic" is part of the `$this->a0_options->get( 'migration_token' );`? would it make sense, can that be done? so that whenever you request it you always obtain a valid value.

tests/testInitialSetupConsent.php
```diff
@@ -204,14 +204,70 @@ public function testThatNewConnectionIsCreatedAndFailedClientGrantRedirects() {
...
+		$this->assertEquals( '/wp-admin/admin.php', $redirect_url['path'] );
+		$this->assertContains( 'page=wpa0-setup', $redirect_url['query'] );
+		$this->assertContains( 'step=2', $redirect_url['query'] );
+		$this->assertContains( 'profile=social', $redirect_url['query'] );
+
+		$this->assertEquals( 'TEST_CLIENT_ID', self::$opts->get( 'client_id' ) );
+		$this->assertEquals( 'TEST_CLIENT_SECRET', self::$opts->get( 'client_secret' ) );
+		$this->assertEquals( 1, self::$opts->get( 'db_connection_enabled' ) );
+		$this->assertEquals( 'TEST_CREATED_CONN_ID', self::$opts->get( 'db_connection_id' ) );
+		$this->assertEquals( 'TEST_MIGRATION_TOKEN', self::$opts->get( 'migration_token' ) );
```
lbalmaceda: Please add a test that asserts that when this token is not set (Empty) a random token (not empty, length >x, etc) is being used. I couldn't find it on this file.

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
joshcanhelp: We don't really want that to happen. The value needs to be added to a script JS file and added to the Connection so generating one out of the blue would not be helpful. Also, there are times when we need to know if the setting is empty or not. Otherwise, we'd have to add an `has()` method or similar. These tokens are not generated often enough to make that worth it.

tests/testInitialSetupConsent.php
joshcanhelp: See the added line in the test above (which is uses that process)

PR #677, opened 2019-04-24
Add issuer and audience ID token validation

joshcanhelp:
### Changes

Adds issuer and audience ID token validation everywhere ID tokens are used.

### Testing

### Checklist

### DIFF
--- a/lib/WP_Auth0_Id_Token_Validator.php
+++ b/lib/WP_Auth0_Id_Token_Validator.php
@@ -1,0 +1,107 @@
+<?php
+/**
+ * Contains WP_Auth0_Id_Token_Validator.
+ *
+ * @package WP-Auth0
+ *
+ * @since 3.11.0
+ */
+
+/**
+ * Class WP_Auth0_Id_Token_Validator.
+ * Wrapper around JWT::decode() to do additional checks and enforce defaults.
+ */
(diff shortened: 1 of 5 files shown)


lib/WP_Auth0_Id_Token_Validator.php
```diff
@@ -0,0 +1,107 @@
...
+	 * @var string
+	 */
+	protected $id_token;
+
+	/**
+	 * Key to use to verify the signature.
+	 *
+	 * @var bool|string
+	 */
+	protected $key;
```
damieng: Can we call this secret to better indicate it is sensitive?

joshcanhelp: @damieng - This may or may not be secret. In the case of HS256 it is but for RS256 it's a public key from the JWKS.

damieng: Looks like it's only assigned from   

>  		$this->key       = $opts->get_client_secret_as_key();

joshcanhelp: Bad legacy method name. It gets the actual client secret from the database if the token algorithm is set to `HS256` and gets the public key from the JWKS if it's set to `RS256`.

PR #682, opened 2019-05-08
Translate plugin links and remove error log

joshcanhelp:
### Changes

- Update Lock to 11.16
- Translate plugin link text
- Remove plugin link to error log

### References

[Lock v11.16 release](https://github.com/auth0/lock/releases/tag/v11.16.0)


WP_Auth0.php
```diff
@@ -309,19 +309,24 @@ public function a0_register_query_vars( $qvars ) {
 	 */
 	public function wp_add_plugin_settings_link( $links ) {
 
-		$settings_link = '<a href="admin.php?page=wpa0-errors">Error Log</a>';
```
damieng: Looks like error log page is also removed but that PR desc doesn't mention that.  Intentional?

joshcanhelp: @damieng - Yes. I'll add to the description.

PR #694, opened 2019-06-04
Fix migration namespace and callback

joshcanhelp:
### Changes

- Change the namespace used when creating a new custom DB connection
- Remove the check for username on the get user endpoint
- Add `client_id` to the update user API call

### Testing

* [x] This change has been tested on WP 5.2.1

### Checklist

### DIFF
--- a/lib/WP_Auth0_Api_Operations.php
+++ b/lib/WP_Auth0_Api_Operations.php
@@ -84,7 +84,7 @@ class WP_Auth0_Api_Operations {
 				'bareConfiguration'            => array(
 					'endpointUrl'    => site_url( 'index.php?a0_action=' ),
 					'migrationToken' => $migration_token,
-					'userNamespace'  => get_auth0_curatedBlogName(),
+					'userNamespace'  => 'DB-' . get_auth0_curatedBlogName(),
 				),
 			);
 
--- a/lib/WP_Auth0_Routes.php
+++ b/lib/WP_Auth0_Routes.php
@@ -217,9 +217,6 @@ class WP_Auth0_Routes {
 			$username = $_POST['username'];
 
 			$user = get_user_by( 'email', $username );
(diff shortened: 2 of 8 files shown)


lib/WP_Auth0_Api_Operations.php
```diff
@@ -84,7 +84,7 @@ public function create_wordpress_connection( $app_token, $migration_enabled, $pa
 				'bareConfiguration'            => array(
 					'endpointUrl'    => site_url( 'index.php?a0_action=' ),
 					'migrationToken' => $migration_token,
-					'userNamespace'  => get_auth0_curatedBlogName(),
+					'userNamespace'  => 'DB-' . get_auth0_curatedBlogName(),
```
joshcanhelp: This is for new installations and has not been released yet so OK to change.

lib/WP_Auth0_Routes.php
```diff
@@ -217,9 +217,6 @@ protected function migration_ws_get_user() {
 			$username = $_POST['username'];
 
 			$user = get_user_by( 'email', $username );
-			if ( ! $user ) {
```
joshcanhelp: This endpoint is only used to check the email address.

lib/api/WP_Auth0_Api_Change_Email.php
```diff
@@ -66,6 +66,7 @@ public function call( $user_id = null, $email = null ) {
 			->add_body( 'email', $email )
 			// Email is either changed by an admin or verified by WP.
 			->add_body( 'email_verified', true )
+			->add_body( 'client_id', $this->options->get( 'client_id' ) )
```
joshcanhelp: [Recommend by the API docs for this endpoint](https://auth0.com/docs/api/management/v2#!/Users/patch_users_by_id)

PR #706, opened 2019-07-31
Fix widget gravatar and language settings

joshcanhelp:
### Changes

- Fix admin gravatar label not attached to correct input
- Fix gravatar override setting not working
- Fix translation setting not working

**Note to reviewers:** PO, MO, and POT files can be ignored.


lib/WP_Auth0_Lock10_Options.php
```diff
@@ -113,9 +118,14 @@ protected function build_settings( $settings ) {
 
 		$options_obj['socialButtonStyle'] = 'big';
 
-		if ( isset( $settings['gravatar'] ) && empty( $settings['gravatar'] ) ) {
+		if ( isset( $settings['gravatar'] ) && '' !== $settings['gravatar'] && empty( $settings['gravatar'] ) ) {
```
lbalmaceda: empty is already "". If you check for `!=""` and `empty` at the same time, this will never be called! What is going on here? 🤕

```
'' !== $settings['gravatar'] && empty( $settings['gravatar'] )
```

lib/WP_Auth0_Lock10_Options.php
```diff
@@ -113,9 +118,14 @@ protected function build_settings( $settings ) {
...
 			$options_obj['avatar'] = null;
 		}
+
+		if ( ! empty( $settings['gravatar'] ) ) {
```
lbalmaceda: you check for `empty` but not for `isset`, why ?

templates/a0-widget-setup-form.php
```diff
@@ -54,7 +54,7 @@
 		<input id="<?php echo $this->get_field_id( 'gravatar' ); ?>_inherit"
 			   name="<?php echo $this->get_field_name( 'gravatar' ); ?>"
 			   type="radio" value="" <?php echo esc_attr( $gravatar ) === '' ? 'checked="true"' : ''; ?> />
-		<label for="<?php echo $this->get_field_id( 'gravatar' ); ?>_no"><?php _e( 'Default Setting', 'wp-auth0' ); ?></label>
+		<label for="<?php echo $this->get_field_id( 'gravatar' ); ?>_inherit"><?php _e( 'Default Setting', 'wp-auth0' ); ?></label>
```
lbalmaceda: how does this impact the page?

joshcanhelp: The label was tied to the `no` value so this fixes that behavior.

lib/WP_Auth0_Lock10_Options.php
joshcanhelp: It's a little wonky because it has a number of settings arrays that override each other with successive option building. 

`''` is an empty string (in this case, means use the default), `isset()` is whether that key exists or not (avoids `Undefined index` errors), and `empty()` checks for `0`, `null`, `false`, etc.

lib/WP_Auth0_Lock10_Options.php
joshcanhelp: `empty()` covers the `isset()` case. Means "no truth-y value to use".

PR #707, opened 2019-08-01
Add check for GET and POST globals for state validation

joshcanhelp:
### Changes

Add checking for state in other PHP globals in case `$_REQUEST` is not set.

### Testing

* [ ] This change adds unit test coverage

### Checklist

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -625,7 +625,7 @@ class WP_Auth0_LoginManager {
 	}
 
 	/**
-	 * Get a value from query_vars or $_REQUEST global.
+	 * Get a value from query_vars or request global.
 	 *
 	 * @param string $key - query var key to return.
 	 *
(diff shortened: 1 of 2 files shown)


lib/WP_Auth0_LoginManager.php
```diff
@@ -639,6 +639,12 @@ protected function query_vars( $key ) {
 		if ( isset( $_REQUEST[ $key ] ) ) {
 			return $_REQUEST[ $key ];
 		}
+		if ( isset( $_GET[ $key ] ) ) {
```
damieng: Is this function only used in the WP_Auth0_LoginManager above?

I'm wondering if now accepting values from GET or POST as well as the previous objects opens up any kind of exploits.  i.e. Perhaps we were validating some parameter we got posted back from auth but now the user could pre-inject a value on the URL so when we get redirected back we take the value from there rather than from the POST...

I think this function is a bit dangerous.

joshcanhelp: > Is this function only used in the WP_Auth0_LoginManager above?

Yes, protected. 

I totally understand your concerns here and thought about implications. It's used to:

- Get an `auth0` parameter, which just triggers callback processing
- Checks for a state param, which is validated one time against a stored value
- Gets a code parameter which is exchanged immediately for tokens

The existing checks can only be internally set in WordPress, the ones I'm adding would just roll up into those if they are present (if everything goes according to plan) so no additional risk added here. 

I'll address this in the upcoming, I agree that we should be looking for these values only where we expect them.

PR #716, opened 2019-08-13
Update auth params method to add filters

joshcanhelp:
### Changes

- Move filter `auth0_authorize_url_params` before state is generated and decoded
- Add filter `auth0_authorize_state` to modify state contents before decoding

### Testing

### Checklist

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -569,22 +569,21 @@ class WP_Auth0_LoginManager {
 	 * @return array
 	 */
 	public static function get_authorize_params( $connection = null, $redirect_to = null ) {
-		$params       = [];
-		$options      = WP_Auth0_Options::Instance();
+		$opts         = WP_Auth0_Options::Instance();
 		$lock_options = new WP_Auth0_Lock10_Options();
-		$is_implicit  = (bool) $options->get( 'auth0_implicit_workflow', false );
+		$is_implicit  = (bool) $opts->get( 'auth0_implicit_workflow', false );
 		$nonce        = WP_Auth0_Nonce_Handler::get_instance()->get_unique();
 
-		$params['client_id']     = $options->get( 'client_id' );
-		$params['scope']         = self::get_userinfo_scope( 'authorize_url' );
-		$params['response_type'] = $is_implicit ? 'id_token' : 'code';
-		$params['redirect_uri']  = $is_implicit
-			? $lock_options->get_implicit_callback_url()
-			: $options->get_wp_auth0_url();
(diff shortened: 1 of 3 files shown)


tests/testLoginManager.php
```diff
@@ -46,49 +46,6 @@ function( $default_scope, $context ) {
 		$this->assertEquals( 'openid email profile auth0', $scope );
 	}
 
-	/**
```
joshcanhelp: Split out into different tests in the class added below 👇

lib/WP_Auth0_LoginManager.php
```diff
@@ -569,47 +569,48 @@ public static function get_userinfo_scope( $context = '' ) {
...
-		$params['state'] = base64_encode(
-			json_encode(
-				[
-					'interim'     => false,
-					'nonce'       => $nonce,
-					'redirect_to' => filter_var( $redirect_to, FILTER_SANITIZE_URL ),
-				]
-			)
-		);
+		if ( empty( $filtered_params['state'] ) ) {
```
joshcanhelp: This allows backwards-compat with anyone generating their own state.

lib/WP_Auth0_LoginManager.php
```diff
@@ -569,47 +569,48 @@ public static function get_userinfo_scope( $context = '' ) {
...
+			'client_id'     => $opts->get( 'client_id' ),
+			'scope'         => self::get_userinfo_scope( 'authorize_url' ),
+			'response_type' => $is_implicit ? 'id_token' : 'code',
+			'response_mode' => $is_implicit ? 'form_post' : 'query',
+			'redirect_uri'  => $is_implicit ? $lock_options->get_implicit_callback_url() : $opts->get_wp_auth0_url(),
+		];
 
 		if ( $is_implicit ) {
-			$params['nonce']         = $nonce;
-			$params['response_mode'] = 'form_post';
```
lbalmaceda: No longer an option? 🤔

lib/WP_Auth0_LoginManager.php
lbalmaceda: nonce can be customized?

joshcanhelp: State-level unique value can be (default generated state would always be the same value if not). Use case here is for adding values, though.

lib/WP_Auth0_LoginManager.php
joshcanhelp: Moved up to be more explicit

joshcanhelp:
@lbalmaceda - Not urgent but *ping* for you here :)

PR #717, opened 2019-08-14
Add wpa0_user_data filter before creating WP_User

horike37:
### Changes

Add filter `wpa0_user_data`  to allow customizing user data before creating WP_User because I would like to separate `display_name` from `username`. The current implementation hardcodes the WordPress `user_login` and `display_name` to come from the Auth0 `username`. However, in many cases, those are separated based on naming style in Japan.

### References

There is no reference. only this PR.

### Testing

### Checklist

* [x] All relevant assets have been compiled as directed in the [Contribution guide](CONTRIBUTION.md), if applicable
* [ ] All active GitHub CI checks have passed

### DIFF
--- a/lib/WP_Auth0_Users.php
+++ b/lib/WP_Auth0_Users.php
@@ -84,7 +84,7 @@ class WP_Auth0_Users {
 			'description'  => $description,
 		];
 
-		$user_data = apply_filters( 'wpa0_user_data', $user_data, $userinfo, $firstname, $lastname );
+		$user_data = apply_filters( 'auth0_create_user_data', $user_data, $userinfo );
 
 		if ( $role ) {
 			// phpcs:ignore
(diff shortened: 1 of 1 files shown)


lib/WP_Auth0_Users.php
```diff
@@ -84,6 +84,8 @@ public static function create_user( $userinfo, $role = null ) {
 			'description'  => $description,
 		];
 
+		$user_data = apply_filters( 'wpa0_user_data', $user_data, $userinfo, $firstname, $lastname );
```
joshcanhelp: The naming convention for hooks isn't 100% consistent but we've been using `auth0_` as the namespace for new ones. Can you update the filter name to `auth0_create_user_data` to add that namespace and make this more explicit?

Also, the `$firstname` and `$lastname` parameters are not necessary since their values are already in the array.

horike37: Thank you for review 👍 Happy to hear that this PR will be merged 😄 
I just updated all what you pointed. Please check it again.

PR #718, opened 2019-08-14
Adding Brazilian Portuguese translations

niugait:
### Changes

Added Brazilian Portuguese translations


joshcanhelp:
*Muito obrigado!*

I have a few PRs coming through (including #715) that will reduce the number of translations we have in these files. I'll ping this thread when we're more stable with those to update your translation files (will mostly be removals).

niugait:
Sure thing! Just let me know and I'll update some of the translations that could use other approach as well.

joshcanhelp:
@niugait - Apologies for the delay! It looks like your fork is deleted now ... I was going to ask you to rebase the `master` branch so we can review (there were a number of strings that were removed). If you're still around and can do that, we'd love to merge this in! Otherwise, do you mind if we use what's here and put through our own PR? 

Thank you!

PR #722, opened 2019-09-05
Remove deprecated basic settings; centralize validation declaration

joshcanhelp:
### Changes

- Remove deprecated method `WP_Auth0_Admin_Basic::render_auth0_app_token()`
- Remove deprecated method `WP_Auth0_Admin_Basic::render_allow_signup_regular_multisite()`
- Remove deprecated method `WP_Auth0_Admin_Basic::render_allow_signup_regular()`
- Remove deprecated method `WP_Auth0_Admin_Basic::render_basic_description()`
- Remove method `WP_Auth0_Admin_Basic::render_auth0_app_token()` and replace with `wp_auth0_ajax_delete_cache_transient()` to handle JWKS delete cache AJAX action.
- Moved `$_description` and `$actions_middlewares` definition to main class

### Testing


WP_Auth0.php
```diff
@@ -561,6 +561,13 @@ public static function get_plugin_dir_url() {
  * Core WP hooks
  */
 
+function wp_auth0_ajax_delete_cache_transient() {
+	check_ajax_referer( 'auth0_delete_cache_transient' );
+	delete_transient( WPA0_JWKS_CACHE_TRANSIENT_NAME );
+	wp_send_json_success();
+}
+add_action( 'wp_ajax_auth0_delete_cache_transient', 'wp_auth0_ajax_delete_cache_transient' );
```
luisrudge: Previously, it just called `exit`. Why are you returning json_success now? https://github.com/auth0/wp-auth0/pull/722/files#diff-729ab756a0f851fef2f036fa7ca87b13L351

joshcanhelp: `wp_send_json_success()` is a core WP function that returns a successful response, then calls `exit` (or `die` which is the same). Just provides a little more feedback to the calling script.

PR #730, opened 2019-10-08
Get new access token via refresh token API

albeja:
Added a class for requesting a new access token via the refresh token (as old "get_token()"-method was deprecated and deleted lately).

Followed the coding style from `WP_Auth0_Api_Exchange_Code.php`.


joshcanhelp:
@albeja - Thanks for this! I'll review as soon as I have a chance.

lib/api/WP_Auth0_Api_Refresh_Access_Token.php
```diff
@@ -0,0 +1,85 @@
...
+
+		return $this
+			->set_path( 'oauth/token' )
+			->add_body( 'grant_type', 'refresh_token' )
+			->add_body( 'client_id', $client_id )
+			->add_body( 'client_secret', $client_secret )
+			->add_body( 'refresh_token', $refresh_token )
+			->post()
+			->handle_response( __METHOD__ );
+
```
joshcanhelp:
```suggestion
```

lib/api/WP_Auth0_Api_Refresh_Access_Token.php
```diff
@@ -0,0 +1,85 @@
...
+	 *
+	 * @return integer
+	 */
+	protected function handle_response( $method ) {
+
+		if ( 401 == $this->response_code ) {
+			WP_Auth0_ErrorManager::insert_auth0_error(
+				__METHOD__ . ' L:' . __LINE__,
+				__( 'An /oauth/token call triggered a 401 response from Auth0. ', 'wp-auth0' ) .
+				__( 'Please check the Client Secret saved in the Auth0 plugin settings. ', 'wp-auth0' )
```
joshcanhelp:
```suggestion
				__( 'Please check the Client ID and Client Secret saved in the Auth0 plugin settings. ', 'wp-auth0' )
```

lib/api/WP_Auth0_Api_Refresh_Access_Token.php
```diff
@@ -0,0 +1,85 @@
...
+			);
+			return self::RETURN_ON_FAILURE;
+		}
+
```
joshcanhelp: Let's also add a check for a `403` response, which would be an invalid refresh token or an Application that does not allow the refresh token grant (**Auth0 Dashboard > Applications > WP Application > Show Advanced Settings (bottom) > Grant Types**:

Screenshot 2019-10-10 12 05 00

The `error_description` that's returned by the API will work for the message used in the log.

albeja: Hey Josh,
thanks for the suggestion :)!
Actually I thought this would already be covered by the call `$this->handle_failed_response( $method )`. I thought in case of a 403 the method of the `WP_Auth0_Api_Abstract` would call the `WP_Auth0_ErrorManager::insert_auth0_error` with the APIs error description. Doesn't it?
If not, I would implement a handling of the 403 close to the handling in this abstract method (if that's OK for you).

joshcanhelp: @albeja - Good call, I think that should suffice.

lib/api/WP_Auth0_Api_Refresh_Access_Token.php
```diff
@@ -0,0 +1,87 @@
+<?php
+/**
+ * Contains CUSTOM WP_Auth0_Api_Refresh_Access_Token.
```
joshcanhelp:
```suggestion
 * Contains WP_Auth0_Api_Refresh_Access_Token.
```

albeja:
Wonderful! Unfortunately I have difficulties with my local phpunit-environment at the moment. That's why I did not write any tests so far.
I'll try to fix that issue this week so that I'm able to contribute to the tests as well.

joshcanhelp:
@albeja - I'll leave this open for now so you can add those tests. Let me know if you need any guidance setting up the testing suite beyond [the docs we have](https://github.com/auth0/wp-auth0/blob/master/CONTRIBUTING.md).

PR #734, opened 2019-11-23
Improve OIDC Compliance

joshcanhelp:
### Changes

This update improves the SDK support for OpenID Connect. In particular, it modifies the sign in verification phase by substituting backchannel based checks with id_token validation.

**Note to reviewers:** This re-uses the majority of the PHP SDK token verifier work. [Please see this draft PR](https://github.com/auth0/auth0-PHP/pull/401/files) to review the differences between the core verification classes. ~540 added lines are represented in that diff.

### References

- [OpenID Connect specification: ID token claims to verify](https://openid.net/specs/openid-connect-core-1_0-final.html#IDToken)
- [PHP SDK classes](https://github.com/auth0/auth0-PHP/pull/401/files)

### Testing

**Manual testing:**

1. In wp-admin > Auth0 > Settings, turn Implicit login flow **off** in settings and universal login page **on**.
2. Logout then log back in.
3. Make sure the redirect response from WP includes both state and nonce cookies:

Screenshot 2019-11-25 09 40 04

4. Make sure the /authorize URL has all the expected params:

Screenshot 2019-11-25 11 06 48

5. Log in at Auth0
6. Make sure you're logged in with no errors in the plugin error log
7. Make sure that the nonce and state cookies are deleted on the callback response:

Screenshot 2019-11-25 12 04 27

* [x] This change has been tested on WP 5.3 and PHP 7.0

### Checklist

* [x] All relevant assets have been compiled as directed in the [Contribution guide](CONTRIBUTION.md), if applicable


functions.php
```diff
@@ -117,6 +117,33 @@ function wp_auth0_can_show_wp_login_form() {
 	return false;
 }
 
+/**
```
joshcanhelp: Both of these functions were pulled from the old JWT library.

composer.json
```diff
@@ -4,7 +4,8 @@
     "homepage": "https://auth0.com/wordpress",
     "license": "GPLv2",
     "require": {
-        "php": "^7.0"
+        "php": "^7.0",
+        "lcobucci/jwt": "^3.3.0"
```
joshcanhelp: Previous library was not managed with Composer.

lib/token-verifier/WP_Auth0_AsymmetricVerifier.php
```diff
@@ -0,0 +1,54 @@
+<?php
```
joshcanhelp: [See this diff](https://github.com/auth0/auth0-PHP/pull/401/files#diff-71973553f78976e53afc2cfa2b20d861) for changes from approved PHP SDK ones.

lib/token-verifier/WP_Auth0_IdTokenVerifier.php
```diff
@@ -0,0 +1,265 @@
+<?php
```
joshcanhelp: [See this diff](https://github.com/auth0/auth0-PHP/pull/401/files#diff-3871126b680fac12ec0aa3124b74e102) for changes from approved PHP SDK ones.

lib/token-verifier/WP_Auth0_JwksFetcher.php
```diff
@@ -0,0 +1,76 @@
+<?php
```
joshcanhelp: [See this diff](https://github.com/auth0/auth0-PHP/pull/401/files#diff-8397559345a7685d64dc018493eeb7a1) for changes from approved PHP SDK ones.

lib/token-verifier/WP_Auth0_SignatureVerifier.php
```diff
@@ -0,0 +1,87 @@
+<?php
```
joshcanhelp: [See this diff](https://github.com/auth0/auth0-PHP/pull/401/files#diff-82a6ce33086c518324a561bdd71b8a9b) for changes from approved PHP SDK ones.

lib/token-verifier/WP_Auth0_SymmetricVerifier.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
joshcanhelp: [See this diff](https://github.com/auth0/auth0-PHP/pull/401/files#diff-25934a723f93d7eb12e57c7f8dcf53fe) for changes from approved PHP SDK ones.

lib/WP_Auth0_LoginManager.php
```diff
@@ -232,8 +228,7 @@ public function implicit_login() {
 		$id_token_param = ! empty( $_POST['id_token'] ) ? $_POST['id_token'] : $_POST['token'];
 		$id_token       = sanitize_text_field( wp_unslash( $id_token_param ) );
 
-		$jwt_verifier  = new WP_Auth0_Id_Token_Validator( $id_token, $this->a0_options );
-		$decoded_token = $jwt_verifier->decode( true );
+		$decoded_token = $this->decode_id_token( $id_token );
 		$decoded_token = $this->clean_id_token( $decoded_token );
```
lbalmaceda: better to return token as is and not let the library modify its contents

lib/WP_Auth0_LoginManager.php
```diff
@@ -573,6 +562,31 @@ protected function die_on_login( $msg = '', $code = 0 ) {
...
+	 */
+	private function decode_id_token( $id_token ) {
+		$idTokenIss  = 'https://' . $this->a0_options->get( 'domain' ) . '/';
+		$sigVerifier = null;
+		if ( 'RS256' === $this->a0_options->get( 'client_signing_algorithm' ) ) {
+			$jwks        = ( new WP_Auth0_JwksFetcher() )->getKeys();
+			$sigVerifier = new WP_Auth0_AsymmetricVerifier( $jwks );
+		} elseif ( 'HS256' === $this->a0_options->get( 'client_signing_algorithm' ) ) {
+			$sigVerifier = new WP_Auth0_SymmetricVerifier( $this->a0_options->get( 'client_secret' ) );
+		}
```
lbalmaceda: "else" scenario????

Also, you're doing `$this->a0_options->get( 'client_signing_algorithm' )` twice. Seems like it can be extracted. Does it have performance impact?

lib/WP_Auth0_LoginManager.php
```diff
@@ -573,6 +562,31 @@ protected function die_on_login( $msg = '', $code = 0 ) {
...
+
+		$verifierOptions = [
+			'nonce'   => WP_Auth0_Nonce_Handler::get_instance()->get_once(),
```
lbalmaceda: was about to say "typo" but I'm not sure if this is intentional or not. On other places, you call `get_unique`. Please, evaluate.

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -461,8 +461,9 @@ public function migration_ws_validation( array $old_options, array $input ) {
 		}
 
 		try {
-			$token_decoded               = JWT::decode( $input['migration_token'], $secret, [ 'HS256' ] );
-			$input['migration_token_id'] = isset( $token_decoded->jti ) ? $token_decoded->jti : null;
+			$signature_verifier          = new WP_Auth0_SymmetricVerifier( $secret );
```
lbalmaceda: It looks similar to the file `WP_Auth0_Routes`. Should it be extracted?

lib/token-verifier/WP_Auth0_AsymmetricVerifier.php
```diff
@@ -0,0 +1,56 @@
...
+	 * @param Token $token Parsed token to check.
+	 *
+	 * @return boolean
+	 *
+	 * @throws WP_Auth0_InvalidIdTokenException If ID token kid was not found in the JWKS.
+	 */
+	protected function checkSignature( Token $token ) : bool {
+		$tokenKid = $token->getHeader( 'kid', false );
+		if ( ! array_key_exists( $tokenKid, $this->jwks ) ) {
+			throw new WP_Auth0_InvalidIdTokenException( 'ID token key ID "' . $tokenKid . '" was not found in the JWKS' );
```
lbalmaceda: Should be

> Could not find a public key for Key ID (kid) "${decoded.header.kid}"

tests/testLoginManagerAuthParams.php
```diff
@@ -120,4 +136,15 @@ public function filter_state_parms( $state, $filtered_params ) {
...
 	}
+
+	/**
+	 * Helper method to filter max_age for testing.
+	 *
+	 * @param null $max_age_null - Existing max_age setting.
+	 *
+	 * @return int
+	 */
+	public function filter_max_age( $max_age_null ) {
```
lbalmaceda: parameter is ignored

.circleci/config.yml
```diff
@@ -55,7 +52,7 @@ commands:
 jobs:
   php_7:
     docker:
-      - image: circleci/php:7.1
+      - image: circleci/php:7.0
```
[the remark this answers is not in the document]
joshcanhelp: 7.0, hence the change :)

lib/WP_Auth0_LoginManager.php
joshcanhelp: I'm not changing how the token is used in this PR, only how it's validated.

lib/WP_Auth0_LoginManager.php
joshcanhelp: `else` is covered above this block, `$sigVerifier` will be `null` and fail.

It's just pulling from an internal array.

lib/WP_Auth0_LoginManager.php
joshcanhelp: `get_unique` gets the unique, stored value and `get_once()` gets the value and deletes it.

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: This logic will likely be removed as part of the major, didn't want to go too far in this PR.

tests/testLoginManagerAuthParams.php
joshcanhelp: Expected.

joshcanhelp:
@lbalmaceda 

> Please pay attention to the claim error messages. Should be wrapping the expected/received values between ( ) rather than " ". I see you're mixing them.

It looks like that was mapped over from the PHP SDK. We'll extract this out into it's own library in Phase 2 so this is consistent everywhere.

lib/WP_Auth0_LoginManager.php
lbalmaceda: on platforms like Java, taking the Map class for example, that would be called [get](https://docs.oracle.com/javase/7/docs/api/java/util/Map.html#get(java.lang.Object)) and [remove](https://docs.oracle.com/javase/7/docs/api/java/util/Map.html#remove(java.lang.Object)) (which also returns the existing value). Unless this is a pattern for PHP, or it's already documented the expected behavior of each method, consider renaming them. Not a blocker

joshcanhelp: The classes and methods here, this one in particular, are unlikely to be used by a developer so I'm not too worried about them. `get_once()` feels clear to me for what it's doing. `get_unique()` could probably change (should probably just be `get` since there is only ever one value for the class).

PR #757, opened 2020-01-07
Remove future iat check

joshcanhelp:
### Changes

Remove `iat` claim value check for ID token validation; continue to check for presence.

### Testing

- [x] This modifies test coverage
- [x] This change was created in PHP 7.2

### Checklist

- [x] All existing and new tests complete without errors.
- [x] The correct base branch is being used.


joshcanhelp:
@damieng - It's a copy of what's in the PHP SDK, will be extracted out to a separate library.

PR #772, opened 2020-01-31
Remove screenshots and GH templates; update readme.txt

joshcanhelp:
### Changes

Cleans up a number of file for this repo and the WP SVN one:

- Remove GitHub templates to start using [our global ones](https://github.com/auth0/.github)
- Remove screenshots and banners (these are stored in the WP SVN `assets` folder)
- Update WP `readme.txt` to include the current stable tag and new plugin directory tags (to make what's currently there)

### References

[Information on the Stable Tag](https://developer.wordpress.org/plugins/wordpress-org/how-your-readme-txt-works/#how-the-readme-is-parsed)


joshcanhelp:
@lbalmaceda - I think it's a good idea. The global ones have more eyes on them and are easier to manage. Unfortunately, we finished adding all the individual templates to these repos right before the global templates were announced 😞

PR #773, opened 2020-01-31
Adding advanced setting validation; normalize values throughout

joshcanhelp:
### Description

- Normalize settings validation for the Advanced tab
- Normalize defaults for the Advanced tab
- Remove `WP_Auth0_Admin_Advanced->render_auto_login()`

### References

[WP guidance on sanitization and escaping](https://developer.wordpress.org/themes/theme-security/data-sanitization-escaping/)

### Testing

- [x] This change improves test coverage for new/changed/fixed functionality

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,8 +42,8 @@
         "test-cov": "\"vendor/bin/phpunit\" --coverage-text",
         "test-group": "\"vendor/bin/phpunit\" --coverage-text --group",
         "test-ci": "\"vendor/bin/phpunit\" --coverage-clover=coverage.xml",
-        "pre-commit-no-tests": [ "@phpcbf", "@phpcbf-tests", "@phpcs-tests", "@compat", "@phpcs-i18n" ],
-        "pre-commit": [ "@phpcbf", "@phpcbf-tests", "@phpcs-tests", "@compat", "@phpcs-i18n", "@test" ]
+        "pre-commit-no-tests": [ "@phpcbf", "@phpcbf-tests", "@compat", "@phpcs-i18n" ],
+        "pre-commit": [ "@phpcbf", "@phpcbf-tests", "@compat", "@phpcs-i18n", "@test" ]
     },
     "autoload-dev": {
         "classmap": ["tests/classes/", "tests/traits/"]
--- a/lib/admin/WP_Auth0_Admin_Advanced.php
+++ b/lib/admin/WP_Auth0_Admin_Advanced.php
@@ -104,12 +104,6 @@ class WP_Auth0_Admin_Advanced extends WP_Auth0_Admin_Generic {
 				'function' => 'render_migration_ws_ips',
 			],
 			[
-				'name'     => __( 'Auto Login', 'wp-auth0' ),
-				'opt'      => 'auto_login',
-				'id'       => 'wpa0_auto_login',
-				'function' => 'render_auto_login',
-			],
-			[
 				'name'     => __( 'Valid Proxy IP', 'wp-auth0' ),
 				'opt'      => 'valid_proxy_ip',
 				'id'       => 'wpa0_valid_proxy_ip',
@@ -319,16 +313,6 @@ class WP_Auth0_Admin_Advanced extends WP_Auth0_Admin_Generic {
 	}
 
 	/**
-	 * Refer to the Features tab for the `auto_login` option.
-	 * IMPORTANT: Internal callback use only, do not call this function directly!
-	 */
-	public function render_auto_login() {
-		$this->render_field_description(
-			__( 'Please see the "Universal Login Page" setting on the Features tab', 'wp-auth0' )
-		);
-	}
-
-	/**
 	 * Render form field and description for the `valid_proxy_ip` option.
 	 * IMPORTANT: Internal callback use only, do not call this function directly!
 	 *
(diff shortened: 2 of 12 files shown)


composer.json
```diff
@@ -42,8 +42,8 @@
         "test-cov": "\"vendor/bin/phpunit\" --coverage-text",
         "test-group": "\"vendor/bin/phpunit\" --coverage-text --group",
         "test-ci": "\"vendor/bin/phpunit\" --coverage-clover=coverage.xml",
-        "pre-commit-no-tests": [ "@phpcbf", "@phpcbf-tests", "@phpcs-tests", "@compat", "@phpcs-i18n" ],
-        "pre-commit": [ "@phpcbf", "@phpcbf-tests", "@phpcs-tests", "@compat", "@phpcs-i18n", "@test" ]
+        "pre-commit-no-tests": [ "@phpcbf", "@phpcbf-tests", "@compat", "@phpcs-i18n" ],
```
joshcanhelp: Removing code style checker on tests ... not a lot of value here.

lib/admin/WP_Auth0_Admin_Appearance.php
```diff
@@ -290,10 +290,11 @@ public function render_extra_conf( $args = [] ) {
 	 * @return array
 	 */
 	public function basic_validation( $old_options, $input ) {
-		$input['form_title']    = empty( $input['form_title'] ) ? '' : sanitize_text_field( $input['form_title'] );
-		$input['icon_url']      = empty( $input['icon_url'] ) ? '' : esc_url( $input['icon_url'], [ 'http', 'https' ] );
-		$input['gravatar']      = empty( $input['gravatar'] ) ? 0 : 1;
-		$input['primary_color'] = empty( $input['primary_color'] ) ? '' : sanitize_text_field( $input['primary_color'] );
+		$input['passwordless_enabled'] = ( isset( $input['passwordless_enabled'] ) ? $input['passwordless_enabled'] : 0 ) == 1;
```
joshcanhelp: Moved this setting to the right place ... no functional changes

lib/admin/WP_Auth0_Admin_Appearance.php
```diff
@@ -305,6 +306,17 @@ public function basic_validation( $old_options, $input ) {
 			self::add_validation_error( __( 'The Lock JS CDN URL used is not a valid URL.', 'wp-auth0' ) );
 		}
 
+		$input['lock_connections'] = isset( $input['lock_connections'] ) ?
```
joshcanhelp: Same here, just moving the settings to the right place. Will address validation here in a future PR

lib/admin/WP_Auth0_Admin_Generic.php
```diff
@@ -258,4 +258,12 @@ protected function get_docs_link( $path, $text = '' ) {
 		$text = empty( $text ) ? __( 'here', 'wp-auth0' ) : sanitize_text_field( $text );
 		return sprintf( '<a href="https://auth0.com/docs/%s" target="_blank">%s</a>', $path, $text );
 	}
+
+	protected function sanitize_switch_val( $val ) {
```
joshcanhelp: This will handle values from the UI (which save a positive as `"1"`) as well as JSON-imported. Need to handle a few (but not all) truth-y values.

lib/admin/WP_Auth0_Admin_Generic.php
```diff
@@ -258,4 +258,12 @@ protected function get_docs_link( $path, $text = '' ) {
...
+		return in_array( $val, [ 1, '1', true ] ) ? true : false;
+	}
+
+	protected function sanitize_text_val( $val ) {
+		return sanitize_text_field( trim( strval( $val ) ) );
```
joshcanhelp: https://developer.wordpress.org/reference/functions/sanitize_text_field/

lib/initial-setup/WP_Auth0_InitialSetup_Consent.php
```diff
@@ -131,7 +131,7 @@ public function consent_callback( $name ) {
 		if ( $should_create_connection ) {
 			$migration_token = $this->a0_options->get( 'migration_token' );
 			if ( empty( $migration_token ) ) {
-				$migration_token = wp_auth0_url_base64_encode( openssl_random_pseudo_bytes( 64 ) );
+				$migration_token = wp_auth0_generate_token();
```
joshcanhelp: https://github.com/auth0/wp-auth0/blob/3d9a497928c138911cbd049e2d9516cd670c5318/functions.php#L17-L20

tests/testRequiredEmail.php
```diff
@@ -73,22 +73,23 @@ public function testRequiredEmailFieldOutput() {
 	 * Test that the required email field is properly validated.
 	 */
 	public function testRequiredEmailValidation() {
-		$opt_name = 'requires_verified_email';
+		$validated_opts = self::$admin->basic_validation( [], [ 'requires_verified_email' => '1' ] );
```
joshcanhelp: Testing out multiple truth-y and false-y values, difference is intensional.

tests/testUserRepoCreate.php
```diff
@@ -49,7 +49,7 @@ public function testRequiredEmailRejected() {
 		$this->createUser( [ 'user_email' => $userinfo->email ] );
```
joshcanhelp: Set the values to be the same as what the UI would save

tests/testRoutesLogin.php
```diff
@@ -64,8 +64,8 @@ public function testThatLoginRouteIsForbiddenByDefault() {
 	 * If the incoming IP address is invalid, the route should fail with an error.
```
joshcanhelp: Set the values to be the same as what the UI would save

tests/testRoutesGetUser.php
```diff
@@ -63,8 +63,8 @@ public function testThatGetUserRouteIsForbiddenByDefault() {
 	 * If the incoming IP address is invalid, the route should fail with an error.
```
joshcanhelp: Set the values to be the same as what the UI would save

tests/testOptionMigrationWs.php
```diff
@@ -126,12 +126,12 @@ public function testThatCorrectFieldDocsShowWhenMigrationIsOn() {
 	public function testThatChangingMigrationToOffKeepsTokenData() {
```
joshcanhelp: Setting these test to use the same value that the validator would expect from the UI.

lib/admin/WP_Auth0_Admin_Appearance.php
```diff
@@ -290,10 +290,11 @@ public function render_extra_conf( $args = [] ) {
 	 * @return array
```
joshcanhelp: Everything in this file is moved from somewhere else. No functional changes, validation will be addressed in a separate PR.

lib/admin/WP_Auth0_Admin_Advanced.php
```diff
@@ -407,11 +378,11 @@ public function basic_validation( array $old_options, array $input ) {
 	 * @return array
 	 */
 	public function migration_ws_validation( array $old_options, array $input ) {
-		$input['migration_ws']    = (int) ! empty( $input['migration_ws'] );
+		$input['migration_ws']    = $this->sanitize_switch_val( $input['migration_ws'] ?? null );
```
lbalmaceda: on the `WP_Auth0_Options` file you changed existing `null` values to `''`. This line is setting one of those entries as null again. Is that intentional?

lib/admin/WP_Auth0_Admin_Appearance.php
lbalmaceda: Just a small suggestion. Might add readability to extract a constant for 0 and 1 like `OPTION_DISABLED`, `OPTION_ENABLED`

joshcanhelp: Well ... it's not as simple as that. "Disabled" and "enabled" depends on context. The UI sends `"1"` as enabled, which should be converted to `true`.

By the end of this series of PRs, we'll be normalized on `true/false` for checking.

lib/admin/WP_Auth0_Admin_Advanced.php
joshcanhelp: Yes. This is doing null coalescing, so we're checking for the key (key will be missing for "off" values sent by the UI), then defaulting to `null`. The sanitize method there does the final conversion (starting with string type casting).

lbalmaceda:
Green

PR #774, opened 2020-01-31
Update feature settings validation

joshcanhelp:
### Description

- Normalize settings validation for the Features tab
- Normalize defaults for the Features tab

### References

[WP guidance on sanitization and escaping](https://developer.wordpress.org/themes/theme-security/data-sanitization-escaping/)

### Testing

- [x] This change improves test coverage for new/changed/fixed functionality


lib/admin/WP_Auth0_Admin_Generic.php
```diff
@@ -259,8 +259,15 @@ protected function get_docs_link( $path, $text = '' ) {
...
+	/**
+	 * Strict-check passed values against possible truth-y ones.
+	 *
+	 * @param mixed $val Value to check.
+	 *
+	 * @return bool
+	 */
 	protected function sanitize_switch_val( $val ) {
-		return in_array( $val, [ 1, '1', true ] ) ? true : false;
+		return in_array( $val, [ 1, '1', true ], true ) ? true : false;
```
joshcanhelp: This makes the check strict so `$val` is not casted in any way.

tests/testOptionSlo.php
```diff
@@ -75,15 +75,21 @@ public function testSloFieldOutput() {
...
+
+		$validated = self::$admin->basic_validation( [], [ 'singlelogout' => '1' ] );
+		$this->assertEquals( true, $validated['singlelogout'] );
+
+		$validated = self::$admin->basic_validation( [], [] );
+		$this->assertEquals( false, $validated['singlelogout'] );
 
 		$validated = self::$admin->basic_validation( [], [ 'singlelogout' => uniqid() ] );
-		$this->assertEquals( 1, $validated['singlelogout'] );
+		$this->assertEquals( false, $validated['singlelogout'] );
```
lbalmaceda: @joshcanhelp if 0 is false and 1 is true, this should be true. What am I not seeing?

joshcanhelp: @lbalmaceda - Validation changed do that all truth-y values are not accepted, only specific truth-y values we expected. Previously, some arbitrary string would set this as `true`

PR #794, opened 2020-05-19
Align the client ID and redirect URI used in the setup wizard

joshcanhelp:
### Description

Fix mis-matched auth code exchange parameters, causing the setup wizard to fail on some sites where the `site_url` does not match the `home_url`.

**Note to reviewers:** Ignore POT, PO, and MO files, minor change in translation

### Testing

- [x] This change adds test coverage for new/changed/fixed functionality

### Checklist

- [x] The correct base branch is being used, if not `master`


templates/initial-setup/end.php
```diff
@@ -7,7 +7,7 @@
 	<p class="a0-step-text"><?php _e( 'Adjust the plugin settings from the WordPress dashboard, and visit the Auth0 dashboard to change how users log in, add connections, enable multi-factor authentication, and more.', 'wp-auth0' ); ?></p>
 
 	<div class="a0-buttons extra-space">
-	  <a href="<?php echo admin_url( 'admin.php?page=wpa0' ); ?>" class="button button-primary"><?php _e( 'GO TO PLUGIN SETTINGS', 'wp-auth0' ); ?></a>
+	  <a href="<?php echo admin_url( 'admin.php?page=wpa0' ); ?>" class="button button-primary"><?php _e( 'Go To Plugin Settings', 'wp-auth0' ); ?></a>
```
joshcanhelp, line 10: Translation change here, caps only

PR #795, opened 2020-05-20
Fix incorrect function in uninstall hook

joshcanhelp:
### Description

Fix an incorrect function loaded in the uninstall hook


joshcanhelp:
@lbalmaceda - There is but this one hook is odd in that it only works when the plugin is removed from the site, not like other hooks.

PR #814, opened 2021-01-25
Support PHP 8.0

evansims:
This PR includes changes necessary to improve compatibility with PHP 8, updates our CircleCI configuration, updates our unit test configuration to reflect and target the minimum PHP and WordPress versions we support, and fixes a number of issues with our documentation around parameter types (they remain functionally unchanged; our docblocks were just incorrect in their existing usage.)

To test these changes locally you will need a local MySQL installation. From the root directory run `bin/install-wp-tests.sh` which will output the proper command syntax. Provide your local database details. For the target WordPress version, specify `5.6` for this PR. After the command finishes, run `composer checkup` to run the test suite.


evansims:
This PR is throwing an erroneous codecov failure for unknown reasons. It is detecting code changes in files that weren't changed. Once this is reviewed I'm going to force merge this and hopefully that will resolve the coverage diff issues.

functions.php
```diff
@@ -153,7 +153,7 @@ function wp_auth0_url_base64_decode( $input ) {
 		$padlen = 4 - $remainder;
 		$input .= str_repeat( '=', $padlen );
 	}
-	return base64_decode( strtr( $input, '-_', '+/' ) );
+	return base64_decode( strtr( $input, '-_', '+/' ) ) ?: false;
```
lbalmaceda, line 156: why is this needed, and what would happen if the caller receives false? The method doc says it returns a string.

lib/WP_Auth0_Import_Settings.php
```diff
@@ -22,12 +22,10 @@ public function import_settings() {
 		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotValidated
 		if ( ! wp_verify_nonce( wp_unslash( $_POST['_wpnonce'] ?? '' ), self::IMPORT_NONCE_ACTION ) ) {
 			wp_nonce_ays( self::IMPORT_NONCE_ACTION );
-			exit;
```
lbalmaceda, line 25: I know `wp_die` will make it quit. But according to the docs, [wp_nonce_ays](https://developer.wordpress.org/reference/functions/wp_nonce_ays/) will prompt for yes/no. If this was previously returning here, why is now this return no longer required?

lib/WP_Auth0_UsersRepo.php
```diff
@@ -132,9 +132,8 @@ public function find_auth0_user( $id ) {
 
 		$users = get_users( $query );
 
-		if ( $users instanceof WP_Error ) {
-			WP_Auth0_ErrorLog::insert_error( __METHOD__ . ' => get_users() ', $users->get_error_message() );
-
+		if ( $users === [] ) {
```
lbalmaceda, line 135: I'm also curious about this change. Where was the default changed?

lib/api/WP_Auth0_Api_Client_Credentials.php
```diff
@@ -67,23 +67,23 @@ protected function handle_response( $method ) {
...
 		// If we have no access token, something went wrong upstream.
-		if ( empty( $response_body->access_token ) ) {
+		if ( ! isset( $response_body['access_token'] ) ) {
 			WP_Auth0_ErrorLog::insert_error( $method, __( 'No access_token returned.', 'wp-auth0' ) );
 			return self::RETURN_ON_FAILURE;
 		}
 
 		// Set the transient to expire 1 minute before the token does.
-		$expires_in  = ! empty( $response_body->expires_in ) ? absint( $response_body->expires_in ) : HOUR_IN_SECONDS;
+		$expires_in  = ! isset( $response_body['expires_in'] ) ? absint( $response_body['expires_in'] ) : HOUR_IN_SECONDS;
```
lbalmaceda, line 79: I think the conditions were reversed here...

! empty --> not empty === a value is set

! isset --> no value is set === the value is empty

lib/initial-setup/WP_Auth0_InitialSetup_AdminUser.php
```diff
@@ -20,12 +20,10 @@ public function callback() {
 		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotValidated
 		if ( ! wp_verify_nonce( wp_unslash( $_POST['_wpnonce'] ?? '' ), self::SETUP_NONCE_ACTION ) ) {
 			wp_nonce_ays( self::SETUP_NONCE_ACTION );
-			exit;
```
lbalmaceda, line 23: same comment as above

lib/initial-setup/WP_Auth0_InitialSetup_ConnectionProfile.php
```diff
@@ -22,12 +22,10 @@ public function callback() {
 		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotValidated
 		if ( ! wp_verify_nonce( wp_unslash( $_POST['_wpnonce'] ?? '' ), self::SETUP_NONCE_ACTION ) ) {
 			wp_nonce_ays( self::SETUP_NONCE_ACTION );
-			exit;
```
lbalmaceda, line 25: here as well

phpcs.xml.dist
```diff
@@ -30,7 +30,7 @@
     https://github.com/PHPCompatibility/PHPCompatibility
     -->
     <config name="testVersion" value="7.0-"/>
-    <config name="minimum_supported_wp_version" value="4.9"/>
+    <config name="minimum_supported_wp_version" value="5.6"/>
```
lbalmaceda, line 33: how would customers running PHP < 5.6 be affected by this change?

lbalmaceda (changes requested):

functions.php
evansims: You might be looking at the method above it, wp_auth0_url_base64_encode; wp_auth0_url_base64_decode returns [either a bool or a string](https://github.com/auth0/wp-auth0/blob/dd1c1268d50b2fdd20e32cc46fc2896fddd1df40/functions.php#L146). base64_decode will return false when it's passed improperly formatted data. However I should have applied this rule with an extra parameter here, rather than a ternary. Fixed in next commit.

lib/WP_Auth0_Import_Settings.php
evansims: The documentation is a bit confusing on this, but wp_nonce_ays actually [calls wp_die internally too](https://core.trac.wordpress.org/browser/tags/5.6/src/wp-includes/functions.php#L3273). It just invokes some frontend UI for that prompt before it quits. So we can trim those exits safely here.

lib/WP_Auth0_UsersRepo.php
evansims: No defaults were changed to my knowledge, the usage of the WP get_users function has remained unchanged since 3.1 as far as I can tell. I think we just made a mistake in writing our error handling here.

In our usage, [get_users()](https://core.trac.wordpress.org/browser/tags/5.6/src/wp-includes/user.php#L753), which is a wrapper for [WP_User_Query()->get_results()](https://developer.wordpress.org/reference/classes/wp_user_query/#return-values), will only ever return an array. The array returned will be empty if there were no results. 🤷‍♂️

lib/api/WP_Auth0_Api_Client_Credentials.php
evansims: Right you are, good catch! Committing a fix.

phpcs.xml.dist
evansims: No effect on customers as this is just for our internal tests. Our plugin supports WP 5.4.4 and higher presently, so we don't have any customers running 4.X actively. As 4.X has been EOL for several years and we don't have any users on it, we can safely bump our testing to only target modern versions.

evansims:
@lbalmaceda Thanks for your feedback! I addressed your questions and committed a few changes, whenever you get time to re-review. Thanks!

functions.php
```diff
@@ -66,7 +66,7 @@ function wp_auth0_is_current_login_action( array $actions ) {
 /**
  * Generate a valid WordPress login override URL, if plugin settings allow.
  *
- * @param null $login_url - An existing URL to modify; default is wp-login.php.
+ * @param string|null $login_url - An existing URL to modify; default is wp-login.php.
  *
  * @return string
```
cocojoe, line 71: Shouldn't this be a string or false? given you've enabled strict checking. I see you do this for params, so should be consistent.

lib/WP_Auth0_LoginManager.php
```diff
@@ -618,21 +619,18 @@ private function clean_id_token( $id_token_obj ) {
 	/**
 	 * Generate the Auth0 logout URL.
 	 *
-	 * @param null $return_to - Site URL to return to after logging out.
+	 * @param string|null $return_to - Site URL to return to after logging out.
```
cocojoe, line 622: Should this be string|null ? null is the default if no string is passed. The intent is to be passed a string? Maybe this is a PHP thing? as I see it all over the place

cocojoe (commented):
Passing back to Lucho but left a couple of comments.

functions.php
evansims: I know it seems strange, it's just an odd PHPism! Traditionally the community used `null` defaults or to pass undefined params for functions, as opposed expecting bools. I suspect this is because null has long had a performance edge over other falsy values in PHP when it comes to comparisons, but that isn't really the case anymore in modern versions

Now that typehinting has been added to the language and people are being a bit more thoughtful about their API desig there's a lot less of this happening, but unfortunately I can't add typehinting to this library quite yet as it would be a breaking change

In these particular cases, since we're not doing any comparisons and any falsy value works, I could just default these to empty strings to simply our docblocks for now

lib/WP_Auth0_LoginManager.php
evansims: We accept null in this case (although any falsy value would technically work, null is just the standard) because in the event none is provided we'll just [use the value returned from the home_url() function](https://github.com/auth0/wp-auth0/pull/814/files/6a54147cd9959efaf87dfd9a380129928459cc79#diff-8b3a370bef1a7697a84121a43cc266867da01dd4c58ec61c238fea8157237126R633)

lbalmaceda (approved):

PR #816, opened 2021-02-02
Fix enqueued scripts

szepeviktor:
- deps is an array
- do not add WP core version to lock.js URL

### Description

This is PHPStan Level 5


evansims (approved):
LGTM! 👍 Thanks again @szepeviktor

szepeviktor:
[**off**] @evansims Does it ring any bell?

```php
(new Script($javascriptUrl))->enqueue();
(new Script($javascriptUrl))->setDeps(['inertia'])->setVer(self::SCRIPT_VERSION)->enqueue();
```

making our daily life OOP :) https://github.com/szepeviktor/SentencePress/blob/main/src/Script.php#L23-L26
Would you like to contribute?

PR #817, opened 2021-02-03
Pass user info to auth0_before_login action

nicecatch:
There are situations in which we want to check whether the user is authorized to access a wordpress website based on the info passed by Auth0 (contained in app_metadata or user_metadata)

This PR aim is for this info to be available in the auth0_before_login action

#### Why not use auth0_user_login?

Because at that point the user is already logged in. 
Letting the user login and then rejecting him is not ideal, especially if we want to show a custom error page

### DIFF
--- a/examples/auth0_before_login.php
+++ b/examples/auth0_before_login.php
@@ -7,9 +7,11 @@
  * @see WP_Auth0_LoginManager::do_login()
  *
  * @param WP_User $user - WordPress user object.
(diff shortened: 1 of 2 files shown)


evansims (approved):

evansims:
Hey, @nicecatch! 👋  This looks like a good addition to me; thanks for your contribution! Appreciate it 👍

PR #831, opened 2021-06-24
Skip email sync when update is coming from Auth0

drobin03:
### Description

When a user's email is updated in Auth0 and they log in on wordpress, the new email doesn't get saved to wp_users. The code I'm modifying here is meant to save that update, and does, but it is reverted by the `wp_auth0_profile_change_email` hook because it is unable to apply that change to Auth0 (due to our app not having management api access I presume).

Since this particular `wp_update_user` call is always called with fresh data from the Auth0 ID token, I don't see a reason we should be attempting to sync with Auth0 afterwards. My code change here temporarily removes the `wp_auth0_profile_change_email` hook to allow us to save the new email address, and then re-enables the hook after it is finished.

I discovered this issue on a clients website when our Auth0 instance stopped including the email in the ID token (due to some account changes). This resulted in a number of users on our site being created with an email following this format `change_this_email@<uniqid> .com` in `wp_users`. Once we resolved the issue with the ID token, I assumed that the next time each user logged in, their email address would be updated to the correct email from the ID token. However when I tested it, I noticed that the email change would save, and then immediately get reverted. With this fix in place locally, the email saved as-expected, and my user accounts were up-to-date.

### Testing

1. With a basic WP site, using the embedded lock form (not universal login), and Auto-provisioning turned on.
2. Log in with an Auth0 user for the first time, verify that the user was created in wp_users
3. Change the user_email value from the wp_users record to something different
4. Log in as the same user again, verify that the user_email was not updated

### DIFF
--- a/lib/WP_Auth0_LoginManager.php
+++ b/lib/WP_Auth0_LoginManager.php
@@ -313,6 +313,9 @@ class WP_Auth0_LoginManager {
 					}
 				}
 
+				// Temporarily disable the email sync, since the changes are coming from Auth0, no need to update them there.
+				remove_action( 'profile_update', 'wp_auth0_profile_change_email', 100 );
+
 				wp_update_user(
 					(object) [
 						'ID'          => $user->data->ID,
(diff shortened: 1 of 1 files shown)


evansims:
Hey, @drobin03 👋 Thanks very much for the PR! Give me a little to review this and I'll circle back on this with you

Issue #910, opened 2024-07-25
Auth0 error message issue


PeterGFernandez:
### Checklist

- [X] I have looked into the [Readme](https://github.com/auth0/wordpress#readme) and the [documentation](https://auth0.com/docs/customize/integrations/cms/wordpress-plugin), and have not found a suitable solution or answer.
- [X] I have searched the [issues](https://github.com/auth0/wordpress/issues) and have not found a suitable solution or answer.
- [X] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [X] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

As shown in the following screenshot, whenever I get an authentication error with 4.6.2 of the Auth0 Plugin for WordPress, it looks like the HTML for the Login link is being rendered as text (rather than HTML). 
![image (2)](https://github.com/user-attachments/assets/01f47921-9714-4677-8237-d21a4cc97caf)

### Reproduction

Cause an error message to be generated during login via Universal Login

### wp-auth0 version

4.6.2

### WordPress version

6.6.1

### PHP version

8.2.16

hncvj:
I'm having the same issue.

wilmersondasilva:
Same issue here

wilmersondasilva:
I just created a PR to fix this issue: https://github.com/auth0/wordpress/pull/912

I hope it's a good solution and it gets merged.

martin-mith:
Hello.  Is there any visibility on time to resolution?  Thank you.

kishore7snehil:
Hi @PeterGFernandez, thanks for reporting this!

Just a heads-up — we’ve officially stopped supporting the `4.x `series of the Auth0 WordPress plugin. We recommend upgrading to the latest `5.x` version, which includes significant improvements and active maintenance.

We’d be happy to help if you run into anything post-upgrade.

Thanks again!

PR #912, opened 2024-08-02
Fix error message sanitization

wilmersondasilva:
### Description

After upgrading from 4.6.0 to 4.6.2 the error messages have started to be interpreted as text instead of HTML. To the projects that use `auth0_verify_email_page` this can break the entire error page. This bug was reported on issue #910. 

My solution was to use `wp_kses_post` instead of just `wp_kses` to sanitize the error messages. This allows the projects to use any HTML tag already allowed in any WordPress post, so it's a safe way to sanitize the error messages.

### References

Issue #910 

### Testing

You can test it by generating authentication errors and checking if the error message is correctly interpreted as HTML instead of text.

### Checklist

- [ ] All active GitHub checks for tests, formatting, and security are passing
- [ ] The correct base branch is being used, if not the default branch

### DIFF
--- a/lib/WP_Auth0_Email_Verification.php
+++ b/lib/WP_Auth0_Email_Verification.php
@@ -51,6 +51,6 @@ class WP_Auth0_Email_Verification
 
 		$html = apply_filters('auth0_verify_email_page', $html, $userinfo, '');
 
-		wp_die(wp_kses($html, ['p' => [], 'a' => ['id' => true, 'href' => true], 'script' => ['src' => true]]));
+		wp_die(wp_kses_post($html));
 	}
 }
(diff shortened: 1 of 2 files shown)


martin-mith:
Hello - can we get someone assigned to review the issue?  Thank you.

robbytx:
I'm not sure this is going to get any traction, but I've observed the same problem, and I think the culprit is actually the `esc_html($html)` call [here](https://github.com/auth0/wordpress/blob/4.x/lib/WP_Auth0_LoginManager.php#L617), since that is taking the existing HTML markup and escaping it as `&lt;`.

So I'm not sure that this PR actually solves the problem identified in #910.

kishore7snehil:
Closing this out as it targets the `4.x` branch, which we're no longer actively maintaining.

PR #914, opened 2024-08-22
Do not update_option for DB version on every request

kadamwhite:
> By submitting a PR to this repository, you agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md). Please see the [contributing guidelines](https://github.com/auth0/.github/blob/master/CONTRIBUTING.md) for how to create and submit a high-quality PR for this repo.

(👍)

### Description

The install_db function is hooked into the plugins loaded action, but includes an `update_option` call which is fired even when the stored option DOES match.

This is inefficient, and can cause unnecessary DB write load. It is especially problematic on some enterprise hosting where an UPDATE call causes all subsequent reads in a request to go through a Writer database instead of a read replica, overwhelming production DB servers.

### References

Example query monitor output demonstrating UPDATE call on average pageview

image

### Testing

Using Query Monitor, inspect the DB requests triggered by frontend traffic on a site. There should not be an identical UPDATE call on every request.

### Checklist

- [ ] All active GitHub checks for tests, formatting, and security are passing

### DIFF
--- a/lib/WP_Auth0_DBManager.php
+++ b/lib/WP_Auth0_DBManager.php
@@ -19,7 +19,11 @@ class WP_Auth0_DBManager
 			$current_ver = (int) get_site_option('auth0_db_version', 0);
 		}
 
-		if (empty($current_ver) || $current_ver === AUTH0_DB_VERSION) {
+		if ($current_ver === AUTH0_DB_VERSION) {
+			return;
+		}
+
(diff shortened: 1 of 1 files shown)


kishore7snehil:
Closing this out as it targets the `4.x` branch, which we're no longer actively maintaining.

Issue #934, opened 2025-06-30
Rolling Sessions still causing nonce invalidation in plugin version 5.3.0


dennis-n-dean:
### Checklist

- [x] I have looked into the [Readme](https://github.com/auth0/wordpress#readme) and the [documentation](https://auth0.com/docs/customize/integrations/cms/wordpress-plugin), and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/wordpress/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

Having Rolling Sessions enabled still appears to be causing nonce invalidation issues in plugin version 5.3.0.

With Rolling Sessions enabled:
- calls from the admin area/Gutenberg to wp-json return 403, making it impossible to edit featured images, category and tags, etc.
- things that rely on the heartbeat, like post check in/check out or taking over a post a colleague is editing break
- deleting posts, activating/deactivating plugins, etc. causes a "this link has expired" error

Disabling rolling sessions appears to fix all these problems (but I assume will cause problems of its own). 

### Reproduction

1. Install Auth0 plugin
2. Leave "Rolling Sessions" enabled
3. Observe 

### wp-auth0 version

5.3.0

### WordPress version

6.8.1

### PHP version

8.4

kishore7snehil:
Hi @dennis-n-dean 👋 , thanks for the detailed repro steps.

To zero in on this, could you also share:

1. Any caching layers you have in front of WP (e.g. Varnish, Cloudflare, WP Super Cache, object cache, etc.)
2. A screenshot or HAR of one failing request in the admin/Gutenberg that shows the X-WP-Nonce header and the 403 response
3. A screenshot of your Auth0 plugin settings page, specifically which Rolling Sessions options you have toggled

Once we’ve got that, we can confirm if it’s purely the rolling-session cookie vs. nonce window mismatch or if another cache layer is exacerbating it—and then recommend the precise tweak. Thanks!

dennis-n-dean:
Hi @kishore7snehil, finally getting a chance to circle back around to this. 

[...]

PR #947, opened 2026-02-24
fix: Wire up Device Storage Method session setting to SDK configuration

Flowgram:
## Summary

The **Device Storage Method** setting in the plugin UI (Advanced > Sessions) allows administrators to choose between "Encrypted Cookies" and "PHP Native Sessions (Recommended)". However, `importConfiguration()` in `Plugin.php` never reads this option - the SDK always defaults to `CookieStore` regardless of the selected setting.

This PR fixes two related bugs:

- **`Plugin.php`**: Read the `sessions.method` option and configure a `SessionStore` on the `SdkConfiguration` when "PHP Native Sessions" is selected
- **`Authentication.php`**: Guard the `setState()` call in `onShutdown()` with an `instanceof CookieStore` check, since `setState()` is `CookieStore`-specific and causes a fatal error (`Call to undefined method Auth0\SDK\Store\SessionStore::setState()`) when `SessionStore` is in use

## Context

When the plugin uses `CookieStore` (the effective default regardless of UI setting), the entire user session - ID token, access token, refresh token, and user profile - is serialized and encrypted into cookies. These get chunked across multiple cookies (`auth0_session_0`, `auth0_session_1`, `auth0_session_2`, etc.) which, combined with standard WordPress auth cookies, can exceed nginx/server default header buffer limits (typically 8KB), resulting in:

```
400 Bad Request
Request Header Or Cookie Too Large
```

Switching to PHP Native Sessions (as the UI already advertises) stores session data server-side and only sends a small session ID cookie, completely avoiding this issue.

### `src/Plugin.php`
- Added `use Auth0\SDK\Store\SessionStore` import
- After creating the `SdkConfiguration`, check if `sessions.method` is `'sessions'` and call `setSessionStorage()` with a new `SessionStore` instance

### `src/Actions/Authentication.php`
- Replaced the `@var CookieStore` type annotation + unchecked `setState()` call with a proper `instanceof CookieStore` guard

## Test plan

- [ ] With **Device Storage Method** set to "Encrypted Cookies": verify login works as before (cookies are set)
- [ ] With **Device Storage Method** set to "PHP Native Sessions (Recommended)": verify login works and no `auth0_session_*` cookies are set
- [ ] With **Rolling Sessions** enabled and **PHP Native Sessions** selected: verify no fatal error on `setState()`
- [ ] With **Rolling Sessions** enabled and **Encrypted Cookies** selected: verify `setState()` still functions correctly


src/Plugin.php
```diff
@@ -319,6 +320,10 @@ private function importConfiguration(): SdkConfiguration
             );
         }
 
+        if ('sessions' === $this->getOptionString('sessions', 'method')) {
+            $sdkConfiguration->setSessionStorage(new SessionStore($sdkConfiguration, $sdkConfiguration->getSessionStorageId()));
```
kishore7snehil, line 324: **SessionStore** also doesn't call `session_write_close()` anywhere, and WordPress doesn't either. PHP sessions lock the session file for the entire request, so concurrent AJAX requests from the same user will queue up and wait for each other instead of running in parallel. Might cause noticeable slowness in the admin.

src/Plugin.php
```diff
@@ -319,6 +320,10 @@ private function importConfiguration(): SdkConfiguration
             );
         }
 
+        if ('sessions' === $this->getOptionString('sessions', 'method')) {
```
kishore7snehil, line 323: Have we checked how this behaves on managed WordPress hosts (WP Engine, Kinsta, Pantheon etc.) SessionStore calls `session_start()` internally, and a lot of these hosts strip or ignore PHP session cookies at the CDN/server level. That would make SessionStore silently non-functional.

src/Actions/Authentication.php
```diff
@@ -579,10 +579,9 @@ public function onShutdown(): void
...
 
-            /**
-             * @var CookieStore $store
-             */
-            $store->setState(true);
+            if ($store instanceof CookieStore) {
+                $store->setState(true);
+            }
 
             wp_set_auth_cookie(get_current_user_id(), true);
```
kishore7snehil, line 586: The `instanceof` guard above fixes the fatal, but this line still breaks rolling sessions for both storage methods. This needs the fix from [#948](https://github.com/auth0/wordpress/pull/948) (passing the existing token) to actually work.

kishore7snehil (changes requested):

src/Plugin.php
Flowgram: That's exactly what happens. Visiting an AJAX-heavy page in the admin causes my local environment to crash with ERR_CONNECTION_REFUSED errors, requiring a server restart. My local dev uses PHP's default file-based session handler, which is where the locking is most severe. Hosts using Redis or database-backed sessions would fare better, but it does highlight wider issues with the SessionStore approach as-is.

src/Plugin.php
Flowgram: Valid point, but the plugin already presents this option in the UI with a warning: "PHP Sessions require external configuration to work securely and reliably." This PR just makes that existing UI choice functional, admins choosing it should already be aware of their hosting constraints. 

Worth noting that the plugin UI actually labels this option "PHP Native Sessions (Recommended)", so there's arguably a broader question about whether that recommendation is appropriate given these hosting caveats.

src/Actions/Authentication.php
Flowgram: Agreed, the wp_set_auth_cookie() nonce rotation issue is real and affects both storage methods. However, that's a pre-existing bug on the current codebase (tracked in #934), not something introduced by this PR. The scope here is specifically wiring up the dead session storage setting. Happy to see #948 addressing the rolling sessions fix separately.

src/Plugin.php
kishore7snehil: Since PHP's default file-based session handler is what most self-hosted WordPress installations use, and you've confirmed it causes server crashes on AJAX-heavy admin pages, I don't think we should ship this without at least warning admins at runtime.

kishore7snehil: Could you add an admin notice when "PHP Native Sessions" is selected? Something along the lines of:

> "You have selected PHP Native Sessions for Auth0 session storage. The default PHP file-based session handler can cause performance issues with concurrent requests. For production use, configure a non-blocking session handler (Redis, Memcached, or database-backed) via session.save_handler in your PHP configuration."

  This way we're not blocking the feature, but we're making sure admins don't hit the locking wall without any heads-up beyond the existing generic warning.

Flowgram: Hi @kishore7snehil, I haven't had a chance to look into this in depth again, especially given the wider issues with SessionStore (I feel this requires a solid fix instead of just an admin notice).

Is there any chance of your team taking ownership of this fix/PR and also the other [open PR](https://github.com/auth0/wordpress/pull/948)? 

Perhaps they could be rolled into the upcoming v6 release?

PR #948, opened 2026-03-10
Reuse existing session token when refreshing sessions

goldenapples:
Fixes a bug when the rolling sessions functionality would update a user's session token on every pageview, invalidating any nonce checks before they're run. See #934 for examples of the issues this caused end users.

The root cause of the issue is:
* this plugin calls `wp_set_auth_cookie()` on a shutdown hook of every pageview, which causes a new session token to be issued for the user
* Ajax functionality triggered from the page will be passing a nonce which is generated from the previous session token.
* Any wp-ajax or REST handlers responding to these request will be checking the old nonce against a value generated from the user's new session token, causing any actions that depend on CSRF check to fail.

This change preserves the existing session token by passing it in to wp_set_auth_cookie(). In addition, the lifetime of the token itself needs to be extended so that it doesn't expire before the logged-in cookies.

### Checklist

- [x] I have added documentation for new/changed functionality in this PR or in auth0.com/docs

### DIFF
--- a/src/Actions/Authentication.php
+++ b/src/Actions/Authentication.php
@@ -579,12 +579,37 @@ final class Authentication extends Base
         if ('false' !== $this->getPlugin()->getOption('sessions', 'rolling_sessions')) {
             $store = $this->getSdk()->configuration()->getSessionStorage();
 
-            /**
-             * @var CookieStore $store
-             */
-            $store->setState(true);
+            if ($store instanceof CookieStore) {
+                $store->setState(true);
+            }
 
-            wp_set_auth_cookie(get_current_user_id(), true);
+            // Extend the existing session token's expiry and reissue the browser cookies.
+            //
+            // The naive approach — calling wp_set_auth_cookie() with no token — rotates the
+            // session token. Because nonces are cryptographically bound to the session token,
+            // any nonce generated during the current request (with the old token) becomes
+            // invalid the moment the new cookie is sent, causing REST requests to fail with
+            // rest_cookie_invalid_nonce.
+            //
+            // wp_set_auth_cookie() accepts a $token argument. Passing the existing token
+            // reissues the browser cookies with a fresh expiration without rotating the token,
+            // so nonces remain valid and the full rolling-session behaviour is preserved.
+            $userId = get_current_user_id();
+            $token  = wp_get_session_token();
+
+            if ('' !== $token) {
+                /** This filter is documented in wp-includes/pluggable.php */
+                $expiration = apply_filters('auth_cookie_expiration', 14 * DAY_IN_SECONDS, $userId, true);
+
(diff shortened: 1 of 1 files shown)


src/Actions/Authentication.php
```diff
@@ -579,12 +579,37 @@ public function onShutdown(): void
...
+            //
+            // wp_set_auth_cookie() accepts a $token argument. Passing the existing token
+            // reissues the browser cookies with a fresh expiration without rotating the token,
+            // so nonces remain valid and the full rolling-session behaviour is preserved.
+            $userId = get_current_user_id();
+            $token  = wp_get_session_token();
+
+            if ('' !== $token) {
+                /** This filter is documented in wp-includes/pluggable.php */
+                $expiration = apply_filters('auth_cookie_expiration', 14 * DAY_IN_SECONDS, $userId, true);
```
kishore7snehil, line 602: The **14 * DAY_IN_SECONDS** default mirrors what WordPress core uses inside `wp_set_auth_cookie()`, but it's an implicit coupling. If core ever changes that default, these two would diverge.

src/Actions/Authentication.php
```diff
@@ -579,12 +579,37 @@ public function onShutdown(): void
...
+
+                // Update the server-side session record (wp_set_auth_cookie does not do this
+                // when a token is supplied).
+                \WP_Session_Tokens::get_instance($userId)->update($token, [
```
kishore7snehil: `WP_Session_Tokens::update()` merges into existing session data. Just confirming , is there any session metadata beyond expiration that should also be refreshed for rolling sessions?

kishore7snehil (changes requested):

src/Actions/Authentication.php
goldenapples: Oh, apparently I misunderstood how the token manager handles updates. Actually, from what I see, `WP_Session_Tokens::update()` overwrites all the data attached to the session, which would wipe out useful audit data like IP address, user agent, and original login date; as well as any custom data attached with the `attach_session_information` filter.

I think it's better here to read the existing token and only update the "expiration" field on it. Updated in [dfbde56](https://github.com/auth0/wordpress/pull/948/commits/dfbde56277ce99b95a345ca8cb6981e6558904df).

src/Actions/Authentication.php
goldenapples: True. I'm not sure what the best default value to use here is anyway, so I just went with the logic from core. Maybe the default should match the Auth0 session expiration lifetime, since we have access to that?

Regardless, sites can modify this using the core "auth_cookie_expiration" filter.

roborourke: The likelihood of core changing that default is very very low, and if you have to choose some other value regardless then you've already diverged. Given there are settings for this in the admin anyway it doesn't feel like this should be a blocker to me at least.

PR #956, opened 2026-06-14
feat: auth0-php v9 compatibility (6.x beta groundwork)

kishore7snehil:
# Auth0 WordPress Plugin - auth0-php v9 compatibility

This PR moves the plugin onto [auth0-php v9](https://github.com/auth0/auth0-php) and lands the groundwork for the upcoming **6.x** major. It targets the `6.x` mainline branch. Version bumps (`.version`, `wpAuth0.php`, CHANGELOG) are intentionally **not** included here - those ship in a separate `release/6.0.0-beta.0` PR, mirroring the auth0-php release model.

## What's New

- **auth0-php v9**: the plugin now requires `auth0/auth0-php` `^9.0@beta`, whose Management API is generated from Auth0's OpenAPI specs (typed requests/responses, property-access sub-clients).
- **`Plugin::getManagement()`**: a new accessor that returns a v9 Management client built from the plugin's existing Domain / Client ID / Client Secret configuration. It fetches and caches a client-credentials token for you.
- **Background user sync on v9**: the WordPress→Auth0 sync (create / update / delete, with password-change and email-verification tickets) is rebuilt onto the v9 typed Management API.
- **PHP 8.2+**: the supported PHP floor is raised to 8.2.

## Breaking Changes

| Area | 5.x | 6.x |
|------|-----|-----|
| **auth0/auth0-php** | `^8.19` | `^9.0` |
| **PHP** | `^8.1` | `>=8.2` |
| **Management access** | `getSdk()->management()` | `getManagement()` |
| **Management responses** | raw PSR-7 | typed objects |

The v8-style `getSdk()->management()` is non-functional in auth0-php v9. Use `getManagement()` instead.

## What's NOT Affected

**Authentication flows are unchanged.** Login, callback, logout, session handling, and the WordPress login fallback behave exactly as they did in 5.x. The v9 migration is confined to the Management API surface (used by background user sync).

## Documentation

- `README.md`: PHP 8.2 requirement, refreshed feature overview (authentication, account mapping, background sync).
- `UPGRADING.md`: a new "5.x to 6.x" section covering the dependency bump and the Management API migration.

## Beta Notice

This is groundwork for a beta release. The authentication surface is stable; the Management API in auth0-php v9 is itself in beta and minor adjustments may occur before the final 6.0.0.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,0 +1,48 @@
+# How to upgrade
+
+## 5.x to 6.x
+
+The 6.0 release moves the plugin onto [auth0-php v9](https://github.com/auth0/auth0-php/tree/v9), which rewrites the Management API. The authentication surface (login, logout, callback, session handling) is unchanged, so most sites only need the environment updates below.
+
+Check that your environment is compatible with 6.0's requirements before upgrading:
+
+| | 5.x | 6.x |
+|---|---|---|
+| **PHP** | `^8.1` | `^8.2` |
+| **auth0/auth0-php** | `^8.19` | `^9.0` |
+
+- Please ensure you are running PHP 8.2 or newer. Support for PHP 8.1 has been dropped.
+- The plugin bundles its dependencies, so updating the plugin pulls in `auth0/auth0-php` v9 automatically. No package rename is required; the package remains `auth0/auth0-php`.
+
+Update your customizations, if necessary:
+
+- **No changes are required for the plugin's built-in features.** Login, logout, the callback handler, user session handling, and the User Sync background jobs continue to work as they did in 5.x. The settings stored in the WordPress admin are unchanged.
+- **Custom code that calls the Management API has changed.** In v9 the old `wpAuth0()->getSdk()->management()` entry point is non-functional and will throw a `TypeError`. Use the new `wpAuth0()->getManagement()` accessor instead. It builds a Management client from the Domain, Client ID, and Client Secret you already configure in the plugin settings, and fetches and caches a client credentials token for you automatically.
+
+  ```php
+  // 5.x
+  $management = wpAuth0()->getSdk()->management();
+  $response = $management->users()->getAll(['per_page' => 25]);
+  $users = HttpResponse::decodeContent($response);
+
+  // 6.x
+  use Auth0\SDK\API\Management\Users\Requests\ListUsersRequestParameters;
+
+  $management = wpAuth0()->getManagement();
+  $users = $management->users->list(
+      new ListUsersRequestParameters(['perPage' => 25, 'includeTotals' => true])
+  );
+  foreach ($users as $user) {
+      echo $user->getEmail();
+  }
+  ```
+
+  Inside a plugin action or filter class (anything extending `Auth0\WordPress\Actions\Base` or `Auth0\WordPress\Filters\Base`), the same client is available as `$this->getManagement()`.
+
+- If your custom code calls the Management API directly, review the [auth0-php v9 migration guide](https://github.com/auth0/auth0-php/blob/v9/v9_MIGRATION_GUIDE.md) for the full set of changes. The most common adjustments are:
+  - Sub-resources are reached by property access, not method calls: `->users->list()` rather than `->users()->getAll()`.
+  - Responses are typed objects instead of raw PSR-7 responses. Call `$response->jsonSerialize()` to get the same snake_case array the v8 `HttpResponse::decodeContent()` returned, or use the typed getters (`$response->getEmail()`).
+  - Errors throw exceptions. A non-2xx response raises `Auth0\SDK\API\Management\Exceptions\Auth0ApiException` (use `getCode()` for the HTTP status), and transport-level failures raise `Auth0\SDK\API\Management\Exceptions\Auth0Exception`. There is no more `HttpResponse::wasSuccessful()` check.
+  - Request parameters are camelCase typed objects. For example, creating a user takes a `CreateUserRequestContent` whose keys are `givenName` and `familyName` (not `given_name` / `family_name`), and the `connection` is set inside that object rather than passed as a separate argument.
+  - Listing users only returns the paginated envelope when `includeTotals` is set to `true`. Omitting it yields an empty result, so pass `'includeTotals' => true` when you page through users.
+  - Some ticket methods were renamed: `tickets()->createPasswordChange()` is now `tickets->changePassword()`, and `tickets()->createEmailVerification()` is now `tickets->verifyEmail()`. The user id moves inside the request object as `userId`.
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -16,7 +16,8 @@ Check that your environment is compatible with 6.0's requirements before upgradi
 
 Update your customizations, if necessary:
 
-- **No changes are required for the plugin's built-in features.** Login, logout, the callback handler, user session handling, and the User Sync background jobs continue to work as they did in 5.x. The settings stored in the WordPress admin are unchanged.
+- **No changes are required for the plugin's built-in features.** Login, logout, the callback handler, and user session handling continue to work as they did in 5.x. The settings stored in the WordPress admin are unchanged.
+- **User Sync now retries a failed event instead of always discarding it.** In 5.x a sync event that failed at the Management API was dropped from the queue regardless of the reason. In 6.x, a transient failure (an HTTP 429 rate limit, a 5xx server error, or a network error) keeps the event in the queue so the next cron pass retries it. Permanent failures (such as an HTTP 400 from an invalid profile) are still dropped, since retrying them would never succeed. No configuration change is needed, but a queue that previously drained to empty on every pass may now hold a retrying event until it succeeds.
 - **Custom code that calls the Management API has changed.** In v9 the old `wpAuth0()->getSdk()->management()` entry point is non-functional and will throw a `TypeError`. Use the new `wpAuth0()->getManagement()` accessor instead. It builds a Management client from the Domain, Client ID, and Client Secret you already configure in the plugin settings, and fetches and caches a client credentials token for you automatically.
 
   ```php
@@ -41,8 +42,9 @@ Update your customizations, if necessary:
 
 - If your custom code calls the Management API directly, review the [auth0-php v9 migration guide](https://github.com/auth0/auth0-php/blob/v9/v9_MIGRATION_GUIDE.md) for the full set of changes. The most common adjustments are:
   - Sub-resources are reached by property access, not method calls: `->users->list()` rather than `->users()->getAll()`.
+  - `users->list()` returns a `Pager` you iterate with `foreach`. It is a lazy iterator that makes further requests as you consume it, so `count()` and array access will not behave the way a plain array would.
   - Responses are typed objects instead of raw PSR-7 responses. Call `$response->jsonSerialize()` to get the same snake_case array the v8 `HttpResponse::decodeContent()` returned, or use the typed getters (`$response->getEmail()`).
   - Errors throw exceptions. A non-2xx response raises `Auth0\SDK\API\Management\Exceptions\Auth0ApiException` (use `getCode()` for the HTTP status), and transport-level failures raise `Auth0\SDK\API\Management\Exceptions\Auth0Exception`. There is no more `HttpResponse::wasSuccessful()` check.
   - Request parameters are camelCase typed objects. For example, creating a user takes a `CreateUserRequestContent` whose keys are `givenName` and `familyName` (not `given_name` / `family_name`), and the `connection` is set inside that object rather than passed as a separate argument.
-  - Listing users only returns the paginated envelope when `includeTotals` is set to `true`. Omitting it yields an empty result, so pass `'includeTotals' => true` when you page through users.
+  - The `includeTotals` parameter only controls whether the response envelope carries the count fields (`total`, `start`, `limit`, `length`). It defaults to `true` and does not affect whether users are returned, so you get results either way.
   - Some ticket methods were renamed: `tickets()->createPasswordChange()` is now `tickets->changePassword()`, and `tickets()->createEmailVerification()` is now `tickets->verifyEmail()`. The user id moves inside the request object as `userId`.
--- a/build.sh
+++ b/build.sh
@@ -34,18 +34,24 @@ echo "Version: $version"
 semver "$version" version
 filename="Auth0_WordPress_${version}.zip"
 
+PLUGIN_NAME="wp-auth0"
+
 echo "# Cleaning up environment..."
 rm -f build.zip
 rm -f build.zip.sig
 rm -rf build
+rm -rf ${PLUGIN_NAME}
 rm -rf vendor
-rm composer.lock
+rm -f composer.lock
 
 echo "# Executing Composer..."
 composer update --no-plugins
 
 echo "# Prefixing Dependencies..."
-vendor/bin/php-scoper add-prefix --force
+php -d memory_limit=512M vendor/bin/php-scoper add-prefix --force
+
+echo "# Copying plugin entry file..."
+cp wpAuth0.php build/wpAuth0.php
 
 echo "# Finalizing Build..."
 cd build
@@ -54,8 +60,21 @@ rm composer.json
 rm composer.lock
 cd ..
 
+echo "# Verifying scoped build (checking for incorrectly prefixed WordPress globals)..."
+if grep -rn 'Auth0\\WordPress\\Vendor\\WP_' build/src/; then
+    echo ""
+    echo "ERROR: php-scoper has incorrectly prefixed WordPress global classes!"
+    echo "The above files contain 'Auth0\\WordPress\\Vendor\\WP_*' references that will cause fatal errors."
+    echo "Check scoper.inc.php expose-global-classes setting."
+    exit 1
+fi
+echo "  No incorrectly prefixed WordPress globals found. Build OK."
+
+echo "# Renaming build folder to ${PLUGIN_NAME}..."
+mv build ${PLUGIN_NAME}
+
 echo "# Archiving Build..."
-zip -vr ${filename} build/ -x "*.DS_Store"
+zip -vr ${filename} ${PLUGIN_NAME}/ -x "*.DS_Store"
 
 echo "# Signing Build..."
 openssl dgst -sign private-signing-key.pem -sha256 -out ${filename}.sig -binary ${filename}
--- a/composer.json
+++ b/composer.json
@@ -33,10 +33,11 @@
         "docs": "https://auth0.com/docs/customize/integrations/cms/wordpress-plugin"
     },
     "require": {
-        "php": "^8.1",
+        "php": "^8.2",
         "ext-json": "*",
         "ext-openssl": "*",
-        "auth0/auth0-php": "^8.19",
+        "auth0/auth0-php": "^9.0@beta",
+        "nyholm/psr7": "^1",
         "psr/cache": "^3.0"
     },
     "require-dev": {
@@ -46,7 +47,6 @@
         "friendsofphp/php-cs-fixer": "^3",
         "hyperf/event": "^2",
         "mockery/mockery": "^1",
-        "nyholm/psr7": "^1",
         "pestphp/pest": "^2",
         "phpstan/phpstan": "^1",
         "phpstan/phpstan-strict-rules": "^1",
--- a/scoper.inc.php
+++ b/scoper.inc.php
@@ -8,10 +8,18 @@ return [
     'prefix' => 'Auth0\\WordPress\\Vendor',
 
     'finders' => [
+        // Allowlist the plugin's own source; root files ship via append() below.
         Finder::create()
             ->files()
             ->ignoreVCS(true)
-            ->notName('/.*\\.dist|Makefile|scoper.inc.php|rector.php|opslevel.yml|build.sh|public-signing-key.pub|composer.json|composer.lock/')
+            ->in('src'),
+
+        // jetbrains/phpstorm-stubs declares PHP internals (incl. assert()) as
+        // global functions; scoping them emits an illegal assert() wrapper that
+        // fatals scoper-autoload.php. Dev-only, stripped by --no-dev regardless.
+        Finder::create()
+            ->files()
+            ->ignoreVCS(true)
             ->exclude([
                 'doc',
                 'test',
@@ -19,11 +27,18 @@ return [
                 'tests',
                 'Tests',
                 'vendor-bin',
+                'jetbrains',
             ])
-			->in(['vendor', '.']),
+            ->in('vendor'),
 
+        // Shippable root files (composer.json drives autoloading; build.sh strips it).
         Finder::create()->append([
             'composer.json',
+            'CHANGELOG.md',
+            'LICENSE.md',
+            'README.md',
+            'UPGRADING.md',
+            'updates.json',
         ]),
     ],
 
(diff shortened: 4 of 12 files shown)


scoper.inc.php
```diff
@@ -33,9 +48,46 @@
...
+        '/^register_/',
+        '/^settings_/',
+        '/^update_/',
+        '/^delete_/',
+        '/^set_/',
+    ],
+
+	'expose-global-constants' => true,
+	'expose-global-classes'   => true,
+	'expose-global-functions' => true,
```
nandan-bhat: This is the change I'd most like to talk through before merge. I think these three flags may not be doing what we want.

My understanding is that `expose-global-*` isn't "leave WordPress symbols alone". It tells php-scoper to take every global symbol the scoped code declares and register it back into the global namespace through `scoper-autoload.php`. Since we bundle `auth0-php` v9, that pulls in `nyholm/psr7`, `php-http/discovery`, `psr-discovery/all` and `php-http/multipart-stream-builder`, and exposing those globally would undo the prefixing we're doing them for. If another plugin on the same site ships a different `Nyholm\Psr7`, both would be reaching for the same global alias. That's the kind of clash prefixing is meant to prevent, and WordPress sites tend to have a lot of plugins.

The encouraging part is that you already added the mechanism that does what I think you're after: `exclude-classes` and `exclude-functions` just below. Those work independently of `expose-global-*`, so they should be enough on their own to protect the WordPress API.

Would setting all three back to `false` and relying on the exclude lists work? If so, please read this together with my note on `exclude-functions`, since I think the two changes need to go in together.

Usual caveat: I couldn't run `./build.sh`, so please confirm against a real build rather than taking my word for it.

scoper.inc.php
```diff
@@ -33,9 +48,46 @@
...
-	'expose-global-classes'   => false,
-	'expose-global-functions' => false,
+    'exclude-classes' => [
+        '/^WP_/',
+        'wpdb',
+        'Walker',
+        'WP_Widget',
+    ],
+
+    'exclude-functions' => [
```
nandan-bhat, line 58: Following on from my note about the `expose-global-*` flags, this is why I think the two need to change together.

I pulled the distinct function calls out of `src/` (162 of them) and compared them against these patterns. A few WordPress calls don't seem to be covered:

- `maybe_create_table` in `src/Database.php:108` and `:128`
- `submit_button` in `src/Actions/Configuration.php:1020`, `:1031`, `:1042`
- `nocache_headers` in `src/Actions/Authentication.php:478`
- `checked` in `src/Utilities/Render.php:109`

As far as I can tell these are fine today only because `expose-global-functions` is `true`, which papers over the gap. If that flag goes back to `false` on its own, I'd expect these to become `Auth0\WordPress\Vendor\maybe_create_table()` and similar, which would mean errors on table creation and on rendering the settings screens. So it would be good to extend this list in whichever commit adjusts the flags.

Smaller thought while you're in here: a few of these patterns reach further than they probably need to. `/^is_/`, `/^get_/`, `/^set_/` and `/^__/` will also match PHP and vendor internals like `is_array`, `get_class` and `__construct`, so we end up skipping prefixing on things unrelated to WordPress. Harmless as far as I can see, just a bit loose. An explicit symbol list, or the set from `php-stubs/wordpress-stubs`, would be more predictable than prefix regexes if you think it's worth the churn.

build.sh
```diff
@@ -54,8 +60,21 @@ rm composer.json
 rm composer.lock
 cd ..
 
+echo "# Verifying scoped build (checking for incorrectly prefixed WordPress globals)..."
+if grep -rn 'Auth0\\WordPress\\Vendor\\WP_' build/src/; then
```
nandan-bhat, line 64: I like the intent behind adding this guard. I think the current form can't fire though, which would make it print `Build OK` regardless.

`scoper.inc.php` sets `exclude-namespaces` to `/^Auth0\\WordPress\\/`, so our own `src/` never gets prefixed. That means nothing under `build/src/` should ever contain `Auth0\WordPress\Vendor\WP_`. The grep is pointed at the one directory that's clean by design.

Two smaller things while we're here:

- It looks for classes (`WP_*`) but not functions. Going by the `exclude-functions` gap I mentioned in `scoper.inc.php`, wrongly prefixed functions seem like the more likely failure.
- The check reads `build/src/` and `mv build ${PLUGIN_NAME}` runs just below. That's fine as ordered, it just means this block has to stay above the `mv` and nothing in the file says so. A short comment would save someone a confusing afternoon later.

Would pointing it at the scoped vendor tree and the generated `vendor/scoper-autoload.php` work better, covering functions too? Something like:

```bash
grep -rnE 'Auth0\\WordPress\\Vendor\\(WP_|wp_|maybe_create_table|submit_button|nocache_headers|checked)' wp-auth0/vendor/
```

A small step that PHP lints the built plugin and checks `scoper-autoload.php` parses would go even further. My only real concern with the current version is that a green message here could give a release engineer more confidence than it's earned.

composer.json
```diff
@@ -33,10 +33,11 @@
         "docs": "https://auth0.com/docs/customize/integrations/cms/wordpress-plugin"
     },
     "require": {
-        "php": "^8.1",
+        "php": "^8.2",
         "ext-json": "*",
         "ext-openssl": "*",
-        "auth0/auth0-php": "^8.19",
+        "auth0/auth0-php": "^9.0@beta",
```
nandan-bhat, line 39: Happy for this to stay as is for the beta. Flagging it mainly so it's on the radar before a tagged release.

We require `^9.0@beta` here, `composer.lock` isn't committed, and `build.sh` runs `rm -f composer.lock` then `composer update` before `php-scoper` and before the `openssl dgst -sign` step. So the zip we sign isn't reproducible: two builds from the same git SHA could bundle different `auth0-php` beta builds, along with different transitive `nyholm/psr7` or `php-http/discovery` versions.

For a plugin in the auth path that ships as a signed artifact, being unable to reproduce what we signed makes it hard to audit after the fact, and it means a yanked or bad upstream beta could get picked up without us noticing.

Some options if you agree it's worth doing: commit `composer.lock`, switch `build.sh` to `composer install` so the lock is authoritative, and pin to an exact version like `9.0.0-beta.1` for tagged releases. Noting the resolved `auth0-php` version in the release notes would help too.

Separate small question on the line below: `nyholm/psr7` moves from `require-dev` to `require`, but I couldn't find any reference to `Nyholm\` in `src/` since we ship our own PSR-7 under `src/Http/`. Is it there to satisfy v9's `psr/http-factory-implementation` virtual requirement? If so, a one line comment would help the next reader, since it's now a shipped scoped dependency and one of the likelier ones to collide under the `expose-global-*` question above.

src/Actions/Sync.php
```diff
@@ -4,9 +4,12 @@
 
 namespace Auth0\WordPress\Actions;
 
-use Auth0\SDK\Utility\HttpResponse;
+use Auth0\SDK\API\Management\Exceptions\Auth0ApiException;
```
nandan-bhat, line 7: Looks like `Auth0ApiException` is imported but not used, since the code catches `Throwable` throughout. `.php-cs-fixer.dist.php:132` has `no_unused_imports` enabled, so I'd expect `composer phpcs` to fail on it.

Before deleting it though, it made me wonder if finer grained handling was the original plan here. If so it might be worth keeping and using, see my comment on the `catch` block below where `Auth0ApiException` and `getCode()` would help separate a permanent failure from a temporary one.

src/Actions/Sync.php
```diff
@@ -232,19 +240,29 @@ public function onBackgroundSync(): void
...
+                } catch (Throwable $throwable) {
+                    // A single failed item must not abort draining the rest of
+                    // the queue; the row is still removed below to avoid a
+                    // poison message blocking the queue indefinitely.
+                    try {
+                        error_log($throwable->getMessage());
+                    } catch (Throwable) {
                     }
                 }
             }
```
nandan-bhat: This is the other one I'd like to sort before merge. I think as written a failed sync gets logged and then dropped permanently, and I'm not sure that's the intent.

The `catch` sends everything to `error_log`, and then `deleteRow` on line 270 runs unconditionally outside the `catch`, so the queue row goes away whether the sync succeeded or not.

The part that makes me nervous is what changed underneath. In v8 a failed Management call came back as a non-2xx PSR-7 response, `getResults()` returned `null`, and the guarded code quietly did nothing without throwing. In v9 every non-2xx throws instead. I checked `UsersClient::create`, `get`, `listUsersByEmail` and `update`, plus `TicketsClient::changePassword` and `verifyEmail` on the `v9` branch, and they all end in `throw new Auth0ApiException(...)`. So the number of failures reaching this `catch` is likely much higher than the v8 path ever saw.

The case I keep coming back to: an admin bulk creates a few hundred users, the Management API rate limits partway through, `users->create` throws with a 429, we log a line, and `deleteRow` removes the job. Those users would then exist in WordPress with no Auth0 account and no connection record, with no retry and nothing an operator would notice unless they were reading the PHP error log. They wouldn't be able to log in.

Worth noting v9's `RetryDecoratingClient` does retry 408, 429 and 5xx, but retries are already used up by the time the exception gets here, so this `catch` tends to fire in exactly the situation where losing the job hurts most.

Some ideas, take or leave:

- On `Auth0ApiException` with a 4xx that won't ever succeed (400, 404, 409), delete the row as the comment intends.
- On 429 and 5xx, and on `Auth0Exception` for transport failures, leave the row and let the next cron pass pick it up.
- An `attempts` counter would let a genuinely bad message still get evicted after N tries so it can't block the queue forever.
- Surfacing repeated failures as an admin notice rather than only `error_log` would help operators.

That would also give the `Auth0ApiException` import above something to do.

One tiny thing: the `try {} catch (Throwable) {}` around `error_log()` on line 264 can probably go, since `error_log` doesn't throw and an empty catch there would just hide a typo.

src/Actions/Sync.php
```diff
@@ -194,12 +202,12 @@ public function getDatabaseName(?string $dbConnection): ?string
         }
 
         if (null !== $dbConnection) {
-            $response = $this->getResults($this->getSdk()->management()->connections()->get($dbConnection));
+            $response = $this->results($this->getManagement()->connections->get($dbConnection));
```
nandan-bhat, line 206: Not from this PR, just noticed it while reading and it's a one character fix if you're already in here.

Line 201 reads `return $dbConnectionName[$dbConnectionName];`, which indexes the static cache by itself rather than by `$dbConnection`. I think it wants to be `$dbConnectionName[$dbConnection]`.

On a second call within the same request that would give an illegal offset error instead of the cached name. It's invisible today because each cron item only calls this once, and the new blanket `catch` above would cover it anyway. Entirely fine to leave for a separate PR if you'd rather keep this one focused.

src/Plugin.php
```diff
@@ -168,6 +171,41 @@ public function getSdk(): Auth0
...
+     * Returns a v9 Management API client built from the plugin's existing
+     * configuration. Use this instead of the v8-style `getSdk()->management()`,
+     * which is non-functional in auth0-php v9.
+     *
+     * The wrapper fetches and caches a client-credentials token internally. The
+     * plugin's own PSR-18 HTTP client is injected so the scoped build is used
+     * rather than runtime discovery, and the token is cached in the same
+     * WP_Object_Cache pool the SDK config uses (unless caching is disabled).
+     */
+    public function getManagement(): ManagementClient
```
nandan-bhat, line 184: Two thoughts about how this behaves on a site that isn't fully configured yet.

First, there's no readiness check before we build the client. `ManagementClient::buildTokenSupplier` throws `InvalidArgumentException` when both token and clientId/clientSecret are missing, and `TokenProvider::__construct` throws on an empty `domain`, `clientId` or `clientSecret`. We pass `getClientId()` and `getClientSecret()` straight through, even though `Plugin::isReady()` just below at line 218 already answers this question.

That shows up in two places, in different ways:

- `Configuration::onUpdateSync()` calls this inside `try { } catch (Throwable) { }`, so the throw is absorbed, but the message the admin sees says the connection isn't valid when really the plugin just isn't configured yet. Slightly misleading.
- `Sync::eventUserCreated()` calls it with no local guard, so it's only caught by the new blanket `catch` in `onBackgroundSync()`, where per my other comment the queue row is then removed.

Second, a failed construction isn't memoized. `$this->management` stays `null` when the constructor throws, so every queue item in the batch retries and throws again, up to 10 times per cron pass. Not expensive, just noisy.

Would guarding with `isReady()`, or at least checking domain, clientId and clientSecret aren't empty, and throwing a typed plugin exception with a clearer message work here? Separating "not configured" from "invalid connection" in the admin path would be a nice bonus.

src/Plugin.php
```diff
@@ -168,6 +171,41 @@ public function getSdk(): Auth0
...
+            // The 'auth0' cache group is not registered as global, so in a
+            // multisite network WordPress prefixes its keys with the current
+            // blog id automatically. That keeps each site's Management token
+            // isolated without any per-site keying here.
+            $tokenCache = 'disable' !== $this->getOption('tokens', 'caching')
+                ? new WpObjectCachePool()
+                : null;
+
+            $this->management = new ManagementClient(new ManagementClientOptions(
+                domain: (string) $configuration->getDomain(),
```
nandan-bhat, line 199: Small thing, and quite possibly deliberate: this uses `getDomain()`, so a site with a Custom Domain configured won't see it used for Management calls. `SdkConfiguration` also exposes `getCustomDomain()`, and `importConfiguration()` on line 349 does populate it from the `client_advanced.custom_domain` setting. The new Features section in the README mentions Custom Domain support too.

To be clear, I think the tenant domain is the defensible choice here. Management tokens are normally issued against `https://{tenant}/api/v2/` and `ManagementClientOptions` derives `audience` from `domain`, so this reads correct to me.

My only concern is that it's an unstated choice sitting next to a setting the user filled in. Someone who configured a Custom Domain and allowlisted it might be puzzled to see background cron traffic going to the raw tenant domain, and that can be slow to track down through a WAF or egress filter.

A short comment here saying Management calls intentionally use the tenant domain would be plenty for me, maybe with a line in `UPGRADING.md`. If custom domain routing is actually wanted, then `domain: customDomain` with an explicit `audience: "https://{tenantDomain}/api/v2/"` would do it.

UPGRADING.md
```diff
@@ -0,0 +1,48 @@
...
+  $management = wpAuth0()->getManagement();
+  $users = $management->users->list(
+      new ListUsersRequestParameters(['perPage' => 25, 'includeTotals' => true])
+  );
+  foreach ($users as $user) {
+      echo $user->getEmail();
+  }
+  ```
+
+  Inside a plugin action or filter class (anything extending `Auth0\WordPress\Actions\Base` or `Auth0\WordPress\Filters\Base`), the same client is available as `$this->getManagement()`.
```
nandan-bhat, line 41: I think this line promises something that isn't there yet, which is why I'd flag it before merge.

The PR adds `getManagement()` to `src/Actions/Base.php`, but `src/Filters/Base.php` isn't touched here and doesn't have it, only `getSdk()` at line 47. So someone following this from a filter class would hit `Error: Call to undefined method`.

Probably easiest to add the matching `final public function getManagement(): ManagementClient` to `src/Filters/Base.php` so it mirrors `Actions/Base.php`. There is a workaround via `$this->getPlugin()->getManagement()`, but since the docs offer the shorter form it seems kinder to make the shorter form real. Alternatively the sentence could just drop the filter mention, whichever you prefer.

UPGRADING.md
```diff
@@ -0,0 +1,48 @@
...
+  - Sub-resources are reached by property access, not method calls: `->users->list()` rather than `->users()->getAll()`.
+  - Responses are typed objects instead of raw PSR-7 responses. Call `$response->jsonSerialize()` to get the same snake_case array the v8 `HttpResponse::decodeContent()` returned, or use the typed getters (`$response->getEmail()`).
+  - Errors throw exceptions. A non-2xx response raises `Auth0\SDK\API\Management\Exceptions\Auth0ApiException` (use `getCode()` for the HTTP status), and transport-level failures raise `Auth0\SDK\API\Management\Exceptions\Auth0Exception`. There is no more `HttpResponse::wasSuccessful()` check.
+  - Request parameters are camelCase typed objects. For example, creating a user takes a `CreateUserRequestContent` whose keys are `givenName` and `familyName` (not `given_name` / `family_name`), and the `connection` is set inside that object rather than passed as a separate argument.
+  - Listing users only returns the paginated envelope when `includeTotals` is set to `true`. Omitting it yields an empty result, so pass `'includeTotals' => true` when you page through users.
```
nandan-bhat: I don't think this one is accurate, and since it's guidance people will follow it seems worth correcting.

Omitting `includeTotals` shouldn't give an empty result. `UsersClient::list()` returns an `OffsetPager` whose `getItems` closure is `fn ($response) => $response?->getUsers() ?? []`, and `include_totals` only adds `start`, `limit`, `length` and `total` to the envelope. The `users` array comes back either way as far as I can tell.

Maybe reword along the lines of: pass `includeTotals` when you want the total count in the response, it isn't needed to get results.

While you're in here, the README example just above iterates the result with `foreach`, which is correct since `Pager` implements `IteratorAggregate`, and I confirmed `getEmail()` exists on `UserResponseSchema`. One sentence noting the return value is a lazy paginating iterator that makes further requests as you consume it might save someone a surprise, since `count()` and array access won't behave the way a reader might assume.

wpAuth0.php
```diff
@@ -6,15 +6,14 @@
  * Description:       Supercharge your WordPress website with Auth0. Improve account security, add support for multifactor, enable social, passwordless and enterprise connections, and much more.
  * Version:           5.6.1
  * Requires at least: 6.0
- * Tested up to:      6.5.3
+ * Tested up to:      7.0
```
nandan-bhat, line 9: Quick question on the `Tested up to` bump: is there a 7.0 check we can point to? I couldn't verify it from the repo side. `.github/workflows/` only has the AI review action, and there are no files under `tests/` yet even though `phpunit.xml.dist` declares a `tests/Unit` suite, so I had no way to confirm it myself.

Since this line shows up on WordPress.org, it would be nice to have something behind it. If a manual 7.0 run was done then I'm happy to leave it as is. If not, maybe keep it at the last version we actually exercised and bump it once we've had a chance to verify.

Also on the same header, `Domain Path: /languages` is dropped here. I looked into it and it's safe today: no `languages/` directory, no `.po`, `.pot` or `.mo` files tracked, no `load_plugin_textdomain()` call, and no `__()` or `_e()` usage in `src/`. The only loose end is that `Text Domain: wp-auth0` stays, so we end up declaring a text domain with no path, which makes translations slightly more awkward to add later. Not a blocker, I just wanted to check it was intentional. Putting the line back or noting it in the description would both work for me, and if i18n is out of scope then dropping `Text Domain` too would be tidier.

src/Actions/Base.php
```diff
@@ -69,6 +70,11 @@ final public function getPriority(string $event, int $default = 10, string $pref
         return $default;
     }
 
+    final public function getManagement(): ManagementClient
+    {
```
nandan-bhat, line 74: This accessor reads well to me. Just linking it to my `UPGRADING.md` note: the upgrade guide says the same method is available on anything extending `Filters\Base`, but that class isn't touched in this PR and doesn't have it yet. Adding the matching accessor there would make the docs accurate and save someone a puzzling error.

scoper.inc.php
kishore7snehil: Fixed in [`6150185`](https://github.com/auth0/wordpress/commit/6150185).

scoper.inc.php
kishore7snehil: Fixed in [`6150185`](https://github.com/auth0/wordpress/commit/6150185).

src/Actions/Sync.php
kishore7snehil: Fixed in [`6150185`](https://github.com/auth0/wordpress/commit/6150185).

src/Actions/Sync.php
kishore7snehil: Fixed in [`6150185`](https://github.com/auth0/wordpress/commit/6150185).

src/Actions/Sync.php
kishore7snehil: Fixed in [`6150185`](https://github.com/auth0/wordpress/commit/6150185).

src/Plugin.php
kishore7snehil: Fixed in [`6150185`](https://github.com/auth0/wordpress/commit/6150185).

UPGRADING.md
kishore7snehil: Fixed in [`6150185`](https://github.com/auth0/wordpress/commit/6150185).

UPGRADING.md
kishore7snehil: Fixed in [`6150185`](https://github.com/auth0/wordpress/commit/6150185).

src/Actions/Base.php
kishore7snehil: Fixed in [`6150185`](https://github.com/auth0/wordpress/commit/6150185).

src/Plugin.php
kishore7snehil: Added a readiness guard, but at the drain rather than in `getManagement()`, and kept it narrow. The concrete problem was the batch retrying construction ten times against an unconfigured plugin, so `onBackgroundSync()` now returns early when the plugin is not ready and leaves the rows for a later pass. I did not add a typed exception or memoize the failed construction. The SDK already throws a clear message on empty credentials, and with the queue no longer draining while unready, `getManagement()` is not reached in that state on the sync path. The settings path still calls it inside a `try/catch` that treats a throw as "not valid", which is the existing behavior.

Fixed in [`6150185`](https://github.com/auth0/wordpress/commit/6150185).

build.sh
kishore7snehil: Looked into this, including pulling the shipped 5.6.1 zip to see how the original problem was fixed. The check does fire, though. `exclude-namespaces` keeps our own classes unprefixed but does not stop references to scoped vendor symbols in `src/`, and the built `src/` is full of them, so a wrongly-prefixed WordPress class there would be caught. That is the regression that shipped in 5.6.1 and what this guards. On the function point you are right it only matches classes, but that case is prevented by the `exclude-functions` config rather than caught after the build, and I confirmed against a real build that those emit as bare calls. The `assert()` wrapper is handled the same way, by excluding the dev-only stubs package that declares it. So the decision is to keep the class check as a cheap guard for the one regression that reached users, and not grow it into scanning for conditions the config already prevents.

composer.json
kishore7snehil: Agreed, and it is noted for the tagged release rather than the beta. For GA we will pin `auth0/auth0-php` to an exact released version, commit the lock, and switch the build to `composer install` so the signed artifact is reproducible from the commit. For the beta the floating `@beta` is intentional while there is no stable `9.0.0` to resolve against.

On the second question, `nyholm/psr7` is in `require` rather than `require-dev` because v9 only requires the virtual `psr/http-factory-implementation`, and Nyholm is the concrete factory that satisfies it, so it has to ship in the production bundle even though it is not referenced directly in `src/`.

wpAuth0.php
kishore7snehil: On `Tested up to: 7.0`, that is the WordPress version we run the plugin against in the test harness, where activation, login, and sync were exercised this cycle, so it reflects what was actually tested. We will reconfirm it against the latest version tested at release.

On `Domain Path`, dropping it was deliberate. `src/` has no translatable strings and there is no `languages/` folder, so `Domain Path: /languages` pointed at a path that does not exist. Since WordPress 4.6 the directory auto-serves community translations without a bundled folder or a `load_plugin_textdomain()` call, so keeping `Text Domain` alone is enough and lets translations key to the plugin later.

nandan-bhat (approved):

src/Actions/Sync.php
```diff
@@ -232,20 +246,40 @@ public function onBackgroundSync(): void
...
+                    if (429 === $status || $status >= 500) {
+                        error_log($auth0ApiException->getMessage());
 
-                    if ('wp_user_updated' === $payload['event'] && $enabledEvents['wp_user_updated']) {
-                        $this->eventUserUpdated($dbConnection, $payload);
+                        continue;
                     }
+
+                    error_log($auth0ApiException->getMessage());
+                } catch (ClientExceptionInterface $clientException) {
```
rmad17, line 276: Would this handle edge case scenarios like network timeouts/failures and retry the syncing?

src/Actions/Sync.php
```diff
@@ -232,20 +246,40 @@ public function onBackgroundSync(): void
...
+                } catch (Auth0ApiException $auth0ApiException) {
+                    $status = $auth0ApiException->getCode();
+
+                    // Keep the row for the next cron pass on transient failures.
+                    if (429 === $status || $status >= 500) {
+                        error_log($auth0ApiException->getMessage());
 
-                    if ('wp_user_updated' === $payload['event'] && $enabledEvents['wp_user_updated']) {
-                        $this->eventUserUpdated($dbConnection, $payload);
+                        continue;
```
rmad17, line 272: Is there a way to cap retry?

rmad17 (changes requested):

src/Actions/Sync.php
kishore7snehil: Yes. A network timeout or transport failure comes through as `ClientExceptionInterface`, which the second catch keeps in the queue with continue, so the next cron pass retries it.There are two layers actually: the SDK's client retries 408 and 429 within the request first, and if it still fails the row stays queued for the next pass. The same applies to a 429 or a 5xx from the API. Only permanent 4xx responses are dropped.

src/Actions/Sync.php
kishore7snehil: Not right now, and that was a intentional call. Capping needs an attempts counter on the sync table, which means a schema change and migration on existing installs, and for a queue that only drains ten rows per cron pass an uncapped transient retry is naturally throttled rather than a tight loop. 

A permanent failure is already dropped on the first  try, so the only thing that retries indefinitely is a genuinely transient error that never resolves, which we'd rather keep trying to deliver than silently discard

rmad17 (approved):

Issue #962, opened 2026-09-02
WordPress's native redirect_to login parameter not followed after login


hanifn:
### Checklist

- [x] I have looked into the [Readme](https://github.com/auth0/wordpress#readme) and the [documentation](https://auth0.com/docs/customize/integrations/cms/wordpress-plugin), and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/wordpress/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

`redirect_to` is a standard WordPress convention. `wp_login_form()`, `wp_loginout()`, and core's own `wp-login.php` all use it to send a user back to whatever page they were trying to reach before being prompted to authenticate. v4 of this plugin supported it as a first-class, documented feature but for whatever reason, this support is missing in v5.

### Reproduction

1. As a logged-out user (or with an expired session), visit any deep admin URL requiring login, e.g. `wp-admin/edit.php?post_type=page`.
2. Complete login via Auth0.
3. Observe you land on the dashboard/site root instead of back on the page list you were trying to reach.

### Additional context

v4 implementation for reference: request parsing in `WP_Auth0_LoginManager.php:71-73`, threaded through state in `WP_Auth0_Lock.php:34-37` and `get_authorize_params()` (~461-503), honored on callback at `WP_Auth0_LoginManager.php:237-238`, documented in `readme.txt:64`. 
Suggested fix: capture `$_REQUEST['redirect_to']`, thread through the `state` param sent to Auth0, validate with `wp_validate_redirect()` on return, use as the final destination.

### wp-auth0 version

5.6.1

### WordPress version

7.1

### PHP version

8.4

PR #965, opened 2026-09-22
fix: respect redirect_to parameter after login

kishore7snehil:
### Description

After a successful Auth0 login, the plugin was unconditionally redirecting to `/` and ignoring the `redirect_to` parameter that WordPress passes on the login URL. This is standard WordPress behavior - when a logged-out user visits a protected page, WordPress appends `?redirect_to=<destination>` to the login URL so they end up back where they started after authenticating. The behavior was present in v4 and dropped in v5.

The destination is now captured at login initiation, stored in a WordPress transient keyed by a hash of the OAuth state nonce, and retrieved after the token exchange completes. `wp_validate_redirect()` is applied before storing to block off-site redirect attempts. The transient has a 10-minute TTL and is deleted immediately on use.

### References

Closes #962

### Testing

Tested manually against the Docker environment (auth0-PHP 9.3.0, WordPress 7.0, PHP 8.4):

- `redirect_to=/wp-admin/`: after login, browser lands at `/wp-admin/`. Pass.
- No `redirect_to` param: after login, browser lands at `/`. Fallback correct. Pass.
- `redirect_to=https://evil.com`: `wp_validate_redirect()` blocks the external URL, browser lands at `/`. Pass.

Also verified end-to-end with Redis Object Cache active - transient routes through Redis (`SETEX` on login initiation, `GET` + `DEL` on callback).

### Checklist


src/Actions/Authentication.php
```diff
@@ -552,7 +565,19 @@ public function onLogin(): void
             exit;
         }
 
-        wp_redirect($this->getSdk()->login());
+        $loginParams = [];
+
+        if (isset($_REQUEST['redirect_to'])) {
+            $redirectTo = wp_validate_redirect(esc_url_raw($_REQUEST['redirect_to']), '');
```
nandan-bhat: If `redirect_to` comes in as an array (like `redirect_to[]=x`), then `esc_url_raw()` gets an array and throws a fatal on PHP 8. Could we add an `is_string()` check on the `isset` above before we sanitize?

src/Actions/Authentication.php
```diff
@@ -552,7 +565,19 @@ public function onLogin(): void
             exit;
         }
 
-        wp_redirect($this->getSdk()->login());
+        $loginParams = [];
+
+        if (isset($_REQUEST['redirect_to'])) {
```
nandan-bhat: This brings back the `redirect_to` support that v4 documented. Could we add a short note about it to the readme? 

There is no mention of `redirect_to` in `README.md` or `readme.txt` right now, and the docs checkbox in the description is still unchecked.

src/Actions/Authentication.php
```diff
@@ -536,7 +536,20 @@ public function onLogin(): void
                     wp_set_current_user($wpUser->ID);
                     wp_set_auth_cookie($wpUser->ID, true);
                     do_action('wp_login', $wpUser->user_login, $wpUser);
-                    wp_redirect('/');
+
+                    $destination = get_site_url();
```
nandan-bhat: `get_site_url()` fixes the Multisite fallback here. The error path and the already logged in path below still use `wp_redirect('/')` though, so they land on the network root on a subdirectory Multisite. Could we align those with `get_site_url()` too? 

Small thing as well: the already logged in branch returns without clearing the redirect transient.

src/Actions/Authentication.php
kishore7snehil: Fixed in [523ea32](https://github.com/auth0/wordpress/commit/523ea3246edca8f1a507a60c104c58126ac6100b).

src/Actions/Authentication.php
kishore7snehil: Fixed in [523ea32](https://github.com/auth0/wordpress/commit/523ea3246edca8f1a507a60c104c58126ac6100b).

src/Actions/Authentication.php
kishore7snehil: Transient cleared and already-authenticated path aligned to `get_site_url()` in [523ea32](https://github.com/auth0/wordpress/commit/523ea3246edca8f1a507a60c104c58126ac6100b). The error path redirect is addressed in #966 (Multisite fix) and will be fully replaced by #967 (login error feedback) which calls `wp_die()` there instead.

Issue #964, opened 2026-09-02
`onLogin()` silently discards Auth0 error responses


hanifn:
### Checklist

- [x] I have looked into the [Readme](https://github.com/auth0/wordpress#readme) and the [documentation](https://auth0.com/docs/customize/integrations/cms/wordpress-plugin), and have not found a suitable solution or answer.
- [x] I have searched the [issues](https://github.com/auth0/wordpress/issues) and have not found a suitable solution or answer.
- [x] I have searched the [Auth0 Community](https://community.auth0.com) forums and have not found a suitable solution or answer.
- [x] I agree to the terms within the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md).

### Description

When Auth0 returns an `error`/`error_description` on the login callback (expired password, breached-password block, blocked account, etc.), `onLogin()` redirects to `/` and exits with no message shown. v4's equivalent (`WP_Auth0_LoginManager::die_on_login()`) displayed "There was a problem with your log in: {description} [error code: {code}]". 
Expected: some indication is shown to the user that login failed and why.

### Reproduction

1. Trigger any Auth0-side login error (e.g. visit `wp-login.php?error=access_denied&error_description=test&state=<valid-state>`, or trigger a real rejection like a blocked/breached-password account).
2. Observe the browser is redirected to `/` with no error message, indistinguishable from a no-op.

### Additional context

```php
// line ~545-548
if (null !== $error) {
    wp_redirect('/');
    exit;
}
```

### wp-auth0 version

5.6.1

### WordPress version

7.1

### PHP version

8.4

kishore7snehil:
👋 @hanifn, thanks for raising this.

A fix is up in #967.

One deliberate difference from v4 is that the default handler shows a fixed message ("There was a problem with your log in.") rather than the raw `error_description` value. Both `error` and `error_description` are attacker-controllable. Anyone can craft a URL like
`/wp-login.php?error=access_denied&error_description=<any text>` and have that text appear on what looks like your login page without going through Auth0. Displaying the raw value verbatim would expose that as a content injection vector, so the default stays generic.

If your site needs the specific Auth0 message shown to users, hook into `auth0_login_failed`. It receives the sanitized `$error` and `$errorDescription` values for you to render as needed.

PR #967, opened 2026-09-22
fix: show error page when Auth0 returns a login error

kishore7snehil:
When Auth0 returns an error on the login callback (`?error=access_denied`, blocked account,
breached-password policy, a Login Action denying access, etc.), the plugin was silently
redirecting to the site root with no message shown. The same behavior applied when the OAuth
token exchange threw an exception. The default handler uses a fixed generic message rather
than the request-supplied `error_description` - both parameters are attacker-controllable
and displaying them verbatim would allow arbitrary text to appear on the login page without
going through Auth0.

## Changes

- `src/Actions/Authentication.php` - registry: registered `auth0_login_failed` action wired to `onLoginFailed()` with two arguments
- `src/Actions/Authentication.php` - `onLogin()`: reads `error_description` from the callback alongside `error`; replaced silent redirect on `?error=` path with `do_action('auth0_login_failed', ...)`
- `src/Actions/Authentication.php` - `onLoginFailed()`: new method - calls `wp_die()` with a fixed generic message and HTTP 200; passes sanitized `$error` and `$errorDescription` to the hook for custom handlers
- `src/Actions/Authentication.php` - `onExchangeFailed()`: replaced silent `wp_redirect('/')` with `wp_die()`

### References

Closes #964

### Testing

Tested manually against the Docker environment (auth0-PHP 9.x, WordPress 7.0, PHP 8.4).

### Checklist


src/Actions/Authentication.php
```diff
@@ -543,8 +547,9 @@ public function onLogin(): void
         }
 
         if (null !== $error) {
-            wp_redirect('/');
-            exit;
+            error_log('Auth0 login error: ' . $error . ' - ' . ($errorDescription ?? ''));
```
nandan-bhat: `$error` and `$errorDescription` come straight from the callback query, and this path is reachable without login. They are HTML escaped but newlines are not stripped, so a crafted `error_description` with `%0A` can forge extra log lines. Could we strip `\r` and `\n` before logging?

src/Actions/Authentication.php
```diff
@@ -556,11 +561,14 @@ public function onLogin(): void
         exit;
     }
 
-    public function onExchangeFailed(Throwable $_)
+    public function onLoginFailed(string $error, string $errorDescription): void
```
nandan-bhat: Small thing: `$error` and `$errorDescription` are not used here, and `Psalm` (in the dev deps) may flag them as unused. Maybe reference them or add a suppression. Also this message says `log in` while `onExchangeFailed` says `sign-in`, worth keeping the wording consistent.

src/Actions/Authentication.php
```diff
@@ -556,11 +561,14 @@ public function onLogin(): void
...
     {
-        // Custom hook ('auth0_token_exchange_failed') to register when token exchange fails.
-        wp_redirect('/');
-        exit;
+        wp_die('There was a problem with your log in.', 'Login Error', ['response' => 200]);
```
nandan-bhat: This `wp_die()` ends the request, and the default `onLoginFailed` runs at priority `10`, so a custom handler added on `auth0_login_failed` at the same priority never gets to run. Could we register the default at a later priority so custom handlers run first? Otherwise people have to hook below `10` to do anything here.

src/Actions/Authentication.php
```diff
@@ -556,11 +561,14 @@ public function onLogin(): void
...
+    }
+
+    public function onExchangeFailed(Throwable $_): void
+    {
+        wp_die('There was a problem completing your sign-in.', 'Login Error', ['response' => 200]);
```
nandan-bhat: We return HTTP `200` here (and on the login error page above) for what is really a failed sign in. That can hide failures from monitoring and log tools. Was `200` intentional for the friendly page, or would a `403` for real denials like `access_denied` fit better?

src/Actions/Authentication.php
kishore7snehil: Fixed in [5f18e93](https://github.com/auth0/wordpress/commit/5f18e9372675cea971d337a88523c438ad139d52).

src/Actions/Authentication.php
kishore7snehil: Fixed in [5f18e93](https://github.com/auth0/wordpress/commit/5f18e9372675cea971d337a88523c438ad139d52).

src/Actions/Authentication.php
kishore7snehil: Fixed in [5f18e93](https://github.com/auth0/wordpress/commit/5f18e9372675cea971d337a88523c438ad139d52). Extended the registry in `Base` to accept an optional third element as priority (`[method, acceptedArgs, priority]`), and registered `onLoginFailed` at priority 20 so custom handlers at the default priority 10 run first.

src/Actions/Authentication.php
kishore7snehil: The 200 is intentional. Managed WordPress hosts (WP Engine, Kinsta, Pressable) intercept any non-2xx response at the edge and replace the body with their own branded error page, so a 4xx or 5xx here would silently swallow the plugin's message and leave the user with a blank or host-branded page. HTTP 200 is the only status that reliably gets `wp_die()` content through on all hosting environments.

